PR #12, opened 2020-01-22
Added `include_default_values` parameter to `to_dict` function

ulasozguler:
If a field has the default value of its type (e.g. int32 -> 0), it is ignored when message is converted to dictionary. This PR adds an optional parameter to `to_dict` function for including these values in the generated dict.


danielgtaylor:
Thanks!  :+1:

PR #16, opened 2020-01-29
Exclude empty lists from to_dict output

cetanu:
I've found to_dict output to be very verbose and in particular filled with empty lists.

Not quite sure if this PR is 100% okay with the author - if it would be better gated behind some argument passed into to_dict then I'm okay to make the necessary changes.

As far as I know, I've never seen a required list field in a protocol buffer, which is why I have proposed the change in its current state


danielgtaylor:
I believe this is correct. Empty lists should not be serialized by default. Thanks! :+1:

PR #18, opened 2020-02-06
Bugfix twice-nested messages naming error

### LINKED ISSUE #17: Naming bug for twice-nested messages
Hello!

I have an interesting one for you.

I have not had time to look into exactly why this occurs, thought I should report it first because I can reproduce it and N brains are better than 1.

## Problem

Twice-nested messages results in type hints which reference the correct dataclass, however the innermost dataclass has prepended the name of the root message twice

## Example proto file

```proto
syntax = "proto3";

message Root {
  message Parent {
    message Child {
      string foo = 1;
    }
    reserved 1;
    repeated Child child = 2;
    bool bar = 5;
  }
  string name = 1;
  Parent parent = 4;
}
```

## Resulting output

```python
# sources: testfile.proto
# plugin: python-betterproto
from typing import List

import betterproto


@dataclass
class Root(betterproto.Message):
    name: str = betterproto.string_field(1)
    parent: "RootParent" = betterproto.message_field(4)


@dataclass
class RootParent(betterproto.Message):
    child: List["RootParentChild"] = betterproto.message_field(2)
    bar: bool = betterproto.bool_field(5)


@dataclass
class RootRootParentChild(betterproto.Message): # <-------- "RootRoot"
    foo: str = betterproto.string_field(1)

```

## Steps to reproduce

```python
from pathlib import Path
from grpc_tools import protoc


protocol_buffer = '''
syntax = "proto3";

message Root {
  message Parent {
    message Child {
      string foo = 1;
    }
    reserved 1;
    repeated Child child = 2;
    bool bar = 5;
  }
  string name = 1;
  Parent parent = 4;
}
'''
testfile = Path('testfile.proto')
testfile.write_text(protocol_buffer, encoding='utf-8')

output = Path('BUILD')

proto_include = protoc.pkg_resources.resource_filename('grpc_tools', '_proto')
args = [
    __file__,
    f'--proto_path=.',
    f'--proto_path={proto_include}',
    f'--python_betterproto_out={output}',
    str(testfile)
]
protoc.main(args)
```

cetanu:
Fixes #17

However, I'm not entirely sure what other effects this change could have

### DIFF
--- a/betterproto/plugin.py
+++ b/betterproto/plugin.py
@@ -133,8 +133,9 @@ def traverse(proto_file):
 
                 if item.nested_type:
                     for n, p in _traverse(path + [i, 3], item.nested_type):
-                        # Adjust the name since we flatten the heirarchy.
-                        n.name = item.name + n.name
+                        if not n.name.startswith(item.name):
+                            # Adjust the name since we flatten the heirarchy.
+                            n.name = item.name + n.name
                         yield n, p
 
     return itertools.chain(


adament:
I do not think this is a full fix, if I understand the problem and code correctly the same issue should occur for enums as well. I will try to come up with a test case. Furthermore while pragmatic this solution would also avoid prefixing the name if the inner name was already prefixed in the proto file. I think this would be surprising behaviour.

adament:
I was wrong, this also fixes the issue for enums, but this solution means that you do not get prefixing of the dataclass declaration in the following case. But you do get it in the type annotation:
```protobuf
syntax = "proto3";

message Root {
  message Parent {
    message RootParentChild {
      string a = 1;
    }
    enum EnumChild{
      A = 0;
      B = 1;
    }
    message Child {
      string foo = 1;
    }
    reserved 1;
    repeated Child child = 2;
    repeated EnumChild enum_child=3;
    repeated RootParentChild root_parent_child=4;
    bool bar = 5;
  }
  string name = 1;
  Parent parent = 4;
}
```

```python
# sources: test.proto
# plugin: python-betterproto
from typing import List

import betterproto


class RootParentEnumChild(betterproto.Enum):
    A = 0
    B = 1


@dataclass
class Root(betterproto.Message):
    name: str = betterproto.string_field(1)
    parent: "RootParent" = betterproto.message_field(4)


@dataclass
class RootParent(betterproto.Message):
    child: List["RootParentChild"] = betterproto.message_field(2)
    enum_child: List["RootParentEnumChild"] = betterproto.enum_field(3)
    root_parent_child: List["RootParentRootParentChild"] = betterproto.message_field(4)
    bar: bool = betterproto.bool_field(5)


@dataclass
class RootParentChild(betterproto.Message):
    a: str = betterproto.string_field(1)


@dataclass
class RootParentChild(betterproto.Message):
    foo: str = betterproto.string_field(1)
```

Issue #27, opened 2020-03-27
Feature request: Python 3.6 support


Minyus:
As some systems still use Python 3.6, I'd like to request (partial) support for Python 3.6.

nat-n:
This would be great for all the poor folks writing code for enterprise linux environments, which are always a version or two behind.

nat-n:
This could be closed given a minor version bump and release to PyPI?

Minyus:
Yes. Thanks!

PR #29, opened 2020-03-30
#27 Make _get_field_default support py3.6 values of __origin__

nat-n:
From my limited testing I believe this is a sufficient change (along with updating the python version in pyproject.toml and setup.py) to support python 3.6.8.


ahobsonsayers:
This is great! Is there anything stopping this PR being accepted? It would be super useful!

nat-n:
I made a more proper attempt at a contribution for the same purpose so I'll close this one: https://github.com/danielgtaylor/python-betterproto/pull/30

Issue #24, opened 2020-03-12
Missing support for request metadata in stubs


nat-n:
Some API's require adding metadata to requests, such as for authentication but this is not supported by the generated stubs.

To support this would require modifying the `_unary_unary(` and `_unary_stream(` 
 to accept a metadata kwarg and pass it on to `self.channel.request`. Additionally for use cases like mine, having the option to set the default metadata on the Stub instance would be most useful.

I can create a PR if this approach is acceptable in principal.

nat-n:
Resolved by #32

PR #32, opened 2020-04-12
Add ability to provide metadata, timeout & deadline args to requests

nat-n:
#24 

I introduced testing of the generated stubs so that I could test the new functionality.

This is an enhancement of the ServiceStub abstract class that makes it more useful by making it possible to pass all arguments supported by the underlying grpclib request function.

In particular this makes the generated stubs usable with APIs that require clients to provide an *authorization* token.

It extends to the existing high level API by allowing values to be set on the stub instance, and the low level API by allowing values to be set per call.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,7 @@ flake8 = "*"
 mypy = "*"
 isort = "*"
 pytest = "*"
+pytest-asyncio = "*"
 rope = "*"
 v = {editable = true,version = "*"}
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,464 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "b8fc738d4e14598e36269ce0d849489f95562ba047e5663caca9ac02550893ef"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.6"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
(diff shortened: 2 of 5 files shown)


nat-n:
*Bump*

Any thoughts on this?

This addresses an issue that prevents the generated clients from being usable as they are for some projects.

nat-n:
@boukeversteegh thanks for the feedback.

On the point of supporting timeout etc in _unary_unary etc methods; you're right that unfortunately there isn't currently a straight forward way to integrate this with the way Stub methods are currently generated, and I'm sure right now of a way to address this. For most cases setting these values at the stub level is probably good enough, though I think it's easily to imagine it being awkward.

However I do think this will need to be addressed at some point because it's also an issue for another fundamental feature i.e. working with stream requests.

But ultimately for now my I still think it's still worthwhile for these lower level methods to support the extra kwargs for the benefit of users who specifically want to set these values per call and don't mind working with a such as "lower level API". 

For example:

```python
from my_service import MyServiceStub as  BasicMyServiceStub, MyRequest, MyResponse

class MyServiceStub(BasicMyServiceStub):
    async def try_get_stuff_quickly(self, params: Dict[str, Any], timeout=0.01) -> MyResponse:
            "my_package.MyService/GetStuff", 
            MyRequest(**params),
            MyResponse,
            timeout=timeout,
        )
```

boukeversteegh:
You're welcome!
So yeah, I see. Someone could still use those features by subclassing the RPC stub, until we figure out how to pass them per request. I'd say, lets get this merged @danielgtaylor :-)

PR #34, opened 2020-05-04
Fix services using non-pythonified field names

b1naryth1ef:
This PR fixes an issue where generated service stubs would utilize non-pythonified field names. This is most obvious when a service references a message that has fields which are Python keywords, as black will fail to parse the generated file entirely.


cetanu:
I've asked for these changes to be included in a pull-request that touches the same lines.

I'll note both changes/PRs separately in the changelog when it comes to release time.

Thanks for this!

nat-n:
Closing after rebasing and merging it to master manually.

PR #38, opened 2020-05-10
Fixes issue where protoc did not recognize plugin.py as win32 application

boukeversteegh:
Hi~

Thanks for this awesome library!

There's a bug however that I want to look into, but I couldn't run `generate` on my windows system. This PR fixes that.

When running `pyenv run generate` on windows, `protoc` does not accept `../plugin.py` as an executable win32 application.

See for example: https://github.com/protocolbuffers/protobuf/issues/3923

This PR adds a `.bat` file that executes python, which is used when running the `generate.py` on windows.

### DIFF
--- a/betterproto/plugin.bat
+++ b/betterproto/plugin.bat
@@ -1,0 +1,1 @@
+@python ../plugin.py %*
--- a/betterproto/plugin.bat
+++ b/betterproto/plugin.bat
@@ -1,1 +1,2 @@
-@python ../plugin.py %*
+@SET plugin_dir=%~dp0
+@python %plugin_dir%/plugin.py %*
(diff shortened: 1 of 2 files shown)


nat-n:
Out of curiosity (since I know very little about windows):

I take it the issue the .bat file solves is windows not understanding the shebang?

Am I correct in thinking this only matters if you're using the --plugin option on the protoc CLI, since if you use the `--python_betterproto_out` option as in the readme it'll pick up the `protoc-gen-python_betterproto` console script?

boukeversteegh:
> Out of curiosity (since I know very little about windows):
> 
> I take it the issue the .bat file solves is windows not understanding the shebang?
> 
> Am I correct in thinking this only matters if you're using the --plugin option on the protoc CLI, since if you use the `--python_betterproto_out` option as in the readme it'll pick up the `protoc-gen-python_betterproto` console script?

Yes, this is only needed for development. Since I need this change for every other PR I'm working on, and my next PR also includes it with a refactor, we could drop this PR in favor of the next one.

Reference coming :-)

boukeversteegh:
Closing in favor of #51

Issue #39, opened 2020-05-10
Generated client cannot handle Well Known Types as RPC return values


boukeversteegh:
Hi!

I'm running into some trouble returning `google.protobuf.Int32Value` from a unary RPC call.

This is my service:

```protobuf
service Go {
    // ...
    rpc StartTrainingGame (TrainingGameRequest) returns (google.protobuf.Int32Value);
}
```

When I call that service (`await self.client.start_training_game(**kwargs)`), an exception is thrown

`'_SpecialForm' object has no attribute 'FromString'`.

Debugging reveals the following stack frame (`proto.py:54`, in `grpclib`)

```python
    def decode(
        self,  # <grpclib.encoding.proto.ProtoCodec object at 0x000001CE2A687BC8>
        data: bytes,  # b'\x08\x04'
        message_type: Type['IProtoMessage'],  # typing.Union[int, NoneType]
    ) -> 'IProtoMessage':
        return message_type.FromString(data)
```
As is visible, `message_type` is expected to be `IProtoMessage`, with a method `FromString` on it. However, because the Well-Known type is converted to `Union[int, NoneType]`, this method cannot be called there.

I suspect that the current implementation handles these well-known types separately, and this works well when they are embedded in a message, but not when they are the top-level _return-value_ from an RPC call. I intended to use Google's wrappers mainly to be able to return primitive values from RPC-calls.

The generated Client shows the following:

```python
    async def start_training_game(...) -> Optional[int]:
        ...
        return await self._unary_unary(
            "/anagon.ai.go.Go/StartTrainingGame", request, Optional[int],
        )
```

I suppose that the generated code should specify the type as the python-generated equivalent of Int32Value, instead of `Optional[int]`.

Issue #9, opened 2020-01-06
Importing well known types


Luminaar:
Hello,

I'm not sure how to work with well known types.

Having a proto file like this:

```
syntax = "proto3";      
      
package simple;      
      
import "google/protobuf/empty.proto";      
      
service SimpleSender {      
    rpc Send(SendParams) returns (google.protobuf.Empty) {}      
}      
      
message SendParams {      
    string body = 1;      
}```

I get output like this:

```python
# Generated by the protocol buffer compiler.  DO NOT EDIT!      
# sources: Simple.proto      
# plugin: python-betterproto      
from dataclasses import dataclass      
      
import betterproto      
import grpclib          
      
from .google import protobuf      
      
      
@dataclass      
class SendParams(betterproto.Message):      
    body: str = betterproto.string_field(1)      
      
      
class SimpleSenderStub(betterproto.ServiceStub):      
    async def send(self, *, body: str = "") -> protobuf.Empty:      
        request = SendParams()      
        request.body = body         
      
        return await self._unary_unary(      
            "/simple.SimpleSender/Send", request, protobuf.Empty,      
        )

**Comment by danielgtaylor** [comment]
It looks like the `google.protobuf.Emtpy` type is not currently supported. Someone will need to implement it in python-betterproto.

As a workaround you can define your own empty message until this is done.

**Comment by JoaoVasques** [comment]
Howdy people!

The same issue happens when I tried to use `Struct`.

```
from unbabelevents.chat_api import *
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-2-4950c13b1c3b> in <module>
----> 1 from unbabelevents.chat_api import *
~/chat_api.py in <module>
      9
     10 from . import commons
---> 11 from .google import protobuf
     12
     13

ModuleNotFoundError: No module named 'unbabelevents.google'
```

The proto file has the following message

```
syntax = "proto3";

import "google/protobuf/struct.proto";

message Nugget {
    repeated google.protobuf.Struct alerts = 1;
}
```

And the generated python class is

```python

import betterproto

from . import commons
from .google import protobuf


@dataclass
class Nugget(betterproto.Message):
    qe_alerts: List[protobuf.Struct] = betterproto.message_field(1)
```

devova:
> It looks like the `google.protobuf.Emtpy` type is not currently supported. Someone will need to implement it in python-betterproto.
> 

[...]

PR #40, opened 2020-05-10
Support using Google's wrapper types as RPC output values

boukeversteegh:
This PR fixes https://github.com/danielgtaylor/python-betterproto/issues/39

## Well known types as Output

Adds support for returning Google's well-known wrapper types directly from Services.

```protobuf
service Test {
    rpc GetInt32 (Input) returns (google.protobuf.Int32Value);
}
```

**Supported added for**

- `BoolValue`
- `BytesValue`
- `DoubleValue`
- `FloatValue`
- `Int32Value`
- `Int64Value`
- `StringValue`
- `UInt32Value`
- `UInt64Value`

--- 

**Not included in this PR**

- Returning well known types that are **not** wrappers  
  - `Empty`  &mdash; However, [Importing well known types #9](https://github.com/danielgtaylor/python-betterproto/issues/9) can be fixed relatively easily now.
- Well known types (wrappers or not) as _Input_ parameters.  
  ```protobuf
  service Test {
      rpc GetSquare (google.protobuf.Int32Value) returns (google.protobuf.Int32Value);
  }
  ```
  It shouldn't be too hard to support this either.
- Automatically _unwrapping_ the wrapped values when they are returned by the server  
  > This requires some further changes, and would be great, but for now this already is a step forward.

---

**Notes:**

- I was not sure about the right way to test this change. If you think a test is necessary, I'd be happy to discuss a strategy for this.
- I did add `googletypes_service.proto` to the test-cases, so the output can be manually verified.

Let me know if anything doesn't match style requirements, or some other changes need to be made.


betterproto/plugin.py
```diff
@@ -30,38 +30,55 @@
...
+    google.protobuf.wrappers_pb2.Int32Value: "int",
+    google.protobuf.wrappers_pb2.UInt32Value: "int",
+    google.protobuf.wrappers_pb2.BoolValue: "bool",
+    google.protobuf.wrappers_pb2.StringValue: "str",
+    google.protobuf.wrappers_pb2.BytesValue: "bytes",
 }
 
 
-def get_ref_type(package: str, imports: set, type_name: str) -> str:
+def get_wrapper_type(type_name: str) -> (Any, str):
```
nat-n, line 49: I'm not sure about the trade off here in keeping WRAPPER_TYPES declarative as a simple mapping at the cost of not being able to actually use it directly as a mapping.

What about replacing WRAPPER_TYPES and get_wrapper_type here with:

```python
WRAPPER_TYPES = {
    goog_type.DESCRIPTOR.full_name: (goog_type, simple_name)
    for goog_type, simple_name in (
        (google.protobuf.wrappers_pb2.DoubleValue, "float"),
        (google.protobuf.wrappers_pb2.FloatValue, "float"),
        (google.protobuf.wrappers_pb2.Int64Value, "int"),
        (google.protobuf.wrappers_pb2.UInt64Value, "int"),
        (google.protobuf.wrappers_pb2.Int32Value, "int"),
        (google.protobuf.wrappers_pb2.UInt32Value, "int"),
        (google.protobuf.wrappers_pb2.BoolValue, "bool"),
        (google.protobuf.wrappers_pb2.StringValue, "str"),
        (google.protobuf.wrappers_pb2.BytesValue, "bytes"),
    )
}
```

betterproto/plugin.py
```diff
@@ -30,38 +30,55 @@
...
+            return f"Optional[{wrapped_type}]"
 
-    if type_name == "google.protobuf.Timestamp":
-        return "datetime"
+        if type_name == "google.protobuf.Duration":
+            return "timedelta"
+
+        if type_name == "google.protobuf.Timestamp":
+            return "datetime"
+    else:
```
nat-n, line 78: elif?

betterproto/plugin.py
boukeversteegh: I see what you mean, good suggestion! That `get_wrapper_type ` a bit ugly indeed, and would be nice to not need it, while at the same time, I did not like the unchecked references to packages. Your idea solves both, although it embarrassingly took me a several re-reads to fully understand it.

To build on your idea, how about two maps? Same concept but less inlining..

```python
WRAPPER_TYPES = {
    google.protobuf.wrappers_pb2.DoubleValue: "float",
    google.protobuf.wrappers_pb2.FloatValue: "float",
    google.protobuf.wrappers_pb2.Int64Value: "int",
    google.protobuf.wrappers_pb2.UInt64Value: "int",
    google.protobuf.wrappers_pb2.Int32Value: "int",
    google.protobuf.wrappers_pb2.UInt32Value: "int",
    google.protobuf.wrappers_pb2.BoolValue: "bool",
    google.protobuf.wrappers_pb2.StringValue: "str",
    google.protobuf.wrappers_pb2.BytesValue: "bytes",
}

WRAPPERS = {w.DESCRIPTOR.full_name: w for w in WRAPPER_TYPES.keys()}

# usage
wrapper = WRAPPERS['google.protobuf.UInt64Value']

print(wrapper.__module__)
print(WRAPPER_TYPES[wrapper])
```

the only think i don't like about it is having two similar maps with similar names.

betterproto/plugin.py
boukeversteegh: fixed 👍

boukeversteegh:
> Left some minor suggestions on the code.
> 
> I assume that new test example would have failed before?
> 
> If you're up for it I think it would be useful to document the things you've identified as still to do in a new issue, to make it easier to pick up and finish.

The provided example indeed would not have worked before, because the generated call to the `client` passes `Optional[int]` as the `response_type`, which is supposed to be a protobuf message type, so somewhere down the line `Exception: '_SpecialForm' object has no attribute 'FromString'` is thrown. SpecialForm is refering to `Optional`.

```python
    async def start_training_game(self) -> Optional[int]:
        ...
            "/anagon.ai.go.Go/StartTrainingGame", request, Optional[int],
        )
```

I'm currently also working on structuring the tests in such a way that these kinds of examples can be tested more easily, but for now i don't have a test for this..

betterproto/plugin.py
nat-n: My preference is for the single WRAPPER_TYPES dict for conciseness, though I get that the idiom might not be as clear.

As you prefer.

boukeversteegh: Well, I found out we could also just extract the desired type from instance().value , so we don't need the tuples in the map

```python
wrappers = {
    'google.protobuf.DoubleValue': google.protobuf.wrappers_pb2.DoubleValue,
    'google.protobuf.FloatValue' : google.protobuf.wrappers_pb2.FloatValue,
    'google.protobuf.Int64Value' : google.protobuf.wrappers_pb2.Int64Value,
    'google.protobuf.UInt64Value' : google.protobuf.wrappers_pb2.UInt64Value,
    'google.protobuf.Int32Value' : google.protobuf.wrappers_pb2.Int32Value,
    'google.protobuf.UInt32Value' : google.protobuf.wrappers_pb2.UInt32Value,
    'google.protobuf.BoolValue' : google.protobuf.wrappers_pb2.BoolValue,
    'google.protobuf.StringValue' : google.protobuf.wrappers_pb2.StringValue,
    'google.protobuf.BytesValue' : google.protobuf.wrappers_pb2.BytesValue,
}
```

boukeversteegh:
⚠️  This PR is DRAFT and is not (anymore) intended to be merged, but serves as a proof of concept and discussion platform which conclusions will applied to the final PR.

- @nat-n and I had a discussion and we want to first make sure we can create proper test-cases for all scenarios. The PR for this is #51.
- Once that is done, some of the refactors that were part of this PR will be extracted, excluding the bugfixes.
- After that, we will work on modularizing the plugin
- Then I will start creating one or more PR's for these importing bugs, including new test-cases. I hope to be able to align with @danielgtaylor  and @cetanu on the code-generation strategy that is needed for this.

boukeversteegh:
Well, oops. The message above was ment for #47 !
I wondered why my comment didn't show up there, and I couldn't find it back after searching everywhere.

This PR can actually be merged... sorry for the confusion

betterproto/plugin.py
boukeversteegh: I've implemented this concept. Please see #8f0caf1

PR #45, opened 2020-05-18
Adding basic support (untested) for client streaming

hozn:
I haven't written tests for this, but added the base methods for `stream_unary` and `stream_stream` support.  I realize this presents a bit of a challenge when it comes to API consistency since the attrs of the request can't be exploded out into kwargs like the `unary_unary` and `unary_stream` methods.  I went the simple route and just added `request_iterator` param to these methods.

Let me know if changes needed, etc.

### DIFF
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -14,9 +14,9 @@ from typing import (
     Dict,
     Generator,
     Iterable,
-    List,
+    Iterator, List,
     Optional,
-    SupportsBytes,
+    Sequence, SupportsBytes,
     Tuple,
     Type,
     TypeVar,
@@ -424,6 +424,7 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
 
 # Bound type variable to allow methods to return `self` of subclasses
 T = TypeVar("T", bound="Message")
+ST = TypeVar("ST", bound="IProtoMessage")
 
 
 class Message(ABC):
@@ -1028,5 +1029,32 @@ class ServiceStub(ABC):
             route, grpclib.const.Cardinality.UNARY_STREAM, type(request), response_type
         ) as stream:
             await stream.send_message(request, end=True)
+            async for message in stream:
+                yield message
+
+    async def _stream_unary(
+        self, route: str, request_iterator: Iterator["IProtoMessage"], request_type: Type[ST], response_type: Type[T]
+    ) -> T:
+        """Make a unary request and return the stream response iterator."""
+        async with self.channel.request(
+            route, grpclib.const.Cardinality.STREAM_UNARY, request_type, response_type
+        ) as stream:
+            for message in request_iterator:
+                await stream.send_message(message)
+            await stream.send_request(end=True)
+            response = await stream.recv_message()
+            assert response is not None
+            return response
+
+    async def _stream_stream(
+        self, route: str, request_iterator: Iterator["IProtoMessage"], request_type: Type[ST], response_type: Type[T]
+    ) -> AsyncGenerator[T, None]:
+        """Make a unary request and return the stream response iterator."""
+        async with self.channel.request(
+            route, grpclib.const.Cardinality.STREAM_STREAM, request_type, response_type
+        ) as stream:
+            for message in request_iterator:
+                await stream.send_message(message)
+            await stream.send_request(end=True)
             async for message in stream:
                 yield message
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -14,7 +14,8 @@ from typing import (
     Dict,
     Generator,
     Iterable,
-    Iterator, List,
+    Iterator,
+    List,
     Optional,
     SupportsBytes,
     Tuple,
@@ -1033,7 +1034,11 @@ class ServiceStub(ABC):
                 yield message
 
     async def _stream_unary(
-        self, route: str, request_iterator: Iterator["IProtoMessage"], request_type: Type[ST], response_type: Type[T]
+        self,
+        route: str,
+        request_iterator: Iterator["IProtoMessage"],
+        request_type: Type[ST],
+        response_type: Type[T],
     ) -> T:
         """Make a unary request and return the stream response iterator."""
         async with self.channel.request(
@@ -1047,7 +1052,11 @@ class ServiceStub(ABC):
             return response
 
     async def _stream_stream(
-        self, route: str, request_iterator: Iterator["IProtoMessage"], request_type: Type[ST], response_type: Type[T]
+        self,
+        route: str,
+        request_iterator: Iterator["IProtoMessage"],
+        request_type: Type[ST],
+        response_type: Type[T],
     ) -> AsyncGenerator[T, None]:
         """Make a unary request and return the stream response iterator."""
         async with self.channel.request(
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -16,7 +16,7 @@ from typing import (
     Iterable,
     Iterator, List,
     Optional,
-    Sequence, SupportsBytes,
+    SupportsBytes,
     Tuple,
     Type,
     TypeVar,
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -1040,7 +1040,7 @@ class ServiceStub(ABC):
         request_type: Type[ST],
         response_type: Type[T],
     ) -> T:
-        """Make a unary request and return the stream response iterator."""
+        """Make a stream request and return the response."""
         async with self.channel.request(
             route, grpclib.const.Cardinality.STREAM_UNARY, request_type, response_type
         ) as stream:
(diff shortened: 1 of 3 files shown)


betterproto/templates/template.py
```diff
@@ -63,11 +63,28 @@ class {{ service.py_name }}Stub(betterproto.ServiceStub):
 
     {% endif %}
     {% for method in service.methods %}
-    async def {{ method.py_name }}(self{% if method.input_message and method.input_message.properties %}, *, {% for field in method.input_message.properties %}{{ field.name }}: {% if field.zero == "None" and not field.type.startswith("Optional[") %}Optional[{{ field.type }}]{% else %}{{ field.type }}{% endif %} = {{ field.zero }}{% if not loop.last %}, {% endif %}{% endfor %}{% endif %}) -> {% if method.server_streaming %}AsyncGenerator[{{ method.output }}, None]{% else %}{{ method.output }}{% endif %}:
+    async def {{ method.py_name }}(self
+        {%- if not method.client_streaming -%}
```
nat-n, line 67: How does the templating engine handle all this extra whitespace?

betterproto/templates/template.py
```diff
@@ -77,21 +94,42 @@ async def {{ method.py_name }}(self{% if method.input_message and method.input_m
         request.{{ field.name }} = {{ field.name }}
             {% endif %}
         {% endfor %}
+        {% endif %}
 
         {% if method.server_streaming %}
+            {% if method.client_streaming %}
```
nat-n, line 100: some indentation is off around here

betterproto/__init__.py
```diff
@@ -1030,3 +1031,30 @@ async def _unary_stream(
...
+                await stream.send_message(message)
+            await stream.send_request(end=True)
+            response = await stream.recv_message()
+            assert response is not None
+            return response
+
+    async def _stream_stream(
+        self, route: str, request_iterator: Iterator["IProtoMessage"], request_type: Type[ST], response_type: Type[T]
+    ) -> AsyncGenerator[T, None]:
+        """Make a unary request and return the stream response iterator."""
```
nat-n, line 1052: wrong comment

betterproto/__init__.py
```diff
@@ -14,9 +14,9 @@
     Dict,
     Generator,
     Iterable,
-    List,
+    Iterator, List,
```
nat-n, line 17: need to run black

betterproto/__init__.py
```diff
@@ -14,9 +14,9 @@
...
     Optional,
-    SupportsBytes,
+    Sequence, SupportsBytes,
```
nat-n, line 19: if Sequence used?

betterproto/templates/template.py
hozn: > I suspect some tests will be required for this.

Yeah, I didn't see a precedent for testing the generated services, but I can look at adding these.  I figured I'd get this out here for discussion before investing that time, in case this wasn't the direction you wanted to go.

> Does it make sense to add these without solving the problem of how to expose it to the end user?

Well, I don't have a better idea than letting the caller provider an iterator of the request objects.  It doesn't match the other methods, but I'm not sure how one would match those `unary_*` methods.  I suppose another direction could be a method that feeds into an iterator under the hood (wrapped within the method), but I don't know how you then signal the end of the method call, etc.  The request_iterator feels clean and simple and certainly works for my own use cases.

> How does the templating engine handle all this extra whitespace?

The extra whitespace is mitigated by the `{%-` stripping tag as well as the params to the jinja environment.  Finally, running everything through `black` in the end also removes any lingering artifacts, so the output is fine / matches current [albeit with the new `request_iterator` when appropriate]..

betterproto/templates/template.py
hozn: I'll take a look at that.  I'm not entirely sure how one likes to indent jinja templates, I can skip the indent of the nested if-statement.

betterproto/__init__.py
hozn: Thanks! :)

betterproto/__init__.py
hozn: Yes, will do!

betterproto/__init__.py
hozn: Doesn't appear to be; must have been an earlier idea; I'll clean that up.

betterproto/__init__.py
hozn: Black makes a change or two to `plugin.py`, which I did not include in my commit/branch (since AFAIK it wasn't related to changes I made).

betterproto/templates/template.py
hozn: Hopefully the new version looks better.

betterproto/templates/template.py
nat-n: I guess using the request_iterator argument as you describe is probably the most straightforward approach. Could also do it with a two way generator or a context provider, but neither of these is as simple to use.

I also have a PR open where I add a service test: https://github.com/danielgtaylor/python-betterproto/pull/32/commits/c762c9c549f693eced8165cb8f9e54a4f440ea55 And I think an example usage in the readme would also be necessary.

That said, I'm not a designated maintainer so can't promise anything about what will be acceptable to merge.

hozn: @nat-n , thanks for that link to the test; that certainly looks less ceremonious than the testing setup I was seeing in grpclib itself.  Thanks!

cetanu:
If we can get a test for this I think it's a worthwhile addition. The jinja seems sane to me, but I would be particularly interested to remove all concern for whitespace weirdness, I know that the hyphens do some trimming but with an indent sensitive language like python it can be interesting.

If you already know how to add a test for this then go ahead, otherwise I (or we) will have to dig around a bit

cetanu:
Also, can I get you to pull in the changes from #34 ?
It's a small change that hits the same lines as yours

betterproto/__init__.py
```diff
@@ -1030,3 +1032,38 @@ async def _unary_stream(
...
+        request_iterator: Iterator["IProtoMessage"],
+        request_type: Type[ST],
+        response_type: Type[T],
+    ) -> AsyncGenerator[T, None]:
+        """Make a stream request and return the stream response iterator."""
+        async with self.channel.request(
+            route, grpclib.const.Cardinality.STREAM_STREAM, request_type, response_type
+        ) as stream:
+            for message in request_iterator:
+                await stream.send_message(message)
```
nat-n, line 1066: It can't be assumed that the client will want to send all the messages before receiving any. 

Ideally I think this method should:

1. accept `request_iterator: Union[Iterator["IProtoMessage"], AsyncIterator["IProtoMessage", None]]`
2. Return type AsyncIterator[T], since it doesn't actually accept data being sent in
3. if Iterator["IProtoMessage"] is given then wrap it in an AsyncIterator to proceed
4. then iterate to receive data from the caller and send it to the server and to receive the data from he server and yield it concurrently, thus allowing upload/download to occur in any valid order.

Or alternatively it might more elegant for this method to just return an AsyncGenerator, and then it's up to the caller to manage sending and receiving from that generator. This would imply an update to the _unary_stream method for consistency, though it should be possible to keep is backwards compatible. I've not totally thought this through but I think it's worth exploring.

Sorry if you were hoping this would be simpler :P I'd be happy to help iterate towards a better solution.

hozn: Good suggestions; I'll incorporate these.

hozn:
Thanks @nat-n  and others!  I'll work to update the PR, though I should say that this has been degraded in my priorities, as I've rewritten the compiler portion to match our idiosyncratic protobuf/grpc needs a little more closely. (We predominantly use the threaded grpcio classes, so I needed generated client/server code to work with those.  We also require `google.protobuf.Any` support, so I implemented that.).  I am hoping that most of my local changes can be rebased off upstream in future.

nat-n:
@hozn I see this as a high priority issue, with a few more problems to solve, do you mind if I work on it a bit from here?

hozn:
@nat-n  -- I don't mind at all; I would appreciate it greatly!  I'm eager to see the implementation for concurrently consuming from the request iterator (note, perhaps better would be to have that be an `Iterable[IProtoMesage]` -- feels like a `list` should be valid input?) and emitting from the stream.  I bookmarked some notes on how to combine the iterators like that, but it sounds like you probably had a clear vision there.

nat-n:
@hozn

Sure thing. Since you've already spent some time thinking about it, if you have more thoughts to share or could spare some time to review related changes then that would be appreciated.

hozn:
@nat-n , sounds great; yes, I'll continue to participate; I'm sure I'll learn a lot.  I appreciate you taking this on, and apologize that I didn't get it to a more complete stage (w/ tests).

I'll also be happy to share the work that I've done with wrapping the default (`grpio`-based) client/server code.  I'm not sure that it's worthy of a pull request, as it's really ugly and uses protected members from that package (I don't really see a practical way around that), but it may be worth separating out the services from the generated messages.  In my fork of the compiler code, I ended up creating a structure like:
```
nspkg/
  __init__.py  # <-- namespace pkg
  module/
    __init__.py  # <-- message classes
    services.py  # <-- grpcio-baesd (threaded) services
    services_aio.py # <-- grpclib-based (asyncio) services
```

Unrelated to this, I also am testing out an implementation of `google.protobuf.Any`; once I've spent more time figuring out the edge cases there, I'll write that up as a pull request/discussion.

nat-n:
@hozn That's sounds similar to what I had in mind. I also have a PoC for grpcio integration, though I think this problem is important and complex enough that it's worth exploring different approaches in order to get it right. If you push what you have to your fork I'll have a look.

nat-n:
Closing in favour of #83

PR #46, opened 2020-05-19
Improve performance of serialize/deserialize by caching type information of fields in class

jameslan:


jameslan:
Here's my test result.

The performance test is executed in WSL using the following command:

```shell
$ time pytest --repeat 3000 betterproto/tests/test_inputs.py
```

which repeats serializing and deserializing 3000 times in the test case after the data is loaded.

Result:


|env |before|after|
|:-:|:-:|:-:|
|py3.6|155.74s|22.660s|
|py3.7|72.57s|20.889s|
|py3.8|55.678s|16.739s|

cetanu:
This is pretty awesome. I assume you did some profiling and found that `_type_hint` and `get_type_hints` take up a bit of time when creating these dataclasses?

I've read through the entire diff but I'm just going to give another look for thoroughness

betterproto/__init__.py
```diff
@@ -443,25 +522,20 @@ def __post_init__(self) -> None:
...
             meta = FieldMetadata.get(field)
 
             if meta.group:
-                # This is part of a one-of group.
-                group_map["fields"][field.name] = meta.group
-
-                if meta.group not in group_map["groups"]:
-                    group_map["groups"][meta.group] = {"current": None, "fields": set()}
-                group_map["groups"][meta.group]["fields"].add(field)
+                group_map.setdefault(meta.group)
```
cetanu, line 530: Can you help me grok this? is this the same result before and after, or does this dict have very different structure with this change?

From what I can tell, some keys are now removed. There is no concern for these groups being overwritten? or the possibility is not there because such behavior cannot occur in the protocol buffer that this will process. Is that right?

betterproto/__init__.py
```diff
@@ -426,6 +430,81 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
...
+        if item == 'cls_by_field':
+            self.init_cls_by_field()
+            return self.cls_by_field
+
+
+def make_protoclass(cls):
+    setattr(cls, "_betterproto", ProtoClassMetadata(cls))
+
+
+def protoclass(*args, **kwargs):
```
cetanu, line 502: Am I right in my understanding, that this causes users to have to use the created `protoclass` instead of creating dataclasses?

However:

1. It's auto-generated
2. It behaves like a dataclass

So this doesn't seem to break anything.

cetanu:
Can you go through and update the README? I think that's the main source of docs at this time

betterproto/__init__.py
nat-n: Unless I'm missing something this looks like classes the inherit from `betterproto.Message` with a regular dataclass decorator will break if the user upgrades due to _betterproto not being set.

Breaking changes may be OK if this comes as a major version bump, but would still be better to avoid it or at least try and handle it gracefully.

I suggest accessing _betterproto via a property that checks whether it is set, and degrades gracefully for dataclasses lacking the extra annotation, maybe with a warning message. If going this route then there should be a test for it.

betterproto/__init__.py
jameslan: Originally, the `group_map` is like:

```python
{
    "fields": {
        "field1": "group1",
        "field2": "group1",
    },
    "groups": {
        "group1": {
            "current": Field1,
            "fields": set(Field1, Field2),
        },
    },
}
```
Within this dict, only `current` is object scoped, others are all class metadata. So that I kept `current` in this dict and moved others into `ProtoClassMetadata`, as `oneof_group_by_field` and `oneof_field_by_group`. Please refer to method `__setattr__`, L555 of changed `__init__.py` file.

In this way, building the lookup table between fields and groups is executed once per class. And also, all the object s of the class share the same lookup table.

Now in the object, `_group_map` indicates the current value of oneof group:

```python
{
    "group1": Field1,
}
```

`__setattr__` will keep this updated

betterproto/__init__.py
jameslan: `protoclass` is a wrapper of `dataclasses.dataclass`. When this decorator is applied, the `dataclass` is still applied and the data class is still generated. The extra thing `protoclass` does, is to call `make_protoclass` to scan metadata of fields, and store in attribute `_betterproto` of the generated class.

In this way, the scan operation is done once per class, instead of per object, or per method call.

So this change requires to replace `@dataclasses.dataclass` with `@betterproto.protoclass` in generated python files. Which means users need to rerun `protoc` to update.

jameslan:
Yes the result of profiling shows that `get_type_hints `, as well as functions in module `dataclasses`, are kind of pricey.

So this PR is to make it not call them on accessing each field since they will never change.

And the next step is to remove invocations to `dataclasses.fields` from instance method, which may improve performance 10%~20%

betterproto/__init__.py
```diff
@@ -426,6 +430,81 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
 T = TypeVar("T", bound="Message")
 
 
+class ProtoClassMetadata:
+    cls: "Message"
```
nat-n, line 434: Should be `Type["Message"]` ?

betterproto/__init__.py
```diff
@@ -426,6 +430,81 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
...
+            self.init_default_gen()
+            return self.default_gen
+
+        if item == 'cls_by_field':
+            self.init_cls_by_field()
+            return self.cls_by_field
+
+
+def make_protoclass(cls):
+    setattr(cls, "_betterproto", ProtoClassMetadata(cls))
```
nat-n, line 499: Maybe this could be achieved in a backwards compatible (and maybe more direct?) way by making ProtoClassMetadata a [non-data descriptor](https://realpython.com/python-descriptors/) of the Message class?

So something like: 

```python
class ProtoClassMetadata:
    def __lazy_init__(self, cls: "Message"):  # rename __init__ to __lazy_init__
        ...
    
    def __get__(self, obj, type=None):
        if self.cls is None:
            self.__lazy_init__(obj.__class__)
        return self

class Message:
    _betterproto = ProtoClassMetadata()
    ...
```

betterproto/__init__.py
nat-n: @jameslan Thanks for clarifying.

betterproto/__init__.py
jameslan: Good idea! But this will add a class field to class `Message`, so that all subclasses share one instance of `ProtoClassMetadata`.

The following approach works:

```python
class ProtoMessageMetadata:
    metadata: Dict[Type["Message"], ProtoClassMetadata]

    def __init__(self):
        self.metadata = {}

    def __get__(self, instance, owner):
        meta = self.metadata.get(owner)
        if not meta:
            meta = ProtoClassMetadata(owner)
            self.metadata[owner] = meta
        return meta
```

which stores metadata of all subclasses in `Message` class. The drawback is that it references all subclasses and those dynamically created classes won't be released.

Let me see if we could use `@property` to solve this.

betterproto/__init__.py
jameslan: You are right, thanks for finding this!

betterproto/__init__.py
jameslan: It works. Now it's transparent to the users. Thanks!

betterproto/__init__.py
jameslan: Now the extra decorator is removed. The metadata is populated at the first access, instead of when data class is created.

jameslan:
> Can you go through and update the README? I think that's the main source of docs at this time

I checked README and think that it needn't any change, since now this PR is fully transparent and backward compatible.

betterproto/__init__.py
nat-n: Nice! Except the memory leak :/ What about using a weakref data structure to avoid the data leak scenario?

Or maybe it’s simpler if the reference to the metadata is actually stored in the target class itself... so there’s no need to keep the reference to the class. The metadata could still be accessed through this ProtoMessageMetadata.__get__, or through a regular @property on message that lazy initialises it.

jameslan:
Rebased to head of master

betterproto/tests/test_inputs.py
```diff
@@ -22,67 +23,74 @@
...
-def test_message_can_be_imported(test_case_name: str) -> None:
-    importlib.import_module(f"{plugin_output_package}.{test_case_name}.{test_case_name}")
+TestData = namedtuple("TestData", "plugin_module, reference_module, json_data")
 
 
-@pytest.mark.parametrize("test_case_name", test_case_names)
-def test_message_can_instantiated(test_case_name: str) -> None:
-    plugin_module = importlib.import_module(f"{plugin_output_package}.{test_case_name}.{test_case_name}")
+@pytest.fixture(scope="module", params=test_case_names)
+def test_data(request):
```
nat-n, line 30: nice.

betterproto/tests/test_inputs.py
```diff
@@ -22,67 +23,74 @@
...
+    yield TestData(
+        plugin_module=importlib.import_module(f"{plugin_output_package}.{test_case_name}.{test_case_name}"),
+        reference_module=importlib.import_module(f"{reference_output_package}.{test_case_name}.{test_case_name}_pb2"),
+        json_data=get_test_case_json_data(test_case_name),
+    )
+
+    sys.path.remove(reference_module_root)
+
+
+def test_message_can_be_imported(test_data: TestData) -> None:
```
nat-n, line 51: looks even more pointless now.

jameslan: Removed.

jameslan:
> LGTM. Don't forget to run black.

Updated. Thanks!

boukeversteegh:
Performance is great! Let's get this merged. I've already prepared an update to #55 to be compatible with the changes done in `test_inputs.py`. Let's get this merged ~

boukeversteegh:
Hey @nat-n ~ your last commit 5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79 created merge conflicts for this Pr and mine.

Unfortunately running black on the PR branches doesn't solve the conflicts because the lines can turn out to be different.

I was able to work around it by reverting 5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79 temporarily on a local copy of master, merging that copy into my branch, then rerunning black on the same files. you could do the same, @jameslan 

```bash
git branch # check that you are on your PR branch
git checkout master
git pull danielgtaylor master # pull from danielgtaylor repository, use your remote name
git checkout -b master_no_blacken
git revert 5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79
git checkout - # go back to PR branch
git merge master_no_blacken
black betterproto/tests/test_inputs.py # rerun black on affected files (if needed)
git commit -am 'Blacken'
```

jameslan:
> Hey @nat-n ~ your last commit [5e2d9fe](https://github.com/danielgtaylor/python-betterproto/commit/5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79) created merge conflicts for this Pr and mine.
> 
> Unfortunately running black on the PR branches doesn't solve the conflicts because the lines can turn out to be different.
> 
> I was able to work around it by reverting [5e2d9fe](https://github.com/danielgtaylor/python-betterproto/commit/5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79) temporarily on a local copy of master, merging that copy into my branch, then rerunning black on the same files. you could do the same, @jameslan
> 
> ```shell
> git branch # check that you are on your PR branch
> git checkout master
> git pull danielgtaylor master # pull from danielgtaylor repository, use your remote name
> git checkout -b master_no_blacken
> git revert 5e2d9febea2a2982b7cc372ef5f4b1b7d2233e79
> git checkout - # go back to PR branch
> git merge master_no_blacken
> black betterproto/tests/test_inputs.py # rerun black on affected files (if needed)
> git commit -am 'Blacken'
> ```

Thanks!

I just 

```shell
$ git rebase master
$ git checkout --theirs betterproto/tests/test_inputs.py
$ git add betterproto/tests/test_inputs.py
$ git rebase --continue
```

To override the conflict.

nat-n:
@jameslan Sorry about that. I did it to simplify some other merging, but should have been more cautious about touching so many lines.

Maybe there should be build time check for black correctness to keep everything in sync after the current round of PRs.

Issue #25, opened 2020-03-16
Import bug - two packages with the same name suffix should not cause naming conflict


andrewmartinfinney:
I am keen to use better proto for a project that contains a simple heirarchy of protobuf files.
There is a top level proto file which contains the include statements:
```
package adapter.v1;

import "Src/Products/HFSS/Adapter/v1/API/aedt-method-data-v0.proto";
import "Src/Products/HFSS/Adapter/v1/API/aedt-properties-v0.proto";
```
The imported proto have package definitions of the form:
```
package aedt.method_data.v0;
```
and
```
package aedt.properties.v0;
```

In the python generated by better_proto for the top level proto I see:

```
from .aedt.method_data import v0
from .aedt.properties import v0
```
with an example reference:
```
@dataclass
class UpdateBigHouseRequest(betterproto.Message):
    # The big_house resource which replaces the resource on the server.
    big_house: v0.BigHouse = betterproto.message_field(1)
    # The update mask applies to the resource. For the `FieldMask` definition,
    # see https://developers.google.com/protocol-
    # buffers/docs/reference/google.protobuf#fieldmask The grpc gateway will add
    # entries for all fields in the request body JSON (i.e. the payload above) to
    # the update_mask. This means a REST client needs to filter-out fields not
    # present in the update_mask from the body JSON it sends.
    update_mask: protobuf.FieldMask = betterproto.message_field(2)
```
I hope you agree this is incorrect.

I *think* the solution is to revise the code in plugin.py lines 79 and 80 in the get_ref_type function
to something like:
```
        imports.add(f"import {'.'.join(parts[:-1])}")
        type_name = '.'.join(parts)
```
I have not tried this as I have had difficulty building from a clone of the repository.
(perhaps the following should be separate issue...)

I tried 
```python -m pipenv install --dev```
at the root of the clone
and got:
```
Traceback (most recent call last):
  File "C:\Users\afinney\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\afinney\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\afinney\AppData\Roaming\Python\Python38\site-packages\pipenv\__main__.py", line 4, in <module>
    cli()
[...]

Issue #37, opened 2020-05-07
Unnecessary critical path runtime dependency on grpclib.client


nat-n:
executing `import betterproto` takes 110ms and consumes >12MB of RAM. Almost all of this is spent on [importing grpclib.client](https://github.com/danielgtaylor/python-betterproto/blob/master/betterproto/__init__.py#L28).

This probably isn't a major concern  for most applications, but may be for some, and is also easily avoided since it is [only used for typechecking](https://github.com/danielgtaylor/python-betterproto/blob/master/betterproto/__init__.py#L1008) so could be imported under the `TYPE_CHECKING` condition so as to have no effect at runtime.

nat-n:
Fixed by #52

PR #47, opened 2020-05-21
[Draft] Fix many import related bugs

boukeversteegh:
Fixes lots of issues with imports.

## Circular imports

two python files referring to eachother, preventing imports

I've solved this by wrapping generated Types in quotes (`List["Foo"]`)

## Imports from parent package

I've solved this for now by using absolute imports there, but I think this still needs a change to use `from ... import [parent]` instead.

## Module and filename conflicts  

consider two messages `animal.domestic.cow`, and `animal.bird` which is in the `animal.proto` file. the python files would look like  
```
animal/domestic/cow.py
animal.py
```
  that means `animal` is now both a file and a module, and it results in conflicts.

I've solved this by exporting _all_ packages as modules:

```
animal/
   __init__.py  # contains bird
   domestic/
     __init__.py  # contains cow
```

You can import the animal package simply with `import animal` as before, but still also `import animal.domestic`.

## import name conflicts

Importing two subpackages with the same name resulted in conflicts. See: #25 

Solved this by always aliasing the imported package with full package path:

`import from .aedt.method_data import v0 as aedt_method_data_v0`

## relative imports

Relative imports work correctly for siblings, children and descendants.

### DIFF
--- a/betterproto/plugin.py
+++ b/betterproto/plugin.py
@@ -213,14 +213,10 @@ def generate_code(request, response):
     template = env.get_template("template.py")
 
     output_files = collections.defaultdict()
-    # def log(message):
-    #     sys.stderr.write(f"{message}\n")
-
-    root_packages = {proto_file.package.split('.')[0] for proto_file in request.proto_file}
 
     for proto_file in request.proto_file:
-        out = proto_file.package
-        if out == "google.protobuf":
+        package = proto_file.package
+        if package == "google.protobuf":
             continue
 
         if package:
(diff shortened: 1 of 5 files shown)


nat-n:
This looks good.

Perhaps rounding up issues of this nature would also fit well within the scope of this change: https://github.com/danielgtaylor/python-betterproto/issues/37

betterproto/plugin.py
```diff
@@ -174,266 +212,253 @@ def generate_code(request, response):
...
-                                    "value": v.number,
-                                    "comment": get_comment(proto_file, path + [2, i]),
-                                }
-                                for i, v in enumerate(item.value)
-                            ],
-                        }
-                    )
-
-                    output["enums"].append(data)
+                read_item(item, path, proto_file, output_file_data['package'], output_file_data['template_data'], global_messages)
```
nat-n, line 249: black would have something to say about the line length here.

betterproto/plugin.py
```diff
@@ -174,266 +212,253 @@ def generate_code(request, response):
...
-
-        for proto_file in options["files"]:
-            # print(proto_file.message_type, file=sys.stderr)
-            # print(proto_file.service, file=sys.stderr)
-            # print(proto_file.source_code_info, file=sys.stderr)
-
+    # read messages
+    global_messages = []
+    for output_file_name, output_file_data in output_files.items():
+        for proto_file in output_file_data["files"]:
```
nat-n, line 247: this bit could be streamlined a bit more since you're here:

```python
global_messages = []
for _, output_file_data in output_files.items():
    output_file_data['template_data'] =  {
        "package": output_file_data["package"],
        "files": [f.name for f in output_file_data["files"]],
        "imports": set(),
        "datetime_imports": set(),
        "typing_imports": set(),
        "messages": [],
        "enums": [],
        "services": [],
    }

    # read messages
    for proto_file in output_file_data["files"]:
        for item, path in traverse(proto_file):
            read_item(
                item, 
                path, 
                proto_file,
                # < No need to pass packages seperately becuase it's in template_data >
                output_file_data['template_data'], 
                global_messages
            )
```

betterproto/plugin.py
```diff
@@ -174,266 +212,253 @@ def generate_code(request, response):
...
+            "server_streaming": method.server_streaming,
+        }
+        # sys.stderr.write(f'METHOD: {method_data}\n')
+        data["methods"].append(method_data)
+
+        if method.server_streaming:
+            template_data["typing_imports"].add("AsyncGenerator")
+    template_data["services"].append(data)
+
+
```
nat-n, line 461: The diff makes it look like this is all new. Is it fair to assume none of it really is?

boukeversteegh: Yes. Its because I've moved some code into functions, and the changed indentation has made it look new.

However, I did make an important change on line 427

     if msg["full_name"] == method.input_type.lstrip('.'):

to compare with the full name, rather than part of the name, so that all messages can be matched consistently.

PR #51, opened 2020-05-22
Organize test-cases

boukeversteegh:
To be able to add test-cases involving multiple `proto` files, and isolate packages and generated code, it's needed to have the proto files in separate folders, and generate the output isolated as well.

It was also hard to work with the tests in the current structure as the input data and generated output was not separated from the actual test files, making it hard to delete/cleanup the output, and navigating to different tests.

After this is merged, we can start creating tests for different import and package structure scenarios and fix the import bugs more reliably and rigorously than my current PR #47 which is now not unit-tested at all.


Pipfile
```diff
@@ -20,6 +20,8 @@ stringcase = "*"
 black = "*"
 backports-datetime-fromisoformat = "*"
 dataclasses = "*"
+betterproto = {editable = true,path = "."}
+pytest-asyncio = "*"
```
boukeversteegh, line 25: i should remove this

nat-n:
@boukeversteegh Awaiting the changes we discussed offline to merge.

boukeversteegh:
Done!

betterproto/tests/generate.py
```diff
@@ -1,84 +1,60 @@
...
-from google.protobuf.descriptor_pool import DescriptorPool
-from google.protobuf.json_format import MessageToJson, Parse
+def generate(whitelist: Set[str]):
+    path_whitelist = {os.path.realpath(e) for e in whitelist if os.path.exists(e)}
+    name_whitelist = {e for e in whitelist if not os.path.exists(e)}
+
+    test_case_names = set(get_directories(inputs_path))
+
+    for test_case_name in sorted(test_case_names):
+        test_case_path = os.path.realpath(os.path.join(inputs_path, test_case_name))
```
nat-n, line 21: Not worth a rewrite, but are you familiar with pathlib as a nicer alternative to os.path?

betterproto/tests/test_inputs.py
```diff
@@ -1,32 +1,99 @@
...
+
+
+excluded_test_cases = {'googletypes_response', 'service'}
+test_case_names = {*get_directories(inputs_path)} - excluded_test_cases
+
+plugin_output_package = 'betterproto.tests.output_betterproto'
+reference_output_package = 'betterproto.tests.output_reference'
+
+
+@pytest.mark.parametrize("test_case_name", test_case_names)
```
nat-n, line 25: nice.

PR #52, opened 2020-05-22
Only import types from grpclib when type checking

nat-n:
#37  This speeds up initial import and reduces the memory foot print.

Also ran black.


boukeversteegh:
Great! Looks like a low hanging fruit~ unfortunately the reformatting will cause merge conflicts in all other open PR's so I would suggest to either revert the blackening here or update the pr after the other branches are merged.

PR #55, opened 2020-05-23
Add intentionally failing test-cases for unimplemented bug-fixes

boukeversteegh:
This PR adds support for test-cases that document known bugs or missing features that are currently expected to fail.

```python
pytest
betterproto/tests/test_features.py ........                                            [  6%]
betterproto/tests/test_inputs.py .......x.X...x..........xx...........x..x..x.xx...... [ 84%]
```

It is intended to help contributors (including myself) fixing bugs by already including the proper test-cases for them, and to document what we know to not work (well) yet.

A short explanation was added to the Readme for how this works.

I encourage everyone to create PR's with failing test-cases, in addition to opening issues, so that the bugs can be clearly defined and are perfectly reproducible.

---
Additional changes:

- json-tests skip cases without a json file
- added several test-cases for import / package related bugs


betterproto/tests/inputs/xfail.py
```diff
@@ -0,0 +1,10 @@
+# Test cases that are expected to fail, e.g. unimplemented features or bug-fixes.
+# Remove from list when fixed.
+tests = {
```
nat-n, line 3: It's not obvious to me what's the point of this centralised approach (and the overhead) over just applying the xfail markers where they're needed?

betterproto/tests/generate.py
```diff
@@ -11,27 +13,38 @@
 os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
 
 
+def clear_directory(path: str):
+    for file_or_directory in glob.glob(os.path.join(path, '*')):
```
nat-n, line 17: Not too big a deal, but using pathlib would make some of this code a bit neater.
e.g.
```python
def clear_directory(path: Path):
    for file_or_directory in path.glob("*"):
        if file_or_directory.is_dir():
            shutil.rmtree(file_or_directory)
        else:
            file_or_directory.unlink()
```

betterproto/tests/util.py
```diff
@@ -46,3 +46,12 @@ def protoc_plugin(path: str, output_dir: str):
 def protoc_reference(path: str, output_dir: str):
     subprocess.run(f"protoc --python_out={output_dir} --proto_path={path} {path}/*.proto", shell=True)
 
+
+def get_test_case_json_data(test_case_name, json_file_name=None):
+    test_data_file_name = f'{json_file_name if json_file_name else test_case_name + ".json"}'
```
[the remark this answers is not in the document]
boukeversteegh: oops! totally right. it ended up like that after refactor/extraction. will fix

betterproto/tests/inputs/xfail.py
boukeversteegh: adding markers to an individual test is easy, using `@pytest.mark.xfail`, but adding it to a specific test-case within `@pytest.mark.parametrize("test_case_name", test_cases)` requires changing `test_cases` as shown [here](https://docs.pytest.org/en/latest/skipping.html#skip-xfail-with-parametrize).

```
@pytest.mark.parametrize(
    ("n", "expected"),
    [
        (1, 2),
        pytest.param(1, 0, marks=pytest.mark.xfail),
        pytest.param(1, 3, marks=pytest.mark.xfail(reason="some bug")),
        (2, 3),
        (3, 4),
        (4, 5),
        pytest.param(
            10, 11, marks=pytest.mark.skipif(sys.version_info >= (3, 0), reason="py2k")
        ),
    ],
)
```
basically, it means that the list of test-cases needs to be hardcoded, instead of auto-generated from the directories. when hardcoding the test-case list, the xfail marks also need to be repeated for every test function individually.

boukeversteegh: so the TestCases class generates those lists, and takes the ones that should be marked as xfail from a list. I've externalized the list because I expect it to be frequently updated, and it was hard to explain where to do it when it somewhere inside a constructor arg.

betterproto/tests/generate.py
boukeversteegh: i see.. ok. i will change this as part of one of the other PR's that will be upcoming, and use pathlib from now on. hope this can be non-blocking for now

betterproto/tests/inputs/xfail.py
nat-n: gotcha. fair enough.

boukeversteegh:
> Haven't decided on a merge order yet (wanna give other a little more time to comment), but be aware of: https://github.com/danielgtaylor/python-betterproto/pull/46/files#diff-e0733f0b556057d1a738d4b4888b3f68R30

Thank you for pointing it out @nat-n ! I've merged #46 locally and solved the merged conflict. I'm waiting for #46 to be merged, then I will be able to update this branch (otherwise i would have to include the entire PR #46 in this one)

betterproto/tests/util.py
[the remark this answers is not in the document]
boukeversteegh: done!

betterproto/tests/mocks.py
```diff
@@ -5,7 +5,7 @@
 
 class MockChannel(Channel):
     # noinspection PyMissingConstructor
-    def __init__(self, responses: List) -> None:
+    def __init__(self, responses: List = []) -> None:
```
nat-n, line 8: You can probably get away with it here but having a mutable default value is dangerous.

Usual workaround:
```python
def __init__(self, responses: Optional[List] = None) -> None:
    self.responses = [] if responses is None else responses
```

boukeversteegh: Wow! First I thought, why would this make any difference? Then I looked it up, and I'm shocked. Thanks! Learned something 😃 for those wondering, see [here](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)

will change this!

cetanu: pylint has a rule `dangerous-default-value` that will pick this up.

If pylint is added to this project it should be configured with allowed rules, instead of excluded ones to avoid headaches :)

PR #62, opened 2020-05-24
Cache field metadata, to avoid calling `dataclasses.fields` to get more than 10% performance improvement

jameslan:
- Added two more `dict`s in `ProtoClassMetadata`: `field_name_by_number` and `meta_by_field_name`,
so that instance methods could use them to loop protobuf fields.
- Renamed `_group_map` to `_group_current` as it stores the current field of each group only
- Used field name in `_group_current` instead of `dataclasses.Field`

### DIFF
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -14,11 +14,10 @@ from typing import (
     Collection,
     Dict,
     Generator,
-    Iterable,
     List,
     Mapping,
     Optional,
-    SupportsBytes,
+    Set,
     Tuple,
     Type,
     TypeVar,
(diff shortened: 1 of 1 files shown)


boukeversteegh:
Looks good to me!

- 10% improvement in speed
- code complexity is same
- readability is same, or even improved
- small changes

Thank you 👍

betterproto/__init__.py
```diff
@@ -435,14 +434,22 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
 
 
 class ProtoClassMetadata:
-    cls: Type["Message"]
+    oneof_group_by_field: Dict[str, str]
```
nat-n, line 437: Probably worthwhile declaring slots here:

```python
__slots__ = (
    'cls', 
    'oneof_group_by_field', 
    'oneof_field_by_group', 
    'default_gen', 
    'cls_by_field', 
    'field_name_by_number', 
    'meta_by_field_name', 
)
```
I'd be curious if you can measure a different in performance by doing so.

betterproto/__init__.py
```diff
@@ -451,30 +458,36 @@ def __init__(self, cls: Type["Message"]):
...
+        self.meta_by_field_name = by_field_name
 
-        self.init_default_gen()
-        self.init_cls_by_field()
+        self.default_gen = self.init_default_gen(cls, fields)
+        self.cls_by_field = self.init_cls_by_field(cls, fields)
 
-    def init_default_gen(self):
+    @staticmethod
+    def init_default_gen(cls, fields):
```
nat-n, line 473: Since it's no longer initializing any fields maybe `get_default_gen` is a slightly more appropriate name.

betterproto/__init__.py
jameslan: Not really. The measure is to execute the following command 3 times:

```shell
$ pytest --repeat 3000 betterproto/tests/test_input.py
```

Before `__slot__` magic, it took 21.35s, 21.59s, 21.35s.
After, it took 21.70s, 21.62s, 21.52s.

According to https://book.pythontips.com/en/latest/__slots__magic.html, the purpose of `__slot__` is to save memory from a large number of objects.

betterproto/__init__.py
jameslan: Done.

nat-n:
Ok, I'm happy, though it looks like black is not.

jameslan:
> Ok, I'm happy, though it looks like black is not.

Oops, forgot it. Now fixed.

PR #69, opened 2020-05-25
Bugreports

boukeversteegh:
Adds test cases for several known bugs. This PR does not include fixes for them.

If you think any of these are not bugs, please comment. We can then also close the related issue. 🙏 thank you


betterproto/tests/inputs/import_service_input_message/request_argument.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto3";
+
+message RequestArgument {
```
nat-n, line 3: What does this relate to?

betterproto/tests/inputs/import_service_input_message/test_import_service.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from betterproto.tests.mocks import MockChannel
+from betterproto.tests.output_betterproto.import_service_input_message.import_service_input_message import (
+    RequestResponse,
+    TestStub,
+)
+
+
+@pytest.mark.xfail(reason="Request Input Messages are not imported for service")
```
nat-n, line 10: Could be useful for the reason to reference the issue number.

betterproto/tests/inputs/import_service_input_message/request_argument.proto
boukeversteegh: Thanks for reviewing 😄 
Accidental left-over. Will remove 👍

betterproto/tests/inputs/import_service_input_message/test_import_service.py
boukeversteegh: Good idea! Will do

betterproto/tests/inputs/import_service_input_message/test_import_service.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from betterproto.tests.mocks import MockChannel
+from betterproto.tests.output_betterproto.import_service_input_message.import_service_input_message import (
+    RequestResponse,
+    TestStub,
+)
+
+
+@pytest.mark.xfail(reason="#68 Request Input Messages are not imported for service")
```
cetanu, line 10: nice ❤️

betterproto/tests/generate.py
```diff
@@ -45,22 +48,39 @@ def generate(whitelist: Set[str]):
...
-        )
-        test_case_output_path_betterproto = os.path.join(
-            output_path_betterproto, test_case_name
+        print(f"Generating output for {test_case_name}")
+        try:
+            generate_test_case_output(test_case_name, test_case_input_path)
+        except subprocess.CalledProcessError as e:
+            failed_test_cases.append(test_case_name)
+
+    if failed_test_cases:
```
nat-n, line 57: 🥇

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.json
```diff
@@ -0,0 +1,16 @@
+{
+  "int": "value-for-int",
```
nat-n, line 2: what's with these values?

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.proto
```diff
@@ -0,0 +1,37 @@
...
+
+// Tests that messages may contain fields with names that are python types
+
+message Test {
+    // https://docs.python.org/2/library/stdtypes.html#numeric-types-int-float-long-complex
+    string int = 1;
+    string float = 2;
+    string complex = 3;
+
+    // https://docs.python.org/3/library/stdtypes.html#generator-types
```
nat-n, line 11: https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.proto
```diff
@@ -0,0 +1,37 @@
...
+
+    // https://docs.python.org/3/library/stdtypes.html#memory-views
+    string memoryview = 24;
+
+    // https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset
+    string set = 25;
+    string frozenset = 26;
+
+    // https://docs.python.org/3/library/stdtypes.html#dict
+    string map = 27;
```
nat-n, line 33: dict?

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.json
boukeversteegh: nothing much :-P

but the fields neem some values.. i don't like `{"int": "int"}` very much, because when you see errors it becomes confusing if its the value or the key. same, if all fields have the same value like "foo", you can't see which field crashed if only the value was printed in the error.
so i did it like this.

i could also make all of them ints with unique values like 1 2 3...

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.proto
boukeversteegh: oops 😅 forgot to invite dict to the party 🥳

betterproto/tests/inputs/namespace_builtin_types/namespace_builtin_types.proto
boukeversteegh: fixed! thanks

Issue #65, opened 2020-05-24
Add black to build pipeline and readme


boukeversteegh:
As @nat-n suggested, enforcing [black](https://pypi.org/project/black/) as a validation step in the pipeline would reduce the number of merge-conflicts, and make sure that all contributors are aware that this project requires the use of black.

We could also mention it in the readme, or keep the readme short and let contributors find out after pushing

I prefer mentioning it.. as before black, i did spend some time formatting my code, which turned out to be a wasted effort

nat-n:
Yip, it should be in the readme too.

nat-n:
It would also be good to rename `template.py` to something like `template.py.j2` since that fact that it looks like a python file but isn't really causes confusion for black and certain other build tools (looking at you rpmbuild).

nat-n:
Addressed in PR #72

Issue #42, opened 2020-05-13
Editable package "v" makes newest pipenv fail


jameslan:
In Pipfile, there's such line in `dev-packages`:

```
v = {editable = true,version = "*"}
```

pipenv 2018.11.26 doesn't complain but the head of pipenv master branch will fail because an editable package should be a local path or git repo URL.

Also,

```sh
$ pip install -e v
ERROR: v is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with svn+, git+, hg+, or bzr+).
```

jameslan:
Forgot to metion

boukeversteegh:
What is this package? I see a placeholder mentioned on pip:

https://pypi.org/project/v/

Can we simply remove this line?

boukeversteegh:
@nat-n Would you know if this line is needed?

nat-n:
Worth a shot, might have actually been a typo via the cli  😳

nat-n:
Addressed in PR#72

PR #72, opened 2020-05-27
Make CI check formatting is black

nat-n:
#65

- rename templay.py to template.py.j2 to stop it being confused for a source file
- Add task to CI workflow
- Update readme
- fix remaining formatting issues
- remove typo from Pipfile #42


Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b8fc738d4e14598e36269ce0d849489f95562ba047e5663caca9ac02550893ef"
+            "sha256": "44ae793965dc2b6ec17f0435a388846248b8a703cf857470b66af84227535950"
```
boukeversteegh, line 4: It looks like dependencies were also updated with newer versions, I guess as a side-effect of removing the `v` line?

Is it possible to remove the package without reinstalling the other packages (and thereby changing lockfile versions?)

> Sidenote: I think we should switch our `pipfile` to use explicit versions, instead of '*', but thats for another pr

nat-n: Possible... a bit hacky (would have to manually edit the lock file).

The way I see it, it's only minor version upgrades which *should* be safe since being loose on these is pretty standard. Carefully reviewing every minor version upgrade of a dependency isn't usually considered worthwhile, the benefit of the lockfile currently is just that we see & sort of control *when* they change.

None of the changes are in direct runtime dependencies.

My plan is to make things a bit more explicit along with migrating to poetry in the near future.

PR #77, opened 2020-05-28
Rearrange plugin import to make import errors more helpful

nat-n:
This addresses an issue where if the user happens to have black installed in their environment but not the other dependencies when running the protoc plugin then the resulting import error (No module named 'google') is not very helpful.


nat-n:
Agreed the wording could be better. Updated.

boukeversteegh:
nice!

PR #78, opened 2020-05-29
Basic general support for Google Protobuf

boukeversteegh:
This refactor serves as a foundation to improve support for Google's well-known types.

What's in this PR 🎁 ?

- We actually use Betterproto to compile Google's types! 💪😄 (no handcrafting, no reliance on Google's compiled output)
- All Google types included, so basic support is there for all types (i.e. they are imported properly and can be instantiated).
- Fixes small issue where `google.protobuf.Value` was mistaken for a wrapper class

To allow full support, we should implement `to/from_json/dict` for each type that has special behavior.

Notably `Struct` and `Value` need will not yet automatically produce the correct json values or cannot be parsed. Test-cases were added for this.


boukeversteegh:
ℹ️  This PR also includes the work of #76 . if others agree with this approach (compiling google proto files with betterproto), then 76 can be closed in favor of this one.

betterproto/lib/google/protobuf.py
```diff
@@ -0,0 +1,1312 @@
...
+    """
+
+    # Null value.
+    NULL_VALUE = 0
+
+
+@dataclass
+class Any(betterproto.Message):
+    """
+    `Any` contains an arbitrary serialized protocol buffer message along with a
```
nat-n, line 113: This makes me think comment handling could be better...

betterproto/plugin.py
```diff
@@ -4,8 +4,11 @@
 import os.path
 import sys
 import textwrap
-from collections import defaultdict
-from typing import Dict, List, Optional, Type
+import re
+from typing import List
+
+import betterproto
```
nat-n, line 10: what's this for?

betterproto/plugin.py
```diff
@@ -251,11 +193,13 @@ def generate_code(request, response):
                         field_type = f.Type.Name(f.type).lower()[5:]
 
                         field_wraps = ""
-                        if f.type_name.startswith(
-                            ".google.protobuf"
-                        ) and f.type_name.endswith("Value"):
-                            w = f.type_name.split(".").pop()[:-5].upper()
-                            field_wraps = f"betterproto.TYPE_{w}"
+                        match_wrapper = re.match(
+                            "\\.google\\.protobuf\\.(.+)Value", f.type_name
```
nat-n, line 197: Using a raw string would be nicer.
```suggestion
                            r"\.google\.protobuf\.(.+)Value", f.type_name
```

betterproto/tests/mocks.py
```diff
@@ -8,6 +8,7 @@ class MockChannel(Channel):
     def __init__(self, responses=None) -> None:
         self.responses = responses if responses else []
         self.requests = []
+        self._loop = None
```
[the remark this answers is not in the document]
boukeversteegh: yup! :-)

README.md
```diff
@@ -354,6 +355,21 @@ $ pipenv run generate
 $ pipenv run test
 ```
 
+### (Re)compiling Google Well-known Types
+
+Betterproto includes compiled versions for Google's well-known types at [betterproto/lib/google](betterproto/lib/google).
+Be sure to regenerate these files when modifying the plugin output format, and validate by running the tests.
+
+Normally, the plugin does not compile any references to `google.protobuf`, since they are pre-compiled. To force compilation of `google.protobuf`, set this environment variable: `PROTOBUF_OPTS=INCLUDE_GOOGLE`.
```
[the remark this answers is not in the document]
boukeversteegh: ah that's actually a mistake. it should be `BETTERPROTO_OPTS`!

betterproto/__init__.py
```diff
@@ -1154,3 +1091,16 @@ async def _unary_stream(
             await stream.send_message(request, end=True)
             async for message in stream:
                 yield message
+
+
+from .lib.google.protobuf import (
```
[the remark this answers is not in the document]
boukeversteegh: yes, i know :-/ it can't be at the top due to circular imports. the compiled google classes depend on the base classes in `__init__.py`, while init also needs the Google classes. Once we split up Init hopefully this can be avoided

betterproto/lib/google/protobuf.py
boukeversteegh: 👍  Apparently, newlines are stripped. I've made an issue: #80

betterproto/plugin.py
boukeversteegh: see line 201
`re` is to match `*Value`-wrapper classes more precisely than before.
`betterproto` is imported to check if we have an actual implementation for that wrapping class.

betterproto/__init__.py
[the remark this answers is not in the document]
nat-n: Oh dang, could you put a comment noting that it's to avoid circular imports?

betterproto/plugin.py
```diff
@@ -251,11 +193,13 @@ def generate_code(request, response):
...
+                        )
+                        if match_wrapper:
+                            wrapped_type = "TYPE_" + match_wrapper.group(1).upper()
+                            if wrapped_type in dir(betterproto):
```
nat-n, line 201: I think `hasattr(betterproto, wrapped_type)` is more direct/idomatic.

boukeversteegh: ah yeah~ i'll change that

betterproto/plugin.py
```diff
@@ -182,6 +120,9 @@ def get_comment(proto_file, path: List[int], indent: int = 4) -> str:
 
 
 def generate_code(request, response):
+    plugin_options = os.environ.get("BETTERPROTO_OPTS")
+    plugin_options = plugin_options.split(" ") if plugin_options else []
```
danielgtaylor, line 124: Is it a common pattern to split on spaces for protobuf compiler options? If not, then maybe a comma would be better?

```sh
$ BETTERPROTO_OPTS=OPT1,OPT2,OPT3 protoc ...
```

betterproto/__init__.py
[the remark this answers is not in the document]
danielgtaylor: Sadly this was one of the reasons for the giant single file. Breaking everything up and resolving circular imports looked like significant work at the time so I put it off. Sorry!

betterproto/plugin.py
boukeversteegh: Actually, I don't know. I think that `protoc` doesn't work with environmental variables in general, as it defines its own options through commandline arguments. But for plugins its not possible to define arguments, hence this workaround through env vars. I haven't checked if other plugins third party plugins have specific ways to handle that.

So perhaps we can do it however we like. For example

- similar like commandline arguments (`--compile-google --opt2 --opt3`)
- or with tags `OPT1,OPT2`. 
- or space delimited `OPT1 OPT2`

If we want flexibility, probably option 1 would be most conventional. Similar to how java runs with `$JAVA_OPTS`.

Let me know which approach you prefer

boukeversteegh: @danielgtaylor 

For the topic of Any, I'm not sure if we can say that we support it, just by importing it. So for now I wouldn't check that box yet 😃 

Please have a look at: [#79 Google Types - Any support](https://github.com/danielgtaylor/python-betterproto/issues/79)

nat-n: Regarding the passing of options I think there's a better way. I can't find any proper documentation for it but after a bit of digging it looks like protoc supports a `--custom_opt` option which can be passed a value to be set in the `parameter` field on CodeGeneratorRequest message received by the plugin!

This is probably the best way to do it. To be cli friendly using a comma separated sequence of flags is probably the best approach.

boukeversteegh: ahhhh! i actually did try this, but it didn't work. turns out i had protoc version 3.0.0 (ubuntu was much behind). 3.12 has this and works perfectly!

updated the handler and readme

betterproto/plugin.py
boukeversteegh: fixed!

Issue #87, opened 2020-06-10
Import bug - Message from Capitalized package is mistaken for Nested Type


boukeversteegh:
The compiler does not distinguish the following two cases:

```protobuf
package somepackage.Core;

message Message {}
```

```protobuf
package somepackage;

message Core {
  message Message {}
}
```

as they are both:

```
somepackage.Core.Message
```

Protoc (as far as I've been able to figure out) reports the full type name, but not which part of that is the package, and which is the real type (`somepackage.Core: Message` vs `somepackage: Core.Message`).

In 99% of cases, this will not be a problem, because if users will follow the Protobuf Style Guide, packages will be `lowercase`, and types `CapitalCamelCase`, and the distinction can be made based on that.

https://developers.google.com/protocol-buffers/docs/style#packages

> Package name should be in lowercase, and should correspond to the directory hierarchy. e.g., if a file is in my/package/, then the package name should be my.package.

https://developers.google.com/protocol-buffers/docs/style?hl=cs-CZ#message-and-field-names

> Use CamelCase (with an initial capital) for message names – for example, SongServerRequest. Use underscore_separated_names for field names (including oneof field and extension names) – for example, song_name.

Despite these style guides, there may be some users who are using Capitalized packages and for some reason cannot change the packages to lowercase.

Then, these packages will not be imported correctly from depending packages / messages.

Issue #61, opened 2020-05-24
Import bug - import root package from root


boukeversteegh:
See commit https://github.com/boukeversteegh/python-betterproto/commit/5b6997870a93d4dd18beec31bf1bdbfdd8610b86

test case `import_root_sibling`

Issue #60, opened 2020-05-24
Import bug - import root package from child package


boukeversteegh:
See commit https://github.com/boukeversteegh/python-betterproto/commit/cdf7645722342b74d7204a08bf2b93b7ec542032

Test case: `import_root_package_from_child`

Issue #59, opened 2020-05-24
Import bug - import parent package from child package


boukeversteegh:
See commit https://github.com/boukeversteegh/python-betterproto/commit/ca20069ca350e3b78707f1470041ce9e8bb0e0bc

Testcase `import_parent_package_from_child`

PR #88, opened 2020-06-10
🍏 Fix imports

boukeversteegh:
Fixes:
- :apple: #23 Import Bug - No arguments are generated for stub methods when using import with proto definition
- :apple: #25 Import bug - two packages with the same name suffix should not cause naming conflict
- :apple: #57 Import bug - import child package from root
- :apple: #58 Import bug - import child package from package
- :apple: #59 Import bug - import parent package from child package
- :apple: #60 Import bug - import root package from child package
- :apple: #61 Import bug - import root package from root
- :apple: #11 ALL_CAPS message fields are parsed incorrectly.

---
Functional changes:
-  :green_apple: Python files are generated as pure packages `my/package/__init__.py` instead of `my/package.py`
   - This prevents import name clashes such as `my/package.py` shadowing `my/package/foo.py`
   - Output is entirely determined by the protobuf package structure, and not filenames.
- :green_apple: All imports statements are generated relatively
   - You can therefor compile into other locations than the root of the python project, e.g. `output` and import the generated code with `from output.messages import PingMessage`
   - Generated files now correctly resolve *children*, *sibilings*, *parents*, *cousins* and *root* packages
-  :green_apple: All import statements are aliased (`import user.v1 as user_v1` / `import post.v1 as post_v1` ) to avoid naming conflicts
- :warning: To correctly resolve **Nested Types** vs **Packaged Types** (in a package), it is assumed that Style Guide is followed, and that package names are lowercase, while Types are Capitalized (`Message.SubType` vs `package.SomeType`). I expect this to not cause anyone trouble, but #87 documents this behavior just in case.

---

Technical improvements:

- :kiwi_fruit: Standard Tests (`inputs/`) will automatically recognize messages called `Test` as the main subject, even if they are in packages or in a proto-file different from the test-case name.
- :kiwi_fruit: Removed dependency on `stringcasing`, which had a very buggy implementation for `snake_case` that we relied on. Replaced with custom implementation that covers our scenarios (let me know if some are not covered!).


betterproto/tests/util.py
```diff
@@ -60,3 +63,25 @@ def get_test_case_json_data(test_case_name, json_file_name=None):
 
     with open(test_data_file_path) as fh:
         return fh.read()
+
+
+def find_module(
+    module: ModuleType, predicate: Callable[[ModuleType], bool]
+) -> Optional[ModuleType]:
+    if predicate(module):
```
nat-n, line 71: This could use a docstring.

betterproto/tests/util.py
```diff
@@ -60,3 +63,25 @@ def get_test_case_json_data(test_case_name, json_file_name=None):
...
+        return module
+
+    module_path = pathlib.Path(*module.__path__)
+
+    for sub in list(module_path.glob("**/")):
```
nat-n, line 76: does the trailing slash mean something?

betterproto/tests/util.py
```diff
@@ -60,3 +63,25 @@ def get_test_case_json_data(test_case_name, json_file_name=None):
...
+        if not sub.is_dir() or sub == module_path:
+            continue
+        sub_module_path = sub.relative_to(module_path)
+        sub_module_name = ".".join(sub_module_path.parts)
+
+        sub_module = importlib.import_module(f".{sub_module_name}", module.__name__)
```
nat-n, line 82: it looks to me like this will get upset if it finds a directory without an __init__.py, maybe searching with `.glob("**/__init__.py")` then `sub_module_path = sub.parent.relative_to(module_path)` would be safer/more explicit?

betterproto/tests/test_inputs.py
```diff
@@ -50,14 +56,17 @@ def apply_xfail_marks(test_set: Set[str], xfail: Set[str]):
...
     path=inputs_path,
     services=test_input_config.services,
-    xfail=test_input_config.tests,
+    xfail=test_input_config.xfail,
 )
 
 plugin_output_package = "betterproto.tests.output_betterproto"
 reference_output_package = "betterproto.tests.output_reference"
 
+TestData = namedtuple("TestData", ["plugin_module", "reference_module", "json_data"])
```
nat-n, line 65: Python trivia of the day: [tuples are generally better than lists where mutability isn't needed](https://stackoverflow.com/questions/68630/are-tuples-more-efficient-than-lists-in-python)

So we can save some microseconds here with:
```
TestData = namedtuple("TestData", ("plugin_module", "reference_module", "json_data"))
```

betterproto/tests/test_inputs.py
```diff
@@ -75,11 +84,19 @@ def test_data(request):
...
     sys.path.append(reference_module_root)
 
+    plugin_module = importlib.import_module(f"{plugin_output_package}.{test_case_name}")
+
+    plugin_module_entry_point = find_module(plugin_module, module_has_entry_point)
+
+    if not plugin_module_entry_point:
+        raise Exception(
+            f"Test case {repr(test_case_name)} has no entry point. "
+            + "Please add a proto message or service called Test and recompile."
```
nat-n, line 94: don't need the + here

betterproto/tests/test_get_ref_type.py
```diff
@@ -77,6 +79,227 @@ def test_importing_google_wrappers_unwraps_them(google_type: str, expected_name:
...
+
+def test_import_nested_child_from_root():
+    imports = set()
+    name = get_ref_type(package="", imports=imports, source_type="nested.child.Message")
+
+    assert imports == {"from .nested import child as nested_child"}
+    assert name == "nested_child.Message"
+
+
+def test_import_deeply_nested_child_from_root():
```
nat-n, line 125: I find "..._from_root" a little confusing, like root implies sys.path...

I would think of this as importing from sibling, as this present module and the module that it's importing from share the same parent package.

I would think of what is described below as import_from_sibling instead as not importing at all, unless I've misunderstood?

betterproto/tests/test_get_ref_type.py
```diff
@@ -77,6 +79,227 @@ def test_importing_google_wrappers_unwraps_them(google_type: str, expected_name:
...
+    name = get_ref_type(package="", imports=imports, source_type="Message")
+
+    assert imports == set()
+    assert name == '"Message"'
+
+
+def test_import_nested_siblings():
+    imports = set()
+    name = get_ref_type(package="foo", imports=imports, source_type="foo.Message")
+
```
nat-n, line 158: Would make sense to still include `assert imports == set()` here no?

betterproto/tests/test_get_ref_type.py
```diff
@@ -77,6 +79,227 @@ def test_importing_google_wrappers_unwraps_them(google_type: str, expected_name:
...
+        package="package",
+        imports=imports,
+        source_type="package.deeply.nested.child.Message",
+    )
+
+    assert imports == {"from .deeply.nested import child as deeply_nested_child"}
+    assert name == "deeply_nested_child.Message"
+
+
+def test_import_root_sibling():
```
nat-n, line 147: Looks like there's no importing here... how about:
```suggestion
def test_reference_...():
```

betterproto/tests/test_get_ref_type.py
```diff
@@ -77,6 +79,227 @@ def test_importing_google_wrappers_unwraps_them(google_type: str, expected_name:
...
+
+    assert imports == {"from .... import ancestor as ___ancestor__"}
+    assert name == "___ancestor__.Message"
+
+
+def test_import_root_package_from_child():
+    imports = set()
+    name = get_ref_type(package="package.child", imports=imports, source_type="Message")
+
+    assert imports == {"from ... import Message"}
```
nat-n, line 209: This case looks like it follows a different rule to the previous test case with respect to importing the renamed module? Is this avoidable?

betterproto/tests/test_get_ref_type.py
```diff
@@ -77,6 +79,227 @@ def test_importing_google_wrappers_unwraps_them(google_type: str, expected_name:
...
+
+
+def test_import_unrelated_deeply_nested_package():
+    imports = set()
+    name = get_ref_type(
+        package="a.b.c.d", imports=imports, source_type="p.q.r.s.Message"
+    )
+
+    assert imports == {"from .....p.q.r import s as ____p_q_r_s__"}
+    assert name == "____p_q_r_s__.Message"
```
nat-n, line 246: 😧

betterproto/plugin.py
```diff
@@ -358,41 +349,25 @@ def generate_code(request, response):
 
         # Fill response
         f = response.file.add()
-        # print(filename, file=sys.stderr)
-        f.name = filename.replace(".", os.path.sep) + ".py"
+        f.name = filename
```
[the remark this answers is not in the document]
boukeversteegh: in the old implementation, filename contained a sort of package name, like `foo.bar`, and in this line, it was converted to a path.

in the new implementation, filename is already a path, so this transform is not needed.

it made more sense to me decide on the filenames earlier on, not at the very last step, since also the variable is called `filename`

betterproto/plugin.py
```diff
@@ -358,41 +349,25 @@ def generate_code(request, response):
...
-    for base in inits:
-        name = os.path.join(base, "__init__.py")
-
-        if os.path.exists(name):
-            # Never overwrite inits as they may have custom stuff in them.
-            continue
+    # Make each output directory a package with __init__ file
+    output_paths = set(pathlib.Path(path) for path in output_map.keys())
+    output_dirs = [directory for path in output_paths for directory in path.parents]
+    init_files = set(d.joinpath("__init__.py") for d in output_dirs) - output_paths
```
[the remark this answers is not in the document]
boukeversteegh: > I think a single multilevel comprehension is easier to grok than nested intermediate comprehensions.

Great idea! I didn't think of making it multiline like that. I'll give it a go.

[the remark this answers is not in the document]
boukeversteegh: ✔️

betterproto/plugin.py
```diff
@@ -1,15 +1,20 @@
...
+import pathlib
 import re
-import stringcase
 import sys
 import textwrap
 from typing import List
-from betterproto.casing import safe_snake_case
-from betterproto.compile.importing import get_ref_type
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent))  # add betterproto to path
```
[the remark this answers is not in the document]
boukeversteegh: ✔️

betterproto/compile/importing.py
```diff
@@ -1,70 +1,168 @@
...
+def import_root(imports, py_package, py_type):
+    string_import = ".".join(py_package)
+    string_alias = safe_snake_case(string_import)
+    imports.add(f"import {string_import} as {string_alias}")
+    return f"{string_alias}.{py_type}"
+
+
+def import_sibling(py_type):
+    """
+    package =
```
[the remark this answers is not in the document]
boukeversteegh: ah yeah... will fix!

betterproto/compile/importing.py
```diff
@@ -1,70 +1,168 @@
...
+    if importing_google_protobuf and not compiling_google_protobuf:
+        py_package = ["betterproto", "lib"] + py_package
+
+    if py_package[:1] == ["betterproto"]:
+        return import_root(imports, py_package, py_type)
+
+    if py_package == current_package:
+        return import_sibling(py_type)
+
+    if py_package[0 : len(current_package)] == current_package:
```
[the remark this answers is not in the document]
boukeversteegh: ✔️  (black adds a space `: len`)

betterproto/tests/test_get_ref_type.py
boukeversteegh: thanks for your comments.

**test_import_deeply_nested_child_from_root**

This test is about importing a deeply nested package from the context of the top-level package (no package).

```python
def test_import_deeply_nested_child_from_root():
    imports = set()
    name = get_ref_type(
        package="", imports=imports, source_type="deeply.nested.child.Message"
    )

    assert imports == {"from .deeply.nested import child as deeply_nested_child"}
    assert name == "deeply_nested_child.Message"
```

as you can see from the generated import, it imports from 2 levels lower, so they are not siblings.

**import_from_sibling**

when a type is referenced from a 'sibling' (same package), then indeed, an import statement is not necessary.

maybe i should rename `import_*` to `reference_*`, because its about referencing types, that sometimes lead to imports.

agree?

betterproto/tests/test_get_ref_type.py
boukeversteegh: yes! will fix, thanks!

betterproto/tests/test_get_ref_type.py
boukeversteegh: by rename, i also mean the functions in the library code

boukeversteegh: Changes done as a result of these comments:
- replaced method comments with functionally descriptive docstrings
- terminology: `reference` instead of `import`
- added asserts for empty import set cases

I'm not sure what to do about `_from_root`. I understand it can be confused with sys.path.
The context from which the import is done, is the root of the module, of which all packages for the generated code are descendants. It's the `root` from the perspective of the protobuf package hierarchy.

Would be great if we can find a term for that, that is not confusing.

betterproto/tests/test_get_ref_type.py
boukeversteegh: ✔️

betterproto/tests/test_get_ref_type.py
boukeversteegh: ✔️

betterproto/tests/test_get_ref_type.py
boukeversteegh: Somewhat yes and no.

What I tried to do wherever possible was alias the modules, but not the imported types.
Although aliases look ugly, at least the type names are preserved this way.

For ancestor imports, there is no identified package to import that wraps the Type, so you cannot give it an alias. e.g, this is not accepted by python:

```python
from ... import . as ___root__

___root__.Message
```

1. If you know for sure that the ancestor is not the root of the package, you can import 1 level higher up, and alias that. This is what i have done in `reference_ancestor` https://github.com/danielgtaylor/python-betterproto/blob/e2d672a422682cb631bfe772f5c51d0d98b02126/betterproto/compile/importing.py#L126-L135  
  - However this approach is unsafe when the target type is in the top level package, as its not guaranteed that the generated code resides in its own package, so we might end up ascending outside of the package
  - So for the case you mentioned, this approach does not work safely.
2. You could alias the type itself. e.g.
   `from ... import Message as ___Message__`, or  
   `from ... import Message as ___root_Message_`

2 would be a safe and valid solution that prevents name collisions when importing 'Message' from root into a subpackage that also contains 'Message'.

Unfortunately, it makes the name ugly, in that case.

In the future, I hope to improve the importing logic such that it keeps track of occupied names per package, and only creates aliases when necessary, although this will complicate things if in the future we want to support parallel compilation (e.g. multiple protoc processes compiling a single package hierarchy).

boukeversteegh: I think robustness should be prioritized over the prettiness of the names in general. The first goal of betterproto should be that it works. Then, that it generates beautiful code. So if you agree, I will also alias the root types.

betterproto/tests/test_get_ref_type.py
boukeversteegh: Yep 😓 100% ugly, but 100% reliable.

Won't ever create name conflicts. Hopefully we can get rid of it in some more advanced stage of betterproto!

betterproto/compile/importing.py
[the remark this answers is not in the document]
boukeversteegh: ✔️

betterproto/tests/test_inputs.py
boukeversteegh: implicit concatenation? didn't know that! thanks!

betterproto/tests/test_inputs.py
boukeversteegh: hm. i don't see the practical benefit to using tuples here. i find the list more readable as `()` are mostly used for function calls and it takes more staring to figure out `(   (  ) )` than `(   [  ] )`

for performance heavy tasks i'll agree with you though, and start using tuples whenever possible.

betterproto/tests/util.py
boukeversteegh: err nope. it was an attempt to match only dirs, but that didn't work. thanks!
removed~ ✔️

betterproto/tests/util.py
boukeversteegh: Technically, all the directories generated by betterproto will contain `__init__.py`, so it should never crash. But, if anyone were to use this method later on another directory, it will cause a problem. On the other hand, this method serves the current purpose, and not any other (for example, it cannot find modules such as `foo/bar.py`), and I'm not sure I should broaden the scope to support usecases beyond what is currently needed.

boukeversteegh: I've changed the method to add a docstring and to search for init 👍

betterproto/tests/test_casing.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.mark.parametrize(
+    ["value", "expected"],
+    [
+        ("", ""),
+        ("a", "A"),
+        ("foobar", "Foobar"),
+        ("FooBar", "FooBar"),
+        ("foo.bar", "FooBar"),
+        ("foo_bar", "FooBar"),
```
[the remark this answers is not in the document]
boukeversteegh: Hm yes it does create collissions. However, I do not see a practical way to avoid collissions and still convert all other cases to PascalCase, since its not guaranteed the that source strings are even snake_case.

For example:
```protobuf
message MyMessage {

}
```
should generate `MyMessage`

and
```protobuf
message my_message {

}
```
should generate `MyMessage` as well.

If we generate for the last case `My_Message`, we are not really converting anything to PascalCase, and we might as well use the original name.

The same problem goes for the fields.
I think we are generating snake_case there to make the code look nice and pythonic, like we are generating PascalCase for classes to make them nicely pythonic.

In my opinion, two messages or fields with names that are only distinguishable by casing style, is not really a valid use case, but maybe I'm biased..

betterproto/tests/test_casing.py
```diff
@@ -0,0 +1,89 @@
...
+        ("foobar", "foobar"),
+        ("FooBar", "fooBar"),
+        ("foo.bar", "fooBar"),
+        ("foo_bar", "fooBar"),
+        ("FOOBAR", "foobar"),
+        ("FOO_BAR", "fooBar"),
+        ("FOOBAR1", "foobar1"),
+        ("FOOBAR_1", "foobar1"),
+        ("FOO1BAR2", "foo1Bar2"),
+        ("foo__bar", "fooBar"),
```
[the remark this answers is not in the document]
boukeversteegh: other popular libraries like npm camelcase and apache CaseUtils also are not lossless encoders, but rather try to generate a string that matches whatever the casing format prescribes.

I'm inclined to not make this more complicated then necessary, as I don't think anyone in their right mind would make two fields in the same class called `FOOBAR1` and `FOOBAR_1`

boukeversteegh:
> Considering functional change number 🍏 (the first one), is there any scenario that this would cause newly generated code to not be a drop in replacement for packages generated with the existing approach with respect to import statements?

Yes, there is:

- proto files **without package** are now generated in the root of the module instead of the equivalent relative path of the proto file
   - so `protoc -I foo foo/bar/baz.proto` would compile to `bar/baz.py`
   - now it will compile to  `__init__.py`, straight in the root
   - for proto files without packages, it will be best to generate them into a subdirectory, e.g. `--custom_out=lib`

I've updated the docs to reflect this.

betterproto/tests/test_inputs.py
nat-n: Not much practical benefit, it's a bit of an academic point. I think it's good habit to default to tuples in general, though this doesn't seem to be a widespread opinion, and I buy that there's some readability trade off. No big deal.

betterproto/tests/test_casing.py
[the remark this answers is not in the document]
nat-n: I'm gonna push this point a little bit more. The question isn't whether this is the preferred or most elegant behaviour for a case mapping function, but rather how to generate classnames from message names.

Now having a message called `foo_bar ` and one called `FOO__BAR` in the same scope might not be advisable, but some users might have reasons to want to, or might not have a choice about it (if they're just consumers of an API). So I think there's value in supporting as many additional use cases of this unfortunate nature as possible without compromising on the more sane use cases.

I see that this could ruin your elegant regex based solution, though there are a number of alternatives that are more flexible. Admittedly this is one weird cases that can be handled among a larger number that can't. Maybe the decider should be whether other tools that specifically deal with pb message names also choke on this case.

I'll let you decide what to do.

betterproto/tests/test_casing.py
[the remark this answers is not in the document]
nat-n: The argument to be made here is that for users following best practices of message naming it doesn't matter. For those who are, it might matter and they might not have a choice, because they might be consuming an API they don't control.

betterproto/tests/test_get_ref_type.py
nat-n: >    `from ... import Message as ___root_Message_`

Until someone calls a package root 🙃

betterproto/tests/test_casing.py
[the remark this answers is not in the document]
boukeversteegh: I've implemented a `strict` parameter for all case functions, which when `strict=True` will output strings in _strict_ format _(snake_case = single underscores, pascalCase+CamelCase = no underscores)_. When `strict=False`, the delimiter count is preserved. E.g. `camel_case('foo__bar') == 'Foo_Bar'`.

At the moment we cannot use this however, because of nested types.

Protobuf plugin will report Nested.Type as `NestedType`, so while generating the class, its not known that this type was originally nested (and which part of the name is the parent class).

References to this type will be broken if we use `strict=False`, since that would result in a type reference of `Nested_Type`, which can then not be resolved.

I'm not sure how to solve this now, but we could make an attempt at a later time.

betterproto/tests/test_get_ref_type.py
boukeversteegh: Good point 😄 

I've aliased them now as `___Message__`, I don't think adding `root` will help a lot in the end

PR #90, opened 2020-06-12
fixed field types should be int

### LINKED ISSUE #14: fixed/sfixed types convert to float
According to the [protobuf language guide](https://developers.google.com/protocol-buffers/docs/proto3#scalar), `fixednn` and `sfixednn` are integer types. However, when betterproto generates dataclasses it treats them as floats.

```
$ cat foo.proto
syntax = "proto3";
package foo;

message Foo {
  fixed32 bar = 1;
  sfixed32 baz = 2;
  fixed64 qux = 3;
  sfixed64 quux = 4;
}

$ protoc --python_betterproto_out . foo.proto
Writing foo.py
$ cat foo.py
# sources: foo.proto
# plugin: python-betterproto

import betterproto


@dataclass
class Foo(betterproto.Message):
    bar: float = betterproto.fixed32_field(1)
    baz: float = betterproto.sfixed32_field(2)
    qux: float = betterproto.fixed64_field(3)
    quux: float = betterproto.sfixed64_field(4)

```

* betterproto 1.2.2 (from pypi)
* protoc 3.6.1
* python 3.7.3

jameslan:
Fix #14


boukeversteegh:
According to https://stackoverflow.com/questions/7604966/maximum-and-minimum-values-for-ints

ints are unbounded, so supported value ranges for `signed64` vs `int` should not be a consideration to use floats.

could you perhaps update your test to include the highest and lowest value for each of the types?

jameslan:
OK, made them `2**32 - 1`, `-2**31`, `2**64 - 1`, `-2**63` respectively

Issue #49, opened 2020-05-21
Split code up into multiple modules


nat-n:
Currently both the plugin and runtime library as single modules. Maybe there's an element of personal preference here but I think splitting them up into more single purpose modules (and possibly sub-packages) would make the code easier to understand and work with. 

It would also add some flexibility for optimisations such as only importing the parts that are required, e.g. excluding client/server code if not needed.

I'd be happy to work on splitting these modules up if there are no objections (maybe at a time when there are fewer open PRs). @danielgtaylor

boukeversteegh:
@nat-n and I had a meeting today and arrived at the following plan:
- abort PR #47 because it is untested and changes a lot at once
- merge #51 first, to help refactor more safely
- implement some of #47's refactoring for `plugin.py`
  - splitting up code into somewhat smaller functions
  - since I've already refactored it before, I can easily redo it without the fixes.
  - it will help @nat-n to modularize
- refactor into modules
- creating separate PR's for the bugs discovered in #47, with testcases included

Please let us know your thoughts @danielgtaylor @cetanu . Would be great to have a call if possible, too many thoughts went into it that would be tedious to write.

boukeversteegh:
hey @nat-n 😃 

In attempt to get a better overview of the internal dependencies of objects inside `__init__.py`, I have  organized the objects into functional and logical groups, and identified the dependencies between those groups.

If you'd like to split up the init-file as part of this issue, the following overview will probably come in handy.

- `base` - Base-classes for compiled output, public (depends on runtime, proto3)
	- `Message`
	- `Enum`
	- `FieldMetadata`
	- `ServiceStub`
	- `T` `dataclass_field` `_PLACEHOLDER` `PLACEHOLDER` `ProtoClassMetadata`

- `util` - part of public api (depends on base)
	- `Casing`
	- `serialized_on_wire`
	- `which_one_of`
- `proto3` - specs of proto3 format (no dependencies)
	- `FIXED_TYPES` `INT_64_TYPES` `PACKED_TYPES`
[...]

PR #91, opened 2020-06-12
Support running plugin without installing betterproto

boukeversteegh:
Let's us run protoc with `betterproto/plugin.py` , without installing betterproto

```bash
git clone git@github.com:danielgtaylor/python-betterproto.git betterproto-script-demo
cd betterproto-script-demo
pipenv install --ignore-pipfile
mkdir example
pipenv run protoc \
  --plugin=protoc-gen-custom=betterproto/plugin.py \
  --custom_out=example \
  betterproto/tests/inputs/bool/bool.proto
```

**Before**:

```yaml
  File "betterproto/plugin.py", line 10, in <module>
    from betterproto.casing import safe_snake_case
ModuleNotFoundError: No module named 'betterproto'
--custom_out: protoc-gen-custom: Plugin failed with status code 1.
```

Here you needed to install `betterproto` so it can reference itself when run from as plugin (script)

**After**

```
Writing __init__.py
Writing betterproto/tests/inputs/__init__.py
Writing betterproto/tests/inputs/bool/__init__.py
Writing betterproto/tests/inputs/bool/bool.py
```

### DIFF
--- a/betterproto/plugin.py
+++ b/betterproto/plugin.py
@@ -2,14 +2,17 @@
 
 import itertools
 import os.path
+import pathlib
 import re
 import stringcase
 import sys
 import textwrap
 from typing import List
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent))  # add betterproto to path
+import betterproto
 from betterproto.casing import safe_snake_case
 from betterproto.compile.importing import get_ref_type
-import betterproto
 
 try:
     # betterproto[compiler] specific dependencies


nat-n:
I have mixed feelings about sys.path hackery.

Also not sure about the value of supporting such a use-case.

That said, it's a nominal improvement with quite limited footprint, assuming no-one has reason to import this module into another script. 

I'd feel better about it if the script were split up into multiple module with a minimal entry-point (maybe as part of #49 after the current round of PRs are merged), or if it attempted to work without path jiggery, and then falls back, such as with a try/except around `import betterproto`.

boukeversteegh:
> I have mixed feelings about sys.path hackery.
> Also not sure about the value of supporting such a use-case.

Yeah, I don't like it either. The use case is that as a developer, I want to clone the repo, install the dependencies and be able to run the plugin as i'm writing code, without "reinstalling" betterproto when I change the plugin implementation. Its a bit strange for a module to have a dependency on itself that needs to be installed as a package.


> That said, it's a nominal improvement with quite limited footprint, assuming no-one has reason to import this module into another script.
> 
> I'd feel better about it if the script were split up into multiple module with a minimal entry-point (maybe as part of #49 after the current round of PRs are merged), or if it attempted to work without path jiggery, and then falls back, such as with a try/except around `import betterproto`.

This is actually what I was also thinking of. `betterproto/__init__.py` and `betterproto/plugin.py` should have basically nothing in it, except imports, and possible this path jiggling to make it work with protoc.

Alternatively, we could provide an executable (e.g. `betterproto/plugin.sh`, or even `bin/betterproto` that simply runs `python -m betterproto.plugin`. I've tested this already and that works too.

nat-n:
Regarding the use-case, maybe poetry eases this... I've not tried but I expect the following to work:

```bash
git install ... && cd ...
poetry install  # build virtualenv etc
poetry run protoc ...
```

Something quite similar should also work with pipenv but poetry is generally a little smoother and more obvious about it.

If it's still a pain point after the other pending PRs then it might still be worthwhile.

I kinda like the idea of providing/installing a betterproto executable that wraps protoc for convenience.

boukeversteegh:
No longer needed since when running `poetry shell`, its possible to just run the plugin with `--python_betterproto_out` as poetry generates the script `protoc-gen-python_betterproto`

PR #92, opened 2020-06-12
Support lists of non-protobufs in Message.to_dict

Gobot1234:
Currently if you have a Message with a typing.List field of anything that isn't a Message, it will fail to convert due to it raising an AttributeError.

This can be fixed with a simple instance check before appending to the list.

### DIFF
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -818,7 +818,13 @@ class Message(ABC):
                         output[cased_name] = v
                 elif isinstance(v, list):
                     # Convert each item.
-                    v = [i.to_dict(casing, include_default_values) for i in v]
+                    new = v
+                    for i in v:
+                        if isinstance(i, Message):
+                            new.append(i.to_dict(casing, include_default_values))
+                        else:
+                            new.append(i)
+                    v = new
                     if v or include_default_values:
                         output[cased_name] = v
                 else:
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -818,13 +818,10 @@ class Message(ABC):
                         output[cased_name] = v
                 elif isinstance(v, list):
                     # Convert each item.
-                    new = v
-                    for i in v:
-                        if isinstance(i, Message):
-                            new.append(i.to_dict(casing, include_default_values))
-                        else:
-                            new.append(i)
-                    v = new
+                    try:
+                        v = [i.to_dict(casing, include_default_values) for i in v]
+                    except AttributeError:
+                        pass
                     if v or include_default_values:
                         output[cased_name] = v
                 else:


boukeversteegh:
Thanks for your PR!

Could you expand a little bit on the use case for this?

Gobot1234:
Running this code snippet
```py
message = b64decode('lwAAgDYAAAAJwkwnEQEAEAEQo+O7hgRRFwAAAAAAAABiHENoYXRSb29tLkdldE1lc3NhZ2VIaXN0b3J5IzEI+MXXBhDLvecVGP////8HIAAoADAAOGQ=')
emsg_value = struct.unpack_from("<I", message)[0]
emsg = EMsg(steam.utils.clear_proto_bit(emsg_value))
msg = MsgProto(emsg, message)

>>> repr(msg)
```

Would raise
```py
  File "<input>", line 1, in <module>
  File "/Users/James/PycharmProjects/steam.py/steam/protobufs/__init__.py", line 172, in __repr__
    resolved.extend([f'{k}={v!r}' for k, v in self.body.to_dict(betterproto.Casing.SNAKE).items()])
  File "/Users/James/PycharmProjects/steam.py/venv/lib/python3.7/site-packages/betterproto/__init__.py", line 807, in to_dict
    v = [i.to_dict(casing, include_default_values) for i in v]
  File "/Users/James/PycharmProjects/steam.py/venv/lib/python3.7/site-packages/betterproto/__init__.py", line 807, in <listcomp>
    v = [i.to_dict(casing, include_default_values) for i in v]
AttributeError: 'int' object has no attribute 'to_dict'
```

I assume this is due to the list of messages being an `int`, so this PR fixes this issue. This makes `repr(msg)` return a proper value for messages in my case.
```py
<MsgProto msg=<EMsg.ServiceMethodCallFromClient: 151> _header=<MsgHdrProtoBuf msg=<EMsg.ServiceMethodCallFromClient: 151> steamid='76561198248053954' client_sessionid=1087304099 jobid_source='23' target_job_name='ChatRoom.GetMessageHistory#1'> messages=[14017272]>
```

Source for `MsgProto` can be found here https://github.com/Gobot1234/steam.py/blob/connect/steam/protobufs/__init__.py#L140-L175.

nat-n:
Thanks for contributing this PR @Gobot1234 

Before merging we should put in place a test case for the issue this change fixes following [the established pattern](https://github.com/danielgtaylor/python-betterproto/tree/master/betterproto/tests)

Would you be will to have a go at this James?

Gobot1234:
I can give it a go for sure 👍

boukeversteegh:
Thanks for clarifying @Gobot1234 ~

Then does it mean you have received a corrupt message over the wire? Because if that field should contain a list of messages, but it contains an `int` instead of a message, that seems like something went wrong on the server, and its not a bug in the library.

Because the example you gave runs mostly non-betterproto code, its not so clear how we ended up with a betterproto.Message that contains invalid data. In general, I suppose that we want some guarantee that betterproto.Message is internally correct. If some bad data entered through one of the public methods, then that's an issue in betterproto, and I suggest then that the input is rejected early on (for example in `parse`, or `from_dict`).

But if the Message was somehow injected with incorrect data, bypassing the betterproto library, then I think this may not be something we want to support.

Let me know if I made some incorrect assumptions or misunderstood the situation.

Gobot1234:
I do have to agree thinking about it, checking `msg.body` returns a list of `int`s not `CChatRoom_GetMessageHistory_ResponseChatMessage` as expected.

Thanks you for your time 🤦 .

PR #97, opened 2020-06-24
Test to_dict with missing enum values [fixes]

boukeversteegh:


boukeversteegh:
Another PR with fix is here: #101

PR #100, opened 2020-07-04
Import bug - Circular Dependencies

boukeversteegh:
Protobuf itself does not allow circular dependencies (`import "a.proto"` in `b.proto`, and `import "b.proto"` in `a.proto`).

So why do we still get circular import errors in python? Because protobuf dependencies are on _file_-level, while betterproto dependencies are on _package_-level.

Example:

The proto messages depend on each other in a non-circular way:

    Test -------> RootPackageMessage <--------------.
      `------------------------------------> OtherPackageMessage

Test and RootPackageMessage are in different files, but belong to the same package (root):

    (Test -------> RootPackageMessage) <------------.
      `------------------------------------> OtherPackageMessage

After grouping the packages into single files or modules, a circular dependency is created:

    (root: Test & RootPackageMessage) <-------> (other: OtherPackageMessage)

---

Because we don't know in advance which imports will cause a circular-dependency (as we currently do single-pass compiling), the only solution I see is to always use forward references (`"package.SomeType"`), and have the import statements at the bottom.

This PR implements that solution.

### DIFF
--- a/betterproto/compile/importing.py
+++ b/betterproto/compile/importing.py
@@ -86,7 +86,7 @@ def reference_absolute(imports, py_package, py_type):
     string_import = ".".join(py_package)
     string_alias = safe_snake_case(string_import)
     imports.add(f"import {string_import} as {string_alias}")
-    return f"{string_alias}.{py_type}"
+    return f'"{string_alias}.{py_type}"'
 
 
 def reference_sibling(py_type: str) -> str:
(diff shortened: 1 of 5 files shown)


betterproto/compile/importing.py
```diff
@@ -86,7 +86,7 @@ def reference_absolute(imports, py_package, py_type):
     string_import = ".".join(py_package)
     string_alias = safe_snake_case(string_import)
     imports.add(f"import {string_import} as {string_alias}")
-    return f"{string_alias}.{py_type}"
+    return f'"{string_alias}.{py_type}"'
```
nat-n, line 89: what are the extra quotes for?

boukeversteegh: These make the typehints come out as

```python
class Example(betterproto.Message):
   foo: "__some_package__Message" = ...
```

It creates a forward reference that will allow referencing objects that are not imported yet.

See: https://www.python.org/dev/peps/pep-0484/#forward-references

PR #102, opened 2020-07-04
Fix: to_dict returns wrong enum fields when numbering is not consecutive

### LINKED ISSUE #93: to_dict returns wrong enum fields when numbering is not consecutive
Protobuf spec file like this:

```
syntax = "proto3";
package message;
message TeraMessage {
	int64 timestamp = 1;
	enum MessageType {
		MESSAGE_TYPE_UNKNOWN = 0;
		MESSAGE_TYPE_ACTION_MESSAGE = 1;
		// @exclude MESSAGE_TYPE_COMMAND_MESSAGE = 2; // DEPRECATED
		MESSAGE_TYPE_CONFIG_MESSAGE = 3;
		MESSAGE_TYPE_HEARTBEAT_MESSAGE = 4;
	}
	MessageType message_type = 5;
	bytes message = 6;
}
```

Generates the following Python bindings:

```python
import betterproto

class TeraMessageMessageType(betterproto.Enum):
    MESSAGE_TYPE_UNKNOWN = 0
    MESSAGE_TYPE_ACTION_MESSAGE = 1
    # NB: notice that 2 is missing
    MESSAGE_TYPE_CONFIG_MESSAGE = 3
    MESSAGE_TYPE_HEARTBEAT_MESSAGE = 4

@dataclass
class TeraMessage(betterproto.Message):
    timestamp: int = betterproto.int64_field(1)
    message_type: "TeraMessageMessageType" = betterproto.enum_field(5)
    message: bytes = betterproto.bytes_field(6)
```

To reproduce the bug:
```
>>> from my.path.message import TeraMessage, TeraMessageMessageType
>>> message = TeraMessage(message_type=TeraMessageMessageType.MESSAGE_TYPE_CONFIG_MESSAGE)
>>> message.to_dict()
{'messageType': 'MESSAGE_TYPE_HEARTBEAT_MESSAGE'}
>>> message.to_json()
'{"messageType": "MESSAGE_TYPE_HEARTBEAT_MESSAGE"}'
>>> TeraMessage().parse(bytes(message)).message_type == TeraMessageMessageType.MESSAGE_TYPE_CONFIG_MESSAGE
True
```
```

boukeversteegh:
closes #93

- When converting to dictionary, get enum values by `Enum()` constructor, instead of index.
- Small refactor: determine if a field is a list, not by its value, but by its type. (i.e. do not rely on the user to tell us what type our fields are).
- Small refactor: removal of magic constants that were proto3 type names.
- Added more tests for enum.


bencwallace:
Thanks for this bug fix! I just ran into this myself and it took me a while to realize the fix was available in version 2. Any when this will come out of beta?

Gobot1234:
> Thanks for this bug fix! I just ran into this myself and it took me a while to realize the fix was available in version 2. Any when this will come out of beta?

No but if you want this fix install betterproto==2.0.0b2

PR #103, opened 2020-07-05
Fix - No arguments are generated for stub methods when using `import` with proto definition

### LINKED ISSUE #23: Import Bug - No arguments are generated for stub methods when using `import` with proto definition
If I'm using the `import` in the proto file, like the following:

https://github.com/huan/python-betterproto-issue/blob/60d232c90c099abca4aea6dc1273121cd4f43d66/proto_with_import/puppet_service.proto#L1-L9

Then the betterproto will fail to generate the argument of the `Ding` method, like this:

https://github.com/huan/python-betterproto-issue/blob/60d232c90c099abca4aea6dc1273121cd4f43d66/gen_proto_with_import/puppet_grpc.py#L13

As we are expecting this:

https://github.com/huan/python-betterproto-issue/blob/60d232c90c099abca4aea6dc1273121cd4f43d66/gen_proto_without_import/puppet_grpc.py#L21

Reproduce Repo at <https://github.com/huan/python-betterproto-issue>

boukeversteegh:
Fixes #23 

Due to too much juggling with different branches, and an incorrect test-case, I mistakenly thought this bug was already fixed earlier.

The bug:

- Define a Service with an rpc method that has a Message argument
- Betterproto will unpack the message fields, and add them as arguments directly to the generated client method
- While compiling, we need to lookup the correct message that acts as the rpc method input argument
- If the Service was compiled before the Message, it cannot be found, and input arguments are skipped.
- The algorithm to match the Service Method Input type with the correct message was also shaky and did not work at all when the arguments were nested or in packages

The solution

- First read and compile all the Messages
- Then compile all the Services
- Match the method input types more precisely by keeping track of the parsed message' packages

Additionally:

- It was very hard to debug `plugin.py` as it is called by `protoc`, so attaching a debugger involves freezing the plugin when it starts, then finding the process id, then attaching, and waiting till the freeze is completed.
   - To make life better, I've added a way to write the input provided by `protoc` to a file, and then run `plugin.py` directly with that input (stdin). Now you can debug the plugin with your IDE 💪


betterproto/plugin.py
```diff
@@ -373,8 +216,207 @@ def generate_code(request, response):
         init = response.file.add()
         init.name = str(init_file)
 
-    for filename in sorted(output_paths.union(init_files)):
-        print(f"Writing {filename}", file=sys.stderr)
+    for output_package_name in sorted(output_paths.union(init_files)):
```
nat-n, line 219: this diff sucks

betterproto/plugin.py
```diff
@@ -373,8 +216,207 @@ def generate_code(request, response):
...
+        print(f"Writing {output_package_name}", file=sys.stderr)
+
+
+def lookup_method_input_type(method, types):
```
nat-n, line 223: could move this method closer to where it's used (swap with read_protobuf_type)

betterproto/plugin.py
boukeversteegh: yeah, it went pretty bad after splitting up code into methods and renaming a bunch of vars :-/

should happen less and less as we get the code more organized

adriangb:
Just a quick note about debugging: I had good success using VSCode with [ptsvd](https://pypi.org/project/ptvsd/).

In `plugin.py`:
```python3
def main():

    import ptvsd
    ptvsd.enable_attach()
    ptvsd.wait_for_attach()  # blocks execution until debugger is attached
```

And set breakpoints in VSCode wherever you want to drop into.

Then in `launch.json`:
```json
{
    "version": "0.2.0",
    "configurations": [

        {
            "name": "Python: Current File",
            "type": "python",
            "request": "launch",
            "program": "${file}",
            "console": "integratedTerminal",
            "justMyCode": false,
        },
        {
            "name": "Attach (Remote Debug)",
            "type": "python",
            "request": "attach",
            "localRoot": "${workspaceRoot}",
            "remoteRoot": "${workspaceRoot}",
            "port": 5678,
            "secret": "",
            "host":"0.0.0.0",
            "justMyCode": false,
        },
    ]
}
```

Now running _anything_ that calls `plugin.py` will just wait until you attach the VSCode debugger before continuing (and stopping at your breakpoints).

boukeversteegh:
> Just a quick note about debugging: I had good success using VSCode with [ptsvd](https://pypi.org/project/ptvsd/).

Awesome tip! I took a note when you mentioned it on slack, and also found that [PyCharm has something similar](https://www.jetbrains.com/help/pycharm/remote-debugging-with-product.html#remote-debug-config) for those who use it. Haven't tried it yet tho!

PR #104, opened 2020-07-07
Add missing async/await keywords when casing

abn:
Simple change to extend list of python keywords to include `async` and `await`. This was causing failures in `make generate` when `black` attempted to format generated file for `namespace_keywords`.


boukeversteegh:
Awesome! Thanks :-D

PR #105, opened 2020-07-07
Handle mutable default arguments cleanly

abn:
When generating code, ensure that default list/dict arguments are initialised in local scope if unspecified or `None`.

The following diff shows the difference in the generated source file from `service.proto`.
```diff
5c5
< from typing import AsyncIterable, AsyncIterator, Iterable, List, Union
---
> from typing import AsyncIterable, AsyncIterator, Iterable, List, Optional, Union
35c35
<         self, *, name: str = "", comments: List[str] = []
---
>         self, *, name: str = "", comments: Optional[List[str]] = None
36a37
>         comments = comments or []
```

Due to the nature of the issue, the included test simply verifies that the default list instance initialised is not the same object used in any subsequent calls. This is done so by intercepting the calls made to the underlying `ServiceStub._unary_unary` method and verifying that the `comments` instance is not reusing the object.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -476,6 +476,20 @@ pytest = ">=4.6"
 testing = ["fields", "hunter", "process-tests (2.0.2)", "six", "pytest-xdist", "virtualenv"]
 
 [[package]]
+category = "dev"
+description = "Thin-wrapper around the mock package for easier use with pytest"
+name = "pytest-mock"
+optional = false
+python-versions = ">=3.5"
+version = "3.1.1"
+
+[package.dependencies]
+pytest = ">=2.7"
+
+[package.extras]
+dev = ["pre-commit", "tox", "pytest-asyncio"]
+
+[[package]]
 category = "main"
 description = "Alternative regular expression module, to replace re."
 name = "regex"
@@ -623,7 +637,7 @@ testing = ["jaraco.itertools", "func-timeout"]
 compiler = ["black", "jinja2", "protobuf"]
 
 [metadata]
-content-hash = "375411698ec644810d09af809ac8a004abc9821f0b718178505424625f407c14"
+content-hash = "7a2aa57a3a2b58e70aa05be75bff08272c2f070ecc1872a1c825f228299b82c2"
 python-versions = "^3.6"
 
 [metadata.files]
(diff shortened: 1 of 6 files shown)


nat-n:
Nice. Maybe worth having a test case for this? I assume you've verified it manually?

abn:
@nat-n I did test it manually. Although turns out I did not handle the edge case when `Optional` was never imported. Had to change tact to modify the code instead of the template. Added a test, although had to make use of `pytest-mock` since the affected code is in the generated files. This should now be good to go.

betterproto/plugin.py
```diff
@@ -314,6 +314,24 @@ def generate_code(request, response):
...
+                        for f in input_message.get("properties"):
+                            if f["zero"] != "None" and (
+                                f["type"].startswith("List[")
+                                or f["type"].startswith("Dict[")
+                            ):
+                                output["typing_imports"].add("Optional")
+                                mutable_default_args.append((f["py_name"], f["zero"]))
+                                f["zero"] = "None"
+                            properties.append(f)
+                        input_message["properties"] = properties
```
boukeversteegh, line 333: I'm a bit nervous to see the entire `properties` field being overridden, as it makes it not immediately clear if it is changed fundamentally or not. Since the values are dictionaries (objects), it should be possible to just modify them, instead of recreating the list?

abn: Good call. Fixed it.

betterproto/tests/grpc/test_grpclib_client.py
```diff
@@ -1,14 +1,18 @@
 import asyncio
+import sys
+
+import grpclib
+import grpclib.metadata
```
nat-n, line 5: redundant import?

boukeversteegh:
Ah, sorry it took me so long, but i finally get it.
The diff in the example was just a tad too short to get the context..

So:

- This is about the arguments of methods of ServiceStubs (aka Clients)
- When a Service accepts a message that contains a `repeated` field, a parameter was created with a default `[]`
- If the caller does not specify the parameter, the default parameter is used
- If that parameter is then modified by the ServiceStub, it will contain the modifications upon the next call.
- This would cause unexpected and hard to debug behavior

Correct?

What I still don't understand is:

- In what way might the default parameter accidentally become modified, and leak into the next call?

From what I see, the request object created on the fly, and then passed to grpcio for serialization and sending. In theory it should not ever be modified.

abn:
@boukeversteegh that is correct. The change is to ensure generated code adheres to best practices. Since the project aims to generate clean pythonic code at best effort, I think this is a  change in that direction. 

I am not across all possible scenarios, present and future, of how this code gets used to say for certain that the default value (initialised at definition) will never be mutated prior to serialisation. This is more a defense-in-depth change.

betterproto/tests/grpc/test_grpclib_client.py
abn: Intentional. Without it, inspection will complain. Think `__all__` is defined in `grpclib/__init__.py`.

nat-n: right, though I didn't see anything using grpclib.metadata specifically

abn: It is used by existing test cases. See L59 for example below.

https://github.com/danielgtaylor/python-betterproto/blob/0cf8aaab6bddbdebcb0602b4ca36428d4cf70a5a/betterproto/tests/grpc/test_grpclib_client.py#L56-L124

abn:
@boukeversteegh this should be ready for review again. The recent refactor in master makes the implementation here a bit cleaner.

abn:
@boukeversteegh @nat-n rebased and ready again; had to rebased due to improt statement collision.

PR #107, opened 2020-07-08
Replace dependency on protoc with grpcio-tools

abn:
This change removes the dependency on platform provided protobuf tools in favour of `grpcio-tools` dependency. This makes both development and compiler use independent from platform dependencies.

This also could potentially allow for a `betterproto` cli tool. I have added the dependency to the `compiler` extra here, but this is not strictly required as currently this is required only for testing when generating test source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,6 @@ jobs:
             ${{ runner.os }}-poetry-
       - name: Install dependencies
         run: |
-          sudo apt install protobuf-compiler libprotobuf-dev
           poetry install
       - name: Run tests
         run: |
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ clean:              ## - Clean out generated files from the workspace
 o=output
 plugin:             ## - Execute the protoc plugin, with output write to `output` or the value passed to `-o`
 	mkdir -p $(o)
-	protoc --plugin=protoc-gen-custom=betterproto/plugin.py $(i) --custom_out=$(o)
+	poetry run python -m grpc.tools.protoc $(i) --python_betterproto_out=$(o)
 
 # CI tasks
 
--- a/README.md
+++ b/README.md
@@ -310,9 +310,6 @@ Join us on [Slack](https://join.slack.com/t/betterproto/shared_invite/zt-f0n0uol
 
 - Python (3.6 or higher)
 
-- [protoc](https://grpc.io/docs/protoc-installation/) (3.12 or higher)
-  *Needed to compile `.proto` files and run the tests*
-
 - [poetry](https://python-poetry.org/docs/#installation)
   *Needed to install dependencies in a virtual environment*
 
--- a/betterproto/tests/generate.py
+++ b/betterproto/tests/generate.py
@@ -11,8 +11,7 @@ from betterproto.tests.util import (
     inputs_path,
     output_path_betterproto,
     output_path_reference,
-    protoc_plugin,
-    protoc_reference,
+    protoc,
 )
 
 # Force pure-python implementation instead of C++, otherwise imports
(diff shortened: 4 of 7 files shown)


nat-n:
Nice. I didn't know about grcp-tools. If we start using betterproto models in the compiler (which we really should) then maybe we don't need to depend on protobuf at all!

Unless the plugin actually uses grpc-tools I don't think it should be listed as a dependency. The user can install it if they want but shouldn't be forced to unnecessarily.

boukeversteegh:
> If we start using betterproto models in the compiler (which we really should) then maybe we don't need to depend on protobuf at all!

Yes indeed! With this PR we only depend on it for testing, so I think it can be moved to dev-dependencies.

abn:
Yeah, I wasn't sure about that. Will drop the dependency from the compiler extra. Should the extra be called "plugin" instead of compiler? :)

aselus-hub:
will it still work with protoc as a plugin w/o the grpcio dep with this change? 
those of us w/ multilanguage build pipelines are very negatively affected by dependancy to  grpcio-tools

abn:
@aselus-hub no changes at all for users of `protoc` and this change does not impact functionality. This is done so that we can generate test files without having to install protobuf compiler and library out of band from the package's toolchain.

boukeversteegh:
> will it still work with protoc as a plugin w/o the grpcio dep with this change?
> those of us w/ multilanguage build pipelines are very negatively affected by dependancy to grpcio-tools

If i understand correctly, the dependency is only there for _development_. So if your pipeline is just installing `betterproto` (including `betterproto[compile]`), `grpcio-tools` will not be installed.

`plugin.py` does not reference `grpcio-tools`, so when you are running `protoc --.....` with the plugin, you'll have nothing to do with `grpcio.-tools`.

boukeversteegh:
> Yeah, I wasn't sure about that. Will drop the dependency from the compiler extra. Should the extra be called "plugin" instead of compiler? :)

Perhaps that makes a bit more sense yes. From the beginning I was confused what the `[compiler]` package was for. It took me some time to realize this library is needed for both _generating_ and for _referencing_ the python files, and that some people might want to reduce their runtime dependencies by excluding the plugin part.

From that respect, `betterproto[protoc-plugin] ` would be even more clear. At least I would understand that immediately.

This also exposes a problem I hadn't realized before. The plugin is actually installed, even without the `[compile]` extras. This is because the script that `protoc` needs is defined in the `project.toml` (`protoc-gen-python_betterproto`), but not as part of the extras. If you then run the plugin, you'll get _missing modules_ if you're lucky, or a `exit code 1`, if you're unlucky, with no failure message from `protoc`.

I'm not sure if its possible to move the script section into the extras, besides creating two different builds/packages.

abn:
> From that respect, `betterproto[protoc-plugin] ` would be even more clear. At least I would understand that immediately.

Maybe another PR? 2.0 might be a good version to make that change in. 

> 
> I'm not sure if its possible to move the script section into the extras, besides creating two different builds/packages.

At the moment no. Although, this is part of an improvement coming in a future poetry version. 

We can mitigate this issue by printing out some useful error message like "please install compiler extras using pip install betterproto[compiler]" when an import error occurs and calling exit. I can create another PR for this.

boukeversteegh:
> > From that respect, `betterproto[protoc-plugin] ` would be even more clear. At least I would understand that immediately.
> 
> Maybe another PR? 2.0 might be a good version to make that change in.

Sure!
 

> We can mitigate this issue by printing out some useful error message like "please install compiler extras using pip install betterproto[compiler]" when an import error occurs and calling exit. I can create another PR for this.

Actually we have that, oops. But I remember the message doesn't show nicely if you run protoc with the plugin. Something to retest, now i'm not sure.

boukeversteegh:
@aselus-hub do above answers take away your concerns? If yes, we can merge.

aselus-hub:
yup no concerns from me thank you for the check @boukeversteegh

PR #108, opened 2020-07-08
Use poetry-core as PEP 517 build backend

abn:
This change replaces the use of poetry as the build backend in favour
of the leaner poetry-core. This speeds up PEP-517 builds for source
installs, tox environment setup etc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,5 +77,5 @@ commands =
 """
 
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0,<2"]
+build-backend = "poetry.core.masonry.api"


nat-n:
Interesting, is this change otherwise transparent to the user?

boukeversteegh:
Without any documentation on `poetry-core` and just 15 stars, I feel its a bit too obscure to switch to a different package.

If it builds faster, does it mean only our packaging will speed up, or also `poetry install`? If its only packaging, I'm inclined to keep things as they are, as we use a well-documented build system and we have no issues with performance at the moment.

abn:
@nat-n yes it is.

@boukeversteegh this is part of the changes coming with poetry's `1.1.x` release. Full disclosure, I am one of the maintainers for `poetry`. With this release, the PEP 517 related logic now will be maintained in `poetry-core` and `poetry` itself depends on `poetry-core`. This is done os because most of the dependencies and packages provided by `poetry` is not required for a PEP 517 build. From a developer experience perspective nothing changes, you still use `poetry` cli to manage dependencies and builds etc. However, for PEP 517 builds like when pip installs from sdists and or scm, this makes things a faster since only `poetry-core`. The noticible difference for a project contributor is when running `tox`. Each time tox creates an isolated build, a PEP 517 build is triggered. And each time currently, `poetry` is installed. With this change it will only install what is required, which is `poetry-core`.

Also note, going forward all default poetry configuration generated will use `poetry-core` and not `poetry` as the build backend.

boukeversteegh:
> @nat-n yes it is.
> 
> @boukeversteegh this is part of the changes coming with poetry's `1.1.x` release. Full disclosure, I am one of the maintainers for `poetry`. With this release, the PEP 517 related logic now will be maintained in `poetry-core` and `poetry` itself depends on `poetry-core`. This is done os because most of the dependencies and packages provided by `poetry` is not required for a PEP 517 build. From a developer experience perspective nothing changes, you still use `poetry` cli to manage dependencies and builds etc. However, for PEP 517 builds like when pip installs from sdists and or scm, this makes things a faster since only `poetry-core`. The noticible difference for a project contributor is when running `tox`. Each time tox creates an isolated build, a PEP 517 build is triggered. And each time currently, `poetry` is installed. With this change it will only install what is required, which is `poetry-core`.
> 
> Also note, going forward all default poetry configuration generated will use `poetry-core` and not `poetry` as the build backend.

Thank you for your explanations @abn ! It sounds like a useful change, and I'd be happy to follow the poetry best-practices, but I do see that `poetry-core` is in still in `1.0.0a8`, pre-release, and changes are being made as we speak.

I'd be more comfortable switching when poetry reaches `1.1`. or `poetry-core` is stable. If we have a strong reason to switch, its different of course, but with all the work to be done, I'd rather not take the risk right now.

Let's keep this one open for a while.

blaisep-vgs:
Dear @boukeversteegh , @nat-n ,
I noticed that Poetry has been at [version 1.0.10](https://github.com/python-poetry/poetry/releases/tag/1.0.10) for a few months now. Would you consider that a threshold of stability to review this PR ?
Thank you.

@abn , thank you for your service!

nat-n:
@blaisep-vgs I think the idea was to switch to using poetry-core directly when it reaches a stable release. It looks like that should be pretty soon ([currently at 1.0.0rc2](https://github.com/python-poetry/poetry-core/releases)).

Otherwise poetry should function normally for this project already, including v1.0.10 if that's what you're using.

abn:
@nat-n `1.0.0` just got released. Updated the PR.

nat-n (approved):
@abn Awesome. Congrats. Let's do this then.

Issue #63, opened 2020-05-24
oneof with enum : which_one_of returns the wrong value when enum is the first field


boukeversteegh:
See commit

Issue #95, opened 2020-06-19
to_dict() fails to write boolean fields without default when value is False


priesgo:
When you have a message with a boolean value as follows:
```
message Patient {
     bool isRnaAvailable = 1;
}
```
then an object like
```
patient = Patient()
patient.is_rna_available = np.random.choice([True, False], 1)[0]
```
fails to write the `is_rna_available` field when it is False when calling `patient.to_dict()`, this does not happen when the value is True.

The issue seems to be in line 824 of file `betterproto/__init__py` where it runs the following comparison `elif v != self._get_field_default(field, meta) or include_default_values:`. It should be something like 'elif v != None or include_default_values', otherwise when you set a field to the same value as the default value and do not set `include_default values` then it is not written. This may be also the case for other types.

boukeversteegh:
I assume you have tried `include_default_values=True`, but it doesn't suite your use case?

If so, could you clarify why this option does not solve your problem?

_Background_

> Protobuf does not distinguish values that are set to the default versus unset values, so _betterproto_ follows the same behavior when calling `to_dict`. To include all values in the output, you can set `include_default_values=True`. This will also output fields not touched by the client, and present their default values.

If you are hinting at a version of `to_dict` that outputs only fields explicitly set by the client, we will need to track set/unset fields, which we currently don't do. A discussion about this was ongoing on slack at `#unset-field-detection`. I've made a new issue for this topic, please refer to #101 if its relevant.

boukeversteegh:
I've looked at how the Dart implementation handles this:

On a client-side generated object, `hasField` returns only true when the field was set by the client, no matter the value.
Values are serialized when set, also when they are serialized with the default values. This is a bit of a surprise. Apparently, optimizing the serialization by omitting default values is not standard in protobuf.

```protobuf
message Position {
    int32 x = 1;
    int32 y = 2;
}
```

```dart
var p = Position();
print (p.writeToBuffer()); // prints []

p = Position()..x=0..y=0;
print (p.writeToBuffer()); // prints [8, 0, 16, 0]
```

priesgo:
Thanks for your answer !

 `include_default_values=True` does not cover my use case. The problem is that to check when there is a value set in order to write it, it does not check for it being none, but instead for it to evaluate to true. In the case of a boolean field this fails to write when false, but other numeric types will also behave similarly by not writing 0 values.

This is important in my opinion as having a field being false or 0 is not the same as not having data for it.

PR #110, opened 2020-07-08
Serialize default values in oneofs when calling to_dict() or to_json()

bradykieffer:
This change is consistent with the official protobuf implementation. If
a default value is set when using a oneof, and then a message is
translated from message -> JSON -> message, the default value is kept in
tact. Also, if no default value is set, they remain null.


boukeversteegh:
Awesome! Thank you for your contribution and investigation! 🐍👍 

So I understand that you're isolating the case for `groups` (`oneof`) and making sure that default values are also serialized there.
This leaves still open the matter of whether to output explicitly set values as per #95 , which we can leave out of scope for now.

I'm curious about two things:

- Does this also fix issue #63 ? Test case [oneof_enum](https://github.com/danielgtaylor/python-betterproto/blob/master/betterproto/tests/inputs/oneof_enum/)
- Have you seen that fixing this for `to_dict` also changes the serialization behavior or `bytes`? Because betterproto currently also doesn't write anything over the wire when the data is default (see [my comment here](https://github.com/danielgtaylor/python-betterproto/issues/95#issuecomment-653810144)), and that may be something we should fix as well (at some point).

bradykieffer:
> This leaves still open the matter of whether to output explicitly set values as per #95 , which we can leave out of scope for now.

Yeah I think that's a good idea to leave out of scope for now. I definitely would personally bias toward either including default fields explicitly or using a type such as `BoolValue` so that `false` is sent by default. Funny thing about this package - we were literally **trying** to make using `BoolValue` less cumbersome internally really great that we can't override  which lead us to finding this library 😄 .

> I'm curious about two things:
> 
>     * Does this also fix issue #63 ? Test case [oneof_enum](https://github.com/danielgtaylor/python-betterproto/blob/master/betterproto/tests/inputs/oneof_enum/)
> 
>     * Have you seen that fixing this for `to_dict` also changes the serialization behavior or `bytes`? Because betterproto currently also doesn't write anything over the wire when the data is default (see [my comment here](https://github.com/danielgtaylor/python-betterproto/issues/95#issuecomment-653810144)), and that may be something we should fix as well (at some point).

I'll check both of these before I open the PR for review, @dhendry actually pointed out that I hadn't covered all of the relevant message types so I went ahead and did so before responding to this. So I'll go ahead and look into those and get back to you!

bradykieffer:
Okay so @boukeversteegh it looks like #65 _should_ be resolved. I just had to update the test cases to reflect proto behavior. Specifically, the fields on the protobufs within those test cases shouldn't ever be null (we'd have to abuse oneofs to make them null... 😄 ). As for changing the behaviour of `bytes`. I actually included the default value `oneof` logic in there too - I think that's what we'd want here as it mimics the behavior in `to_dict`, we will write a oneof field with a default value explicitly now. I don't know if we want to include tracking if the field is set though, I don't think that's how it's done within Google's Protobuf implementation for Python. But, I could be wrong and if I am I think we can raise another issue for that one, I wouldn't mind picking it up at some point 😄

boukeversteegh:
Thank you for your time and efforts to investigate and fix this @bradykieffer !

betterproto/tests/inputs/oneof_enum/test_oneof_enum.py
```diff
@@ -9,34 +9,36 @@
...
     """
     message = Test()
     message.from_json(get_test_case_json_data("oneof_enum", "oneof_enum-enum-0.json"))
-    assert message.move is None
+
+    assert message.move == Move(
+        x=0, y=0
+    )  # Proto3 will default this as there is no null
     assert message.signal == Signal.PASS
     assert betterproto.which_one_of(message, "action") == ("signal", Signal.PASS)
```
boukeversteegh, line 24: Is this change needed to solve the bug? As I understand the bug is in `to_dict` and not `from_dict`..

I think the idea that _proto doesn't have null values_ might actually be somewhat muddled, and incorrect. Because values can be skipped when they are sent over the wire, effectively making them undefined. However, the protobuf libraries will then simply return the default value for those skipped fields. That behavior is indeed part of protobuf specification, but for _oneof_ groups, I consider it not part of the spec that all other fields should have a default value as well (unless its actually wrtten in the spec, sorry I didn't have time to read it 😅 ). At least it doesn't make much sense to me why it would be needed.

Unless it is somehow needed for this bugfix, I would suggest to exclude that part from this PR, because its a functional change (and not sure if its an improvement). And that is a bit harder to judge than the bugfix, which we could merge quickly.

betterproto/tests/inputs/oneof_default_value_serialization/test_oneof_default_value_serialization.py
```diff
@@ -0,0 +1,74 @@
...
+                tzinfo=datetime.timezone.utc,
+            )
+        ),
+        Test(duration_value=datetime.timedelta(0)),
+        Test(wrapped_message_value=Message(value=0)),
+        # NOTE: Do NOT use betterproto.BoolValue here, it will cause JSON serialization
+        # errors.
+        # TODO: Do we want to allow use of BoolValue directly within a wrapped field or
+        # should we simply hard fail here?
+        Test(wrapped_bool_value=False),
```
boukeversteegh, line 43: Good point!

It seems odd that passing a `BoolValue` to a field that is defined as such will break the library.
Of course, it was meant as a convenience that the BoolValue is not required, but breaking it goes a bit far.

It does raise some questions.
Do we always store a `bool` internally? Then what happens if you store a BoolValue, then read it again.. it becomes a `bool`. That may be confusing, and potentially cause bugs.

We may need to look into this, in a separate issue.

Its fine with me to keep the TODO

betterproto/__init__.py
```diff
@@ -866,7 +900,9 @@ def from_dict(self: T, value: dict) -> T:
                     elif meta.wraps:
                         setattr(self, field_name, value[key])
                     else:
-                        v.from_dict(value[key])
+                        v = v.from_dict(value[key])
+                        if v is not None:
+                            setattr(self, field_name, v)
```
boukeversteegh, line 905: I'm a bit confused how `from_dict` ever worked properly, if message fields were not assigned to their parent before..

any idea of why this was needed, or why `from_dict` did actually work?

Also, lines `930`-`931` assign `v` to the field. Perhaps just assigning `v=` would suffice?

boukeversteegh: Ah, I realized it!

`v` is an existing Message object. The required behavior is that the data from the map is _merged_ into the existing fields. I think no assignment is needed here.

betterproto/tests/inputs/oneof_enum/test_oneof_enum.py
```diff
@@ -9,34 +9,36 @@
...
     assert betterproto.which_one_of(message, "action") == ("signal", Signal.RESIGN)
 
 
-@pytest.mark.xfail
 def test_which_one_of_returns_second_field_when_set():
     message = Test()
     message.from_json(get_test_case_json_data("oneof_enum"))
     assert message.move == Move(x=2, y=3)
-    assert message.signal == 0
+    assert message.signal == Signal.PASS
```
boukeversteegh, line 43: thanks 👍

betterproto/tests/test_features.py
```diff
@@ -263,3 +263,14 @@ class TestParentMessage(betterproto.Message):
...
+
+
+def test_oneof_default_value_set_causes_writes_wire():
+    @dataclass
+    class Foo(betterproto.Message):
+        bar: int = betterproto.int32_field(1, group="group1")
+        baz: str = betterproto.string_field(2, group="group1")
+
+    assert bytes(Foo(bar=0)) != b""
+    assert bytes(Foo(baz="")) == b""  # Baz is just an empty string
```
boukeversteegh, line 275: Does this mean its not possible to detect `which_one_of` if one of the groups is a `String` and the value was the empty string?

Same with other default values? 😕

betterproto/tests/inputs/oneof_enum/test_oneof_enum.py
bradykieffer: So this is actually default behavior in Google's protobuf implementation as well. What's actually happening here is that the `move` field is never itself null but it isn't actually sent over the wire - calling to_dict() _should_ only show `signal` there. I think what's confusing is that a `oneof` field can have nothing set, an attribute set to a default value, OR an attrbiute set to a non-default value. But, on the actual message object, the attributes themselves are never actually null, they're always in a default state. I replicated this behaviour using these two messages:
```
message Point{
    int64 x = 1;
    int64 y = 2;
}

message Test{
    oneof test_value {
        Point point = 1;
        google.protobuf.BoolValue bool_value = 2;
        int64 int_value = 3;
    }
}
```

Compiling with the Google proto impl I then checked the following:
```python
In [4]: t = Test()

In [5]: t
Out[5]:

In [6]: t.point
Out[6]:

In [7]: t.point.x
Out[7]: 0
...

In [9]: type(t.point.x)
Out[9]: int

In [10]: type(t.point)
Out[10]: foo.Point

In [11]: type(t.int_value)
Out[11]: int

In [12]: t.bool_value
Out[12]:

In [13]: type(t.bool_value)
Out[13]: google.protobuf.wrappers_pb2.BoolValue

In [14]: from google.protobuf.json_format import MessageToDict

In [15]: MessageToDict(t)
Out[15]: {}

In [16]: t.point.x = 1

In [17]: MessageToDict(t)
Out[17]: {'point': {'x': '1'}}

In [18]: t.int_value = 1

In [19]: MessageToDict(t)
Out[19]: {'intValue': '1'}

In [20]: t.bool_value.value = True

In [21]: t
Out[21]:
bool_value {
  value: true
}

In [22]: MessageToDict(t)
Out[22]: {'boolValue': True}
```

I can either add an assert to this test case or a separate test case demonstrating that betterprotos has similar behaviour

betterproto/tests/inputs/oneof_default_value_serialization/test_oneof_default_value_serialization.py
bradykieffer: Yeah if I had to make a choice (with just thinking about this briefly) I would err on the side of failing fast when attempting to assign a non-primitive to a wrapped field. That at least makes the usage of the field unambiguous

betterproto/__init__.py
bradykieffer: Yeah correct, the `v = v.from_dict(value[key])` isn't needed, I'm just used to transforming objects vs mutating them that I wrote it like that! Also, it looks like I can just move lines `930`-`931` out an indentation layer and achieve the same effect 😄

betterproto/tests/test_features.py
bradykieffer: Good catch - this actually works with the other default values but not for a string at the moment, I'll get a fix in 😄

bradykieffer:
Hey @boukeversteegh I just pulled in upstream changes, this branch should be ready for another pass any time!

dhendry:
@boukeversteegh @danielgtaylor it would be amazing if we could get this in sooner rather than later. 

I want to live in the brave new world of betterproto and stop having the official protobuf code gen slowly sap my will to live.

boukeversteegh:
Hey @bradykieffer! Thanks so much for keeping the branch in sync. Your PR is first on the list (obviously) now. Unfortunately I need some time to further review, and I'll be busy till after this weekend. Hope you could wait a bit, or perhaps @nat-n could have a look?

What I wanted to confirm before merging is that betterproto behaves (still, or more) correctly (according to specs), intuitively and consistently. Personally speaking, as long as we are compatible with protocol and other clients, we don't necessarily need to follow other clients behavior precisely. Hope to come back to this soon!

bradykieffer:
Hey @nat-n & @boukeversteegh, just pinging again on this PR, we'd really love to start using betterproto internally!

boukeversteegh:
I currently don't have time to really formalize the way betterproto should behave, and this PR seems to fix some bugs and bad behavior, so I'm going to merge it.

At some point though, we should really specify clearly how the library handles missing values, default values etcetera, and evaluate if it is logical, intuitive and consistent with specs.

Thanks you for your work and especially patience!

PR #113, opened 2020-07-10
Change source layout to use src/ and tests/

abn:
This change does not have any functional impact to contributors nor package users. This as discussed on slack allows for a cleaner source and test separation. Additionally, it prevents some
potential side-effects when testing and packaging code.


nat-n:
Moving tests out of the package is obvious, not so obvious to me what the actual benefit is of putting the package under src?

Is it just to get it out of the way of accidental sys.path inclusion?

abn:
> Is it just to get it out of the way of accidental sys.path inclusion?

That is one of the benefits. A better write up than I can articulate is available [here](https://hynek.me/articles/testing-packaging/).

abn:
@nat-n can I please get this approved again?  Had to rebase with master.

PR #115, opened 2020-07-11
Convert inited values to corresponding Enum members

Gobot1234:
Converts any values passed to their Enum counterparts.

```py
@dataclasses.dataclass
    attr: 'Enum' = betterproto.enum_field(1)

class Enum(betterproto.Enum):
    val = 1

>>> Test(attr=1)
Test(attr=<Enum.val: 1>)
# or
>>> Test().parse(b'\x08\x01')
Test(attr=<Enum.val: 1>)
```

### DIFF
--- a/betterproto/__init__.py
+++ b/betterproto/__init__.py
@@ -256,9 +256,17 @@ class Enum(int, enum.Enum):
 
     @classmethod
     def from_string(cls, name: str) -> int:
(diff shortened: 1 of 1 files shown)


boukeversteegh:
Hi Godot, thank you for your improvement.
This change is already included in Pr #102.

Gobot1234:
Ok then.

PR #116, opened 2020-07-11
Just some minor consistency changes

Gobot1234:


betterproto/__init__.py
```diff
@@ -661,7 +661,7 @@ def _get_field_default_gen(cls, field: dataclasses.Field) -> Any:
             elif t.__origin__ in (list, List):
                 # This is some kind of list (repeated) field.
                 return list
-            elif t.__origin__ == Union and t.__args__[1] == type(None):
+            elif t.__origin__ == Union and t.__args__[1] is type(None):
```
nat-n, line 664: while we're going for consistency it might as well be:

```suggestion
            elif t.__origin__ is Union and t.__args__[1] is type(None):
```

Gobot1234: That makes sense

PR #117, opened 2020-07-11
Use a custom Enum class for improved speed

Gobot1234:
Significant speed improvements over the builtin Enum class.
- 3x faster at creating instances of the enums.
- Roughly 2x as fast at creating small (15 member) enums.
- About 3x times faster at creating larger (50 member) enums.
- About 7x times faster direct attribute access, then 9x faster for name/values.

Full tests here https://gist.github.com/Gobot1234/e3883e9db151ce3b33adf4d38c140936

Fixes a bug discussed on slack relating to `to_dict` and `IndexErrors` when using Enums.


boukeversteegh:
Hi Gobot, thank you for your efforts.
For the bug, there is already an open pull request: #102

betterproto/enums.py
```diff
@@ -0,0 +1,149 @@
...
+            return instance._actual_enum_cls_ is self
+        except AttributeError:
+            return False
+
+    @property
+    def __members__(cls) -> Mapping[str, "EnumMemeber"]:
+        return MappingProxyType(cls._enum_member_map_)
+
+
+class Casing(metaclass=EnumMeta):
```
boukeversteegh, line 125: I think it would be useful to separate betterproto enum instances from the enum base class. Could you perhaps move Enum and its required Meta class to enum.py?

betterproto/__init__.py
```diff
@@ -776,7 +757,7 @@ def FromString(cls: Type[T], data: bytes) -> T:
         return cls().parse(data)
 
     def to_dict(
-        self, casing: Casing = Casing.CAMEL, include_default_values: bool = False
+        self, casing: Optional[Casing] = None, include_default_values: bool = False
```
boukeversteegh, line 760: Since Enums are not mutable, do we need this change? Or was passing a default not possible with the new class?

I think it was helpful for developers to see the default casing from the parameter.

betterproto/__init__.py
```diff
@@ -741,10 +722,10 @@ def parse(self: T, data: bytes) -> T:
                 pos = 0
                 value = []
                 while pos < len(parsed.value):
-                    if meta.proto_type in ["float", "fixed32", "sfixed32"]:
```
boukeversteegh, line 744: Great, thank you!

Since you had another PR with these changes, we can merge that one first, if you wouldn't mind to resolve the merge conflicts there.

betterproto/enums.py
```diff
@@ -0,0 +1,149 @@
...
+    """Casing constants for serialization."""
+
+    CAMEL = camel_case
+    SNAKE = snake_case
+
+
+class Enum(int, metaclass=EnumMeta):
```
boukeversteegh, line 132: Would it be possible to extend the python Enum class (IntEnum)? Or would that cause problems?

The benefit is that any library that handles enums by checking `value instanceof Enum` will then automatically support betterproto's Enum too.

It will also ensure we don't break clients that do this check.

betterproto/enums.py
```diff
@@ -0,0 +1,149 @@
...
+    @classmethod
+    def from_string(cls, name: str) -> EnumMember:
+        """Return the value which corresponds to the string name."""
+        try:
+            return cls._enum_member_map_[name]
+        except KeyError as e:
+            raise ValueError(f"Unknown member {name!r} for enum {cls.__name__}") from e
+
+    @classmethod
+    def from_value(cls, value: int) -> EnumMember:
```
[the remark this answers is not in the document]
Gobot1234: It was just for consistency sake I can remove both of them if needs be.

betterproto/enums.py
Gobot1234: I'll have to run the timings again on that, but that seems like a sensible change

Gobot1234: Righto I've fixed this not too sure how I feel about it however
```py
def patched_instance_check(self, instance):
    if isinstance(instance, (EnumMeta, EnumMember)):
        return True

    return type.__instancecheck__(self, instance)


_EnumMeta.__instancecheck__ = patched_instance_check
```

betterproto/__init__.py
Gobot1234: Unfortunately it raises an AttributeError as the class hasn't been created yet.

Gobot1234:
Everything should now work

betterproto/__init__.py
boukeversteegh: Actually, I guess we don't need to use Betterproto Enums for Casing. Perhaps just keep things simple and use the python Enum instead?

betterproto/enums.py
boukeversteegh: Oh interesting, I didn't know that was even possible!

Was extending not an option? Like literally

 ```python
from enum import IntEnum as PythonIntEnum
class Enum(PythonIntEnum):
   pass
```

boukeversteegh: > Everything should now work

Thank you for your hard work! 🙌

Gobot1234: Extending the class would make the class slower

betterproto/__init__.py
Gobot1234: I don't think it's that big of a deal, if someone wants to see the default they can just view source.

Gobot1234:
This is a bad idea, making these consistent is too much of a pain, sticking with the current implementation is fine.

Issue #119, opened 2020-07-14
Add support for custom options


adriangb:
Custom options are a vanilla proto feature documented [here](https://developers.google.com/protocol-buffers/docs/proto#customoptions). I personally have a use case for them to apply constraints, similar to [protoc-gen-validate](https://github.com/envoyproxy/protoc-gen-validate). I think supporting generic custom options would be a great addition to python-betterproto.

In order for this to happen, I see a couple changes that would be needed:

1. The options proto needs to be compiled by protoc and imported into the namespace from which `CodeGeneratorRequest. ParseFromString` is called. This req is documented in protocolbuffers/protobuf#3321.
2. `plugin.py` needs the ability to parse the custom options and somehow store them.
3. We'd need to figure out how to _use_ the custom options: Python has no concept of applying "options" (aside from comments). There are many mechanisms that could be employed (getters/setters, pydantic, comments, etc.). My feeling is that to cover all use cases, `plugin.py` would need to be made extendable so that individual implementations can decide how to parse custom options and how to use them in the generation of the output `.py` files.

Is there any interest in this? I'd be willing to try and work on something.

adriangb:
Separate comment to keep the first post clean.

Towards point (3) I started working on a refactor of `plugin.py` to make it object oriented since I think that would make it easier to extend/modify. Something like:

```python3
@dataclass
class Message:
    """Representation of a protobuf message.
    """
    parent: Union[OutputFile, Message]
    proto_obj: DescriptorProto
    path: List[int]
    fields: List[Union[Field, Message]] = field(default_factory=list)

    @property  # really this should be functools.cached_property but that's >=3.8 only
    def proto_name(self) -> str:
        return self.proto_obj.name

    @property
    def py_name(self) -> str:
        return pythonize_class_name(self.proto_name)

    def process_options(self, ??) -> ??:
        pass
```

Where the `parent` attribute allows recursively retrieving an `OutputFile` object from the field level to add imports, etc and the `process_options` method could be overridden to implement handling of custom options.

boukeversteegh:
Hi Adrian!

### Options and extensibility

Yes! Its a great idea, and it would be great to support extending betterproto with options.

1. I haven't had time to get into and properly understand the option parsing process, but if betterproto needs to do a bit of extra work to make it happen, I think thats fine.

[...]

PR #121, opened 2020-07-15
REF: Refactor plugin.py to use modular dataclasses in tree-like structure to represent parsed data

adriangb:
This is an attempt to modularize plugin.py so that it can easily be extended to make new plugins. For reference, discussion in #119.


boukeversteegh:
just a quick reply while i'm packing for a holiday.. please have a look also at #49 and my comment there.
i think it may be helpful if we first split up plugin.py into separate files, and merge that, before we end up with even more objects/variables in plugin.py.
on another point, small steps may be wise also, because if the refactor takes a long time, it will start to block new incoming work

adriangb:
Yeah splitting things up I think would be great.

Perhaps we can do this in the following steps:
1) Define the data structures and constants in a seperate file (`plugin_dataclasses.py`?)
2) Insert the data structures into the current `generate_code` method (i.e. still a single monolithic function, but replace the nested dicts with these new classes).
3) Refactor `generate_code` into a modular class.

Regarding these dataclasses specifically: how do you feel about their structure, and about moving the logic for each protobuf object into it's corresponding data class? Thus generating the response mainly consists of instantiating these days classes passing them (1) their parent, (2) their path and (3) their corresponding protobuf object.

boukeversteegh:
> Yeah splitting things up I think would be great.
> 
> Perhaps we can do this in the following steps:
> 
> 1. Define the data structures and constants in a seperate file (`plugin_dataclasses.py`?)
> 2. Insert the data structures into the current `generate_code` method (i.e. still a single monolithic function, but replace the nested dicts with these new classes).
> 3. Refactor `generate_code` into a modular class.
> 
> Regarding these dataclasses specifically: how do you feel about their structure, and about moving the logic for each protobuf object into it's corresponding data class? Thus generating the response mainly consists of instantiating these days classes passing them (1) their parent, (2) their path and (3) their corresponding protobuf object.

I like the idea of gradually moving towards OOP. It still feels somewhat more risky coming up with a top-down design and fitting the code into it, as opposed to isolating a responsibility from the code, and moving it into a (small) class, and see immediately that it made the code more simple.

I see a number of responsibilities that might get mixed up (they are already mixed up in the current codebase):

- analysis (reading all the source files to gather information that we need to start do mapping)
- mapping (deciding on the proper output structure, based on input structure, for example the files, packages and classes)
- rendering (transforming the output structure into actual python code)

It's not clear if we can split up the code immediately into clean components that separate these concerns, but even if we don't, we should know to some level which responsibilities each class takes.

But even if we mix it up, it would still be a step forward, and I support your proposal to have each protobuf type handled by separate components.

I haven't spent time to think about the paths/parents you mentioned, but i suppose it can be handy.

In any case, I would stick to _always working code_ as much as possible, instead of a big bang change, so that we will be able to keep releasing. It could be a good start to just extract a single object like Service, Message, Enum, or even ServiceMethod. Whichever is easiest. It will already be challenging.

betterproto/plugin.py
```diff
@@ -119,265 +130,547 @@ def get_comment(proto_file, path: List[int], indent: int = 4) -> str:
...
+    comment_indent: int = 4
+
+
+    def __post_init__(self):
+        """Checks that no fake default fields were left as placeholders."""
+        for field_name, field_val in self.__dataclass_fields__.items():
+            if field_val is PLACEHOLDER:
+                raise ValueError(f"`{field_name}` is a required field with no default value.")
+
+    @property   # TODO: replace with functools.cached_property
```
Gobot1234, line 146: This would bump the required version to generate protobufs, might be better to just copy the source for it somewhere

adriangb: Yeah. I see three alternatives:

1. Copy it from source (as you suggested).
2. Leave it as `@property` which means it will be recalculated every time.
3. Use `@functools.lru_cache(maxsize=1)` which would achieve the same thing, albeit with overkill complexity in the backend.

adriangb:
I pushed a version that has a separate `plugin_dataclasses.py` and is built on the latest version in master.

The way it is currently structured, there is no way to go one dataclass at a time, since they all rely on the `parent` and child tree-like architecture.

Makefile
```diff
@@ -8,8 +8,8 @@ help:               ## - Show this help.
 generate:           ## - Generate test cases (do this once before running test)
 	poetry run python -m tests.generate
 
-test:               ## - Run tests
-	poetry run pytest --cov betterproto
+test:               ## - Run tests, ingoring collection errors (ex from missing imports)
+	poetry run pytest --cov betterproto --continue-on-collection-errors
```
adriangb, line 12: This is obviously not needed for this PR, but I found it very useful, will remove before final review.

adriangb:
Moving from draft -> real PR.

Still have some failing tests, but most are passing.

betterproto/plugin.py
adriangb: I tested pulling the source, it's pretty straightforward. Only question is if it's worth the trouble or if the performance is okay as is.

src/betterproto/plugin_dataclasses.py
```diff
@@ -0,0 +1,699 @@
...
+            Names of the input files used to build this output.
+        """
+        return [f.name for f in self.input_files]
+
+
+@dataclass
+class Message(ProtoContentBase):
+    """Representation of a protobuf message.
+    """
+    parent: Union[Message, OutputTemplate] = PLACEHOLDER
```
adriangb, line 234: One of the main features of this refactor is the tree structure.

adriangb:
Tests are working on 3.8 and 3.7. 3.6 is broken because I was using [PEP 563](https://www.python.org/dev/peps/pep-0563/). It's easy enough to just use string literal type hints, I'll change that when needed.

adriangb:
Makes sense to me! I was also thinking that what is currently the `generate_code` function could also be split up, similar to [here](https://github.com/googleapis/protoc-docs-plugin/blob/2bdf14e394bbaa44b81286b1a19c5f229b51c667/protoc_docs/parser.py#L24), but that'd be another PR.

adriangb:
@nat-n please take a look at the structure I just created and let me know if that falls within the framework you were thinking of.

src/betterproto/plugin/plugin.bat
```diff
@@ -0,0 +1,2 @@
+@SET plugin_dir=%~dp0
+@python -m %plugin_dir% %*
```
adriangb, line 2: Not sure how this will pan out, not a .bat person...

src/betterproto/plugin.bat
```diff
@@ -1,2 +0,0 @@
-@SET plugin_dir=%~dp0
-@python %plugin_dir%/plugin.py %*
```
adriangb, line 2: moved to `src/betterproto/plugin/plugin.bat`

src/betterproto/plugin.py
```diff
@@ -1,480 +0,0 @@
-#!/usr/bin/env python
```
adriangb, line 1: split up into `src/betterproto/plugin/__init__.py` and `src/betterproto/plugin/parser.py`

src/betterproto/plugin/plugin.bat
boukeversteegh: I will test this out when I have time in the next couple of days :+1:

adriangb:
Happy to change any wording in the docs or name of variables! I was trying
to use general data structure terms so that someone unfamiliar with
protobuf might be able to understand the structure faster.

The tree is built using the same logic that used to build the nested dicts.
The main difference is that when you "add" a field to a message, the field
object registers the message as it's parent, and adds a reference to itself
into the message's `fields` list.

Unfortunately, I do not see a way of implementing these classes one at a
time, so I had to actually fully implement them into the parser (by parser
I mean the `generate_code` method). The good news is that _most_ of the
logic in these dataclasses was just copied over from `plugin.py` and the
logic in `generate_code` remains largely unchanged.

The specific goals of this PR would be:
- Create object oriented representations of protobuf object with references
to their parents/children.
- Refactor `generate_code` to use these classes.

I wasn't planning on restructuring the package, but I took that on based on
@nat-n 's suggestion. I agree with them that it makes things nicer and
cleaner, but it does increase the diff complexity considerably.

src/betterproto/plugin/parser.py
```diff
@@ -0,0 +1,180 @@
...
+        if item.options.map_entry:
+            # Skip generated map entry messages since we just use dicts
+            return
+        # Process Message
+        message_data = Message(parent=output_package, proto_obj=item, path=path)
+        for index, field in enumerate(item.field):
+            if is_map(field, item):
+                MapField(parent=message_data, proto_obj=field, path=path + [2, index])
+            elif is_oneof(field):
+                OneOfField(parent=message_data, proto_obj=field, path=path + [2, index])
```
boukeversteegh, line 165: Does this represent a field in a group, or the group itself?

If its the group, we could call it Group, or OneOfGroup. If that is in line with the descriptor name passed by the plugin.

src/betterproto/plugin/parser.py
```diff
@@ -0,0 +1,180 @@
...
+) -> None:
+    if isinstance(item, DescriptorProto):
+        if item.options.map_entry:
+            # Skip generated map entry messages since we just use dicts
+            return
+        # Process Message
+        message_data = Message(parent=output_package, proto_obj=item, path=path)
+        for index, field in enumerate(item.field):
+            if is_map(field, item):
+                MapField(parent=message_data, proto_obj=field, path=path + [2, index])
```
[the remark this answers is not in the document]
adriangb: Yeah either compiler or Parsed..Definition sound good to me.

src/betterproto/plugin/parser.py
adriangb: This would be a field that is part of a OneOf group. I don't think the oneof group itself is represented as an entity by protoc.

src/betterproto/plugin/parser.py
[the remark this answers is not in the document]
adriangb: With "Compiler" it would be:
- RequestCompiler
- MessageCompiler
- FieldCompiler
- MapEntryCompiler
- OneOfFieldCompiler
- EnumDefCompiler
- ServiceCompiler
- ServiceMethodCompiler

Right?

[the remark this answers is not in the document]
boukeversteegh: > With "Compiler" it would be:
> - RequestCompiler
> - MessageCompiler
> - FieldCompiler
> - MapEntryCompiler
> - OneOfFieldCompiler
> - EnumDefCompiler
> - ServiceCompiler
> - ServiceMethodCompiler
> 
> Right?

Yes :+1:

Although I haven't looked at protobuf descriptions to see if the classnames match, but if it doesn't, it would be an easy fix. So this looks fine to me.

EnumDefCompiler could just be EnumCompiler, I guess?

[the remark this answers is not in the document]
adriangb: The reason I suggested EnumDefCompiler instead of EnumCompiler is that _I_ originally was confused because I thought the enum objects protoc was supplying were enum fields or something like that instead of the actual enum defenitions. But maybe that was just me not knowing enough about protobuf structures... Again I really am open to anything you or others decide on 😃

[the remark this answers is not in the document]
adriangb: Acknowledged in https://github.com/danielgtaylor/python-betterproto/pull/121#discussion_r456908125.

src/betterproto/plugin/parser.py
boukeversteegh: I see, thanks for the reply. In the current library implementation, one of fields and normal fields are declared in pretty much the same way, except that one of has a group attribute set.

It seems a simple way to model the domain from the outside, as it avoids a nested structure (as does yours) but also another concept altogether.

Would you say keeping this approach in the OOP version could work, or did you have a specific motivation to split them into two classes?

adriangb: My reasoning for still splitting up into two classes is:

1. Be able to use an `isinstance` check instead of adding an `is_oneof` attribute or something like that.
2. Don't pollute the already complex `Field` namespace with a method that isn't relevant to a field unless it's a OneOf

src/betterproto/plugin/parser.py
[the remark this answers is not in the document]
boukeversteegh: Ah right, I understand. I may have had the same confusing, until I realized enums are also independent objects.

What further confuses things is that Enum values are also distinct from the Enum definition itself, although I think the Enum values end up being just fields on the Enum, but not sure (?).

How about EnumTypeCompiler?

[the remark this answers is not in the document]
adriangb: Yep "Enum Fields" are just Fields of the type <EnumDef name> if that makes sense. They're treated just like any other field. So the only time you encounter anything "Enum" related is the enum definitions. Why not `EnumDefinitionCompiler`? It's a bit long but I think it's most explicit. But otherwise `EnumTypeCompiler` sounds fine as well.

src/betterproto/plugin/models.py
```diff
@@ -0,0 +1,711 @@
...
+
+    @property
+    def proto_file(self) -> FieldDescriptorProto:
+        current = self
+        while not isinstance(current, OutputTemplate):
+            current = current.parent
+        return current.package_proto_obj
+
+    @property
+    def request(self) -> "Request":
```
boukeversteegh, line 174: Is this a convenience method or do we need it for compilation?
If possible, I'd prefer to avoid too much treewalking unless really needed, because it makes it harder to think about the code, especially when the tree walking is very abstract.

If we really need the plug-in request object from any object defined in it, we could also just pass it as a constructor arg, and keep things flat. Or would that make things more difficult?

src/betterproto/plugin/models.py
```diff
@@ -0,0 +1,711 @@
...
+        """Crawl the proto source code and retrieve comments
+        for this object.
+        """
+        return get_comment(
+            proto_file=self.proto_file, path=self.path, indent=self.comment_indent,
+        )
+
+
+@dataclass
+class Request:
```
boukeversteegh, line 191: Perhaps to distinguish it from Service method arguments (also called Method Input or Request), we could call it PluginRequest

src/betterproto/plugin/models.py
adriangb: At the moment, it's only being used by `ServiceMethod.py_input_message` to get all of the messages. I guess it could be a constructor arg, that might simplify things, or make them more complicated, not sure, I'd have to implement it.

src/betterproto/plugin/models.py
adriangb: Sure, sounds good to me.

adriangb:
@nat-n  I *think* the structure is now how you indicated and all of the entry points should work
@boukeversteegh I renamed the classes as discussed

src/betterproto/plugin/models.py
```diff
@@ -0,0 +1,713 @@
...
+        map_entry = f"{self.proto_obj.name.replace('_', '').lower()}entry"
+        for nested in self.parent.proto_obj.nested_type:
+            if nested.name.replace("_", "").lower() == map_entry:
+                if nested.options.map_entry:
+                    # Get Python types
+                    self.py_k_type = FieldCompiler(
+                        parent=self, proto_obj=nested.field[0],  # key
+                    ).py_type
+                    self.py_v_type = FieldCompiler(
+                        parent=self, proto_obj=nested.field[1],  # key
```
boukeversteegh, line 483: *value?

src/betterproto/plugin/models.py
```diff
@@ -0,0 +1,713 @@
...
+        # Get entries
+        self.entries = [
+            self.EnumEntry(
+                name=v.name,
+                value=v.number,
+                comment=get_comment(
+                    proto_file=self.proto_file, path=self.path + [2, i]
+                ),
+            )
+            for i, v in enumerate(self.proto_obj.value)
```
boukeversteegh, line 535: A more descriptive variable name could help to make clear what elements are iterated over. Are they Enum entries, EnumEntryDescriptors, or fielddescriptors?

src/betterproto/plugin/models.py
adriangb: Good catch!

src/betterproto/plugin/models.py
adriangb: Agreed. I tried to clarify, let me know if this helps!

adriangb:
I think you are giving me too much credit, I could not have made this PR without the work you and others have done before me!

I made the minor edits from your comments.

Since these changes are all internal, I think it's okay if we have to make changes again later, so I too am on board for merging what we've got so far and going from there.

src/betterproto/plugin/plugin.bat
abn: Is this used anymore?

src/betterproto/templates/template.py.j2
```diff
@@ -1,13 +1,13 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
-# sources: {{ ', '.join(description.files) }}
+# sources: {{ ', '.join(description.input_filenames) }}
 # plugin: python-betterproto
 from dataclasses import dataclass
 {% if description.datetime_imports %}
-from datetime import {% for i in description.datetime_imports %}{{ i }}{% if not loop.last %}, {% endif %}{% endfor %}
+from datetime import {% for i in description.datetime_imports|sort %}{{ i }}{% if not loop.last %}, {% endif %}{% endfor %}
```
abn, line 6: Would it make sense to implement this using isort similar to the way we use black today? Obviously as part of another changeset and not within here.

src/betterproto/templates/template.py.j2
```diff
@@ -1,13 +1,13 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
-# sources: {{ ', '.join(description.files) }}
+# sources: {{ ', '.join(description.input_filenames) }}
```
abn, line 2: Just for sanity sake it would be great to see the changes to names etc (and the template as a whole) to be split out into another PR just so that in case of any downstream issues we can revert easily. Just food for thought.

src/betterproto/plugin/__init__.py
```diff
@@ -0,0 +1 @@
+from .main import main
```
abn, line 1:
```suggestion
from betterproto.plugin.main import main
```

Or better yet, drop this in favour of consolidating `main.py` into `__main__.py` and update `pyproject.toml` to use `betterproto.plugin.__main__:main`.

src/betterproto/templates/template.py.j2
adriangb: I kind of agree. There's two main reasons why names were changed:
(1) When using the old names in the new structure really didn't make sense (ex: `properties` being a container for `fields`).
(2) When I just renamed things as I was developing for the sake of keeping my sanity and understanding everything.

I'd say the former makes sense for this PR but the latter (along with other re-naming for clarity's sake) would have made sense for another PR. The issue is that I don't remember which is which at this point 😣

src/betterproto/templates/template.py.j2
adriangb: That definitely makes sense for a future PR. Previously sort was being used somewhere else, I just moved it.

boukeversteegh:
It may not be needed anymore, as within the poetry shell we can call the
plugin with `--betterproto_out` or something, but i haven't checked the
code if we actually do that

boukeversteegh:
As proposed earlier, I will merge this PR since there don't seem to be showstoppers in it.
Any small improvements / nice to haves are welcome as new PR's.

Thank you!

PR #122, opened 2020-07-19
Fix the readme gRPC usage example

theunkn0wn1:
The sample gRPC client implementation in the readme had two defects in it which this PR corrects.

The first is a simple syntax error in the `async for` line, and should require no further explanation.

The second is a library usage error relating to the application of `grpclib.client.Channel`.
The issue stems from `grpclib.client.Channel` needing to be manually closed at the end of the session, which the readme example failed to do. This leads to runtime warnings when the channel object gets GC'ed.
As discussed in slack, not calling `channel.close()` is a implementation error and is the primary motivation of this PR.

Changes in tone:
I rewrote the client example to run from a script file instead of assuming a REPL, which I believe to be more realistic a use case. 

The other examples in the README are, imho, sufficient to demonstrate the intended functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,8 @@ async def main():
 
     # don't forget to close the channel when done!
     channel.close()
-
-
-asyncio.run(main())
+loop = asyncio.get_event_loop()
+loop.run_until_complete(main())
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -37,7 +37,6 @@ This project exists because I am unhappy with the state of the official Google p
   - Uses `SerializeToString()` rather than the built-in `__bytes__()`
   - Special wrapped types don't use Python's `None`
   - Timestamp/duration types don't use Python's built-in `datetime` module
-
 This project is a reimplementation from the ground up focused on idiomatic modern Python to help fix some of the above. While it may not be a 1:1 drop-in replacement due to changed method names and call patterns, the wire format is identical.
 
 ## Installation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,22 +152,33 @@ service Echo {
...
+    async for response in service.echo_stream(value="hello", extra_times=1):
         print(response)
 
-EchoStreamResponse(value="hello")
-EchoStreamResponse(value="hello")
+    # don't forget to close the channel when done!
+    channel.close()
+loop = asyncio.get_event_loop()
+loop.run_until_complete(main())
+
```
abn, line 174:
```suggestion


if __name__ == '__main__':
    asyncio.run(main())

```

README.md
```diff
@@ -153,22 +152,33 @@ service Echo {
...
 
-You can use it like so (enable async in the interactive shell first):
-
+A client can be implemented as follows:
 ```python
->>> import echo
->>> from grpclib.client import Channel
+import echo
+from grpclib.client import Channel
+import asyncio
```
abn, line 159:
```suggestion
import asyncio
import echo

from grpclib.client import Channel
```

README.md
```diff
@@ -153,22 +152,33 @@ service Echo {
...
->>> channel = Channel(host="127.0.0.1", port=1234)
->>> service = echo.EchoStub(channel)
->>> await service.echo(value="hello", extra_times=1)
-EchoResponse(values=["hello", "hello"])
 
->>> async for response in service.echo_stream(value="hello", extra_times=1)
+async def main():
+    channel = Channel(host="127.0.0.1", port=50051)
+    service = echo.EchoStub(channel)
+    print(await service.echo(value="hello", extra_times=1))
```
abn, line 165:
```suggestion
    response = await service.echo(value="hello", extra_times=1)
    print(response)
```

README.md
theunkn0wn1: i specifically didn't use `asyncio.run` (cause of second commit) as It is introduced into the standard library as of [python 3.7](https://docs.python.org/3/library/asyncio-task.html#asyncio.run), since this library targets >= 3.6 this method cannot be used without breaking in that older interpreter. 

Moving the runtime into a entry point check, however, is a good idea.

PR #124, opened 2020-07-21
Fix static type checking for grpclib client

abn:


abn:
@boukeversteegh can you please re-trigger the 3.6 build please? Not sure why it is failing on the CI, local tests pass.

boukeversteegh:
@abn I've rerun the tests, and this time there are some more detailed errors being shown.

Perhaps we need to extend _both_ ABC and IProtoMessage?

abn:
I suspect it is an issue with the dataclasses backport implementation for 3.6. I am curious why local tests ran okay for 3.6. Will look at it soon.

abn:
@boukeversteegh seems that by fixing the inheritence we corrected a code path that triggered a python 3.6 compatibility issue. As mentioned above,  looks like the issue has to do with subtle differences between dataclasses and dataclasses backport.

nat-n:
> I suspect it is an issue with the dataclasses backport implementation for 3.6. I am curious why local tests ran okay for 3.6. Will look at it soon.

That would be surprising. My understanding is the the dataclasses version is exactly what was then copied to the stdlib in 3.7?

src/betterproto/__init__.py
```diff
@@ -664,7 +666,8 @@ def _cls_for(cls, field: dataclasses.Field, index: int = 0) -> Type:
         """Get the message class for a field from the type hints."""
         field_cls = cls._type_hint(field.name)
         if hasattr(field_cls, "__args__") and index >= 0:
-            field_cls = field_cls.__args__[index]
+            if field_cls.__args__ is not None:
```
nat-n, line 669: Is there not something better we can do here? Won't this mean we just miss the annotation on 3.6?

abn: Don't think so, the difference is that `__args__` does not exist in 3.7+ and in the same scenario it returns `None` in 3.6. Not sure who behavior has changed here, we are just being stricter with the condition to avoid a `ValueError`.

abn:
@nat-n @boukeversteegh I have updated the fix to appraoch it a bit differently; rather than changing the inheritence I have now changed the type hinting so that it explicitly suggests `betterproto.Message` child class or something that implements the protocol `grpclib._typing.IProtoMessage`. For now, this fixes the issue.

nat-n:
Thanks for the changes, this looks good now.

PR #125, opened 2020-07-21
Add Documentation

Gobot1234:
Adds documentation for the library, any feedback is welcome and a live version of the current testing docs can be found at https://test-betterproto.rtfd.io/en/docs.


docs/conf.py
```diff
@@ -0,0 +1,113 @@
...
+
+# -- Options for HTML output -------------------------------------------------
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = "friendly"
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = "sphinx_iowmd_theme"
```
nat-n, line 63: Personally I rarely read docs cover to cover so I'm not a fan of having to click the burger menu to see the contents. Also I usually don't read docs on a small touch screen so a UI that optimises for that use case seems suboptimal.

Did you consider other themes?

nat-n: Also is there a way to get two levels of content in the side menu?

pyproject.toml
```diff
@@ -19,6 +19,9 @@ dataclasses = { version = "^0.7", python = ">=3.6, <3.7" }
 grpclib = "^0.3.1"
 jinja2 = { version = "^2.11.2", optional = true }
 protobuf = { version = "^3.12.2", optional = true }
+toml = { version = "0.10.1", optional = true }
```
nat-n, line 22: curious why the pinned version.

src/betterproto/__init__.py
```diff
@@ -634,7 +655,19 @@ def __bytes__(self) -> bytes:
         return output + self._unknown_fields
 
     # For compatibility with other libraries
-    SerializeToString = __bytes__
+    def SerializeToString(self: T) -> bytes:
```
nat-n, line 658: Is it necessary to creating a wrapping function to document an method alias?

What if you just stick the docstring above attribute declaration?

src/betterproto/__init__.py
```diff
@@ -780,20 +822,46 @@ def parse(self: T, data: bytes) -> T:
...
         """
-        Returns a dict representation of this message instance which can be
-        used to serialize to e.g. JSON. Defaults to camel casing for
-        compatibility but can be set to other modes.
-
-        `include_default_values` can be set to `True` to include default
-        values of fields. E.g. an `int32` type field with `0` value will
-        not be in returned dict if `include_default_values` is set to
-        `False`.
+        Returns a dict representation of this message instance which can be used to
```
nat-n, line 849:
```suggestion
        Returns a JSON serializable dict representation of this object
```

note: I'm note sure referring to "this message instance" is the most technically correct, since the object will actually be an instance of a subclass of Message.

src/betterproto/__init__.py
```diff
@@ -918,13 +995,41 @@ def from_dict(self: T, value: dict) -> T:
...
-        """
-        Parse the key/value pairs in `value` into this message instance. This
-        returns the instance itself and is therefore assignable and chainable.
+        """A helper function to return the message instance from its JSON
+        representation. This returns the instance itself and is therefore assignable
+        and chainable.
+
+        This is equivalent to::
+
+            >>> return message.from_dict(json.loads(value))
```
nat-n, line 1023: I don't think `>>>` is appropriate here since it suggests to me that we're talking about a REPL context.

docs/conf.py
boukeversteegh: Same thought~

I see that some `readthedocs` themes have a more extensive sidebar, so you can easily jump to different pages.

For example: https://docs.scrapy.org/en/latest/

To be honest, I find a lot of issues with the Material theme

- code blocks have no border or distinct color, so the page looks disorganized, and code is not readable
- the size of headings in the TOC blocks is inconsistent (higher level headings are smaller)
- no overview of pages / sidebar

I suggest we use the theme / structure of Scrapy

pyproject.toml
```diff
@@ -19,6 +19,9 @@ dataclasses = { version = "^0.7", python = ">=3.6, <3.7" }
 grpclib = "^0.3.1"
```
nat-n, line 19: After rebasing I'd suggest adding a "docs" poe task to build sphinx locally

boukeversteegh:
Thank you for investigating a more mature documentation workflow @Gobot1234 !

It seems sensible to switch over to RST and use Sphinx, as it apparently is quite standard within the Python ecosystem, although I'm not very familiar with it myself.

I have some remarks that I think need to be addressed before we can switch over to these docs as the 'official' docs.

- Docs generation from code:
  - Most of the methods and classes are not intended for direct use, and I would even actively discourage using the internal betterproto classes, except by proxy through the generated Message classes. Adding docs for the internals gives the impression that these methods/classes are part of the public API. I'm concerned for example if users will manually extend the `Message` baseclass, and that compatibility will break when we decide to refactor our baseclass and generated classes. We should therefor carefully select a very few methods that users may need documentation for (such as `Parse` and `which_one_of`), and omit everything else.
  - By generating docs for all classes and objects, we create a lot of noise in the docs. Suppose you want to know how to use protobuf Enums with Betterproto, and you search `Enum`, and ended up on the generated docs for the Enum baseclass, instead of the Getting Started page (enum section). That is going to cause frustrations with the docs and the library. This is what has made javadocs almost entirely useless, unless the tutorials are actually part of the class-doc, which i wouldn't recommend, because we want to structure our docs to match the needs of the user, and not the internal code structure.
- Theme
  - as mentioned in the comment
- How to update our documentation
  - We will need some instructions for how to update the docs. Before it was straightforward, just commit the markdown files. Now I'm not sure what I need to do, and if I need another account or separate permissions to fix documentation.

docs/conf.py
nat-n: I would second using the basic RTD theme for clearer presentation: https://sphinx-rtd-theme.readthedocs.io/en/stable/

pyproject.toml
nat-n: Also, it's easily to make syntax errors with sphinx, and since sphinx builds often aren't exactly instant, I'd suggesting adding both a poe task and a CI step that runs rstcheck (which will need to be added as a dev dep)

pyproject.toml
Gobot1234: No particular reason it was just the most recent

src/betterproto/__init__.py
Gobot1234: I think it improves readability and the same thing could be said about `FromString` could it not? As it could be done with `FromString = classmethod(parse)`.

pyproject.toml
nat-n: The version itself is fine, though I'm curious why it's pinned without a `^`?

docs/index.rst
```diff
@@ -0,0 +1,40 @@
+Welcome to betterproto's documentation!
+=======================================
+
+betterproto is a protobuf compiler and interpreter. It was written for the modern age making use of
+Python 3.6+ features to generate readable, understandable and idiomatic Python code.
+
+
+Features:
+~~~~~~~~~
```
nat-n, line 9: The logical flow of the features section is a bit unclear to me. What I think is important is that someone who just landed here from google who doesn't know what betterproto is about, gets a clear idea of the scope of the library and why it's a good choice.

How about:

```
Idiomatic code generation for Protobuf 3 & gRPC
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

- Python source code is generated as python packages with the same structure as the source protobuf packages
- Message types are modelled as python dataclasses with support for binary as JSON serialisation
- Message contents are represented using the relevant python type such as enum, datetime, or timedelta
- Generated code includes complete type annotations
- Generated gRPC clients support async/await
```

docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
nat-n, line 1: It doesn't seem optimal to have a separate makefile for the docs... That and the main project doesn't use make anymore.

docs/conf.py
```diff
@@ -0,0 +1,68 @@
...
+import toml
+
+
+# -- Project information -----------------------------------------------------
+
+project = "betterproto"
+copyright = "2019 Daniel G. Taylor"
+author = "danielgtaylor"
+pyproject = toml.load(open(pathlib.Path(__file__).parent.parent / "pyproject.toml"))
+
```
nat-n, line 22: Is license a supported field here?

src/betterproto/__init__.py
nat-n: Readability for the docs or the code? FromString is different because it's wrapping an instance method in a class method.

docs/Makefile
Gobot1234: It's the default generated by sphinx, not too sure if its needed however

src/betterproto/__init__.py
Gobot1234: I think it's just better to be more consistent, keep the code more readable and make it slightly clearer what is happening.

pyproject.toml
Gobot1234: So what I was thinking was to setup a test that just runs https://pypi.org/project/restructuredtext-lint and checks if that finds any errors

docs/conf.py
Gobot1234: I don't think so.

docs/index.rst
Gobot1234: Does this seem better?
```
- Generated messages are both binary & JSON serializable
- Messages use relevant python types, e.g. ``Enum``, ``datetime`` and ``timedelta`` objects
- ``async``/``await`` support for gRPC Clients
- Relative imports, mocking protobuf package structure
- Mypy type checkable with full type annotations
- Modern idiomatic python naming conventions
```

nat-n: It's better. As a feature list without much introduction I feel it needs to say things in a way that doesn't assume too much context.

I think it's worth mentioning dataclasses.
I'm not totally sure what you mean about "naming conventions".
I don't feel "Relative imports" per se is that interesting a feature so much as that the output is a valid python package with a logical structure (unlike the standard compiler). I wouldn't use the word mocking in this context.

docs/Makefile
nat-n: Let's ditch it.

Gobot1234:
@boukeversteegh 
- I've added a note to the docs that None of the classes are meant to be manually extended by the user to hopefully clear up any confusion.
- For the rST linting that can be configured once this has passed, see (https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html) along with an extra set of steps to the Code Quality workflow which lints the docs for errors.
- For commiting to docs process should be the same as before just using rST over markdown.
- For noise issue, I've tried to keep documentation as sparing as possible to try to mitigate this.

pyproject.toml
```diff
@@ -39,6 +41,7 @@ protoc-gen-python_betterproto = "betterproto.plugin:main"
 
 [tool.poetry.extras]
 compiler = ["black", "jinja2", "protobuf"]
+docs = ["toml", "sphinx"]
```
nat-n, line 44: what's the idea of adding sphinx as a extras dep here? does the rtd build process expect it?

nat-n: Should also add the following to  dev-dependencies to support building the docs in dev.

```toml
sphinx = "3.1.2"
sphinx-rtd-theme = "0.5.0"
```

Then adding the following to the `tool.poe.tasks` section will allow developers to build docs locally via `poe docs`

```toml
docs        = { cmd = "sphinx-build docs docs/build", help = "Build the sphinx docs"}
```

docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
nat-n, line 1: Is this file needed?

Gobot1234: I think it's how most people build docs on there machines, I'll do a quick check though.

pyproject.toml
Gobot1234: It will just make the building a bit faster as it doesn't need compiler dependencies.

docs/quick-start.rst
```diff
@@ -1,6 +1,27 @@
...
+
+    python3 -m pip install -U betterproto
+
+If you are using Windows, then the following should be used instead:
+
+.. code-block:: sh
+
+    py -3 -m pip install -U betterproto
+
+To get compiling support, you should use betterproto[compiler] instead of betterproto,
```
nat-n, line 19:
```suggestion
To install the protoc plugin, you should use betterproto[compiler] instead of betterproto,
```

pyproject.toml
```diff
@@ -19,6 +19,8 @@ dataclasses = { version = "^0.7", python = ">=3.6, <3.7" }
 grpclib = "^0.3.1"
 jinja2 = { version = "^2.11.2", optional = true }
 protobuf = { version = "^3.12.2", optional = true }
+toml = { version = "^0.10.1", optional = true }
+sphinx = { version = "3.1.2", optional = true }
```
nat-n, line 23: I don't think these belong here. Documentation isn't a concern of the distributed library.

nat-n (changes requested):

pyproject.toml
Gobot1234: I thought all dependencies had to go there whoops.

nat-n (approved):

PR #126, opened 2020-07-21
Support deprecated message and fields

abn:
Initial attempt at warning on deprecated field use.

Generated sample:
```py
# sources: deprecated.proto
# plugin: python-betterproto
import warnings

import betterproto


@dataclass
    """Some documentation about the Test message."""

    v: int = betterproto.int32_field(1)
    value: int = betterproto.int32_field(2)

        warnings.warn("Test is deprecated", DeprecationWarning)
        super().__post_init__()
        if self.v:
            warnings.warn("Test.v is deprecated", DeprecationWarning)
```


boukeversteegh:
Great work! Unfortunately the OOP refactor has made this PR in conflict.
Would you be willing to port it to the new structure?

abn:
@boukeversteegh this should be ready again.

tests/inputs/deprecated/deprecated.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+// Some documentation about the Test message.
+message Test {
+    // Some documentation about the value.
+    option deprecated = true;
+    int32 v = 1 [deprecated=true];
```
boukeversteegh, line 7: Is `int32 v = 1 [deprecated=true]` redundant with the `deprecated_field` test?

abn: It was done specificaly to verify that we handle the case where the message and the field deprecation explicitly.

adriangb:
Happy to see that this worked well with that refactor 😄

PR #129, opened 2020-07-26
Properly serialize zero-value messages in a oneof group

chris-chambers:
Zero-value messages are now emitted when they are the selected field in a `oneof` group.

There is some existing code that checks for this condition here:

https://github.com/chris-chambers/python-betterproto/blob/66bf37d966793c01055772a65edee313d7ffe637/src/betterproto/__init__.py#L598-L600

But the decision needed to be carried through to the `serialize_empty` parameter of `_serialize_single` to support zero-value messages.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -640,7 +640,7 @@ class Message(ABC):
                     meta.number,
                     meta.proto_type,
                     value,
-                    serialize_empty=serialize_empty,
+                    serialize_empty=serialize_empty or selected_in_group,
                     wraps=meta.wraps or "",
                 )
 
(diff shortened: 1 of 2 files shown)


nat-n:
Thanks for the fix and test! And sorry for the slow response.

This should make betterproto's serialization more consistent with grpcio's right? It would be useful for ease of maintenance to have a simple test specifically to that effect, i.e. a test example in the `tests/inputs` dir that from which both betterproto and grpcio models get generated, and then compare then serialisation results of each. 

Does this make sense? would you be willing to add this?

chris-chambers:
Sorry, while these PRs have been in review, I discovered that the encoding and decoding latency of betterproto is too high for my use-case, so I need to put my efforts into another solution.  Please, feel free to add whatever additional tests are needed, if you want.  I've set this PR to be directly editable by contributors to make that straightforward.

nat-n:
Closing this in favour of #176 since I wasn't able to update the original branch for some reason.

PR #130, opened 2020-07-26
Support for recursive messages

### LINKED ISSUE #13: Indirect recursive message - RecursionError: maximum recursion depth hit on encoding initial default values
Hi there,
I really prefer the betterproto probuf output to Google's, so well done on that!  I've come across the following issue, and narrowed it down to the simplest proto I can to reproduce.
 
Initialising the following proto, (after compiling with python-betterproto)
```
syntax = "proto3";
package schemas;

message Action {
    oneof action_oneof {
        ref not_repeated = 2;
    }
}

message ref {
    string name = 2;
    Action action = 3;
}
```
with a simple `test = schemas.Action()`

is causing recursion and a failure to create the test object.  It looks as far as I can tell that its attempting to work out the default values for the initial proto.

```
File "<string>", line 4, in __init__
  File "/lib/python3.7/site-packages/betterproto/__init__.py", line 460, in __post_init__
    setattr(self, field.name, self._get_field_default(field, meta))
  File "/lib/python3.7/site-packages/betterproto/__init__.py", line 593, in _get_field_default
    value = t()
  File "<string>", line 3, in __init__
  File "/lib/python3.7/site-packages/betterproto/__init__.py", line 460, in __post_init__
    setattr(self, field.name, self._get_field_default(field, meta))

repeated then
RecursionError: maximum recursion depth exceeded while calling a Python object
```

When the proto definition is changed to 
```
message ref {
    string name = 2;
    repeated Action action = 3;
}
```
then the recursion error does not occur.  

The standard 'protoc --python_out='  does not exhibit this behaviour, hence I thought I'd raise this.
cheers
Nick

### LINKED ISSUE #74: Recursive message support
I am working on implementing a GRPC client which has a message such as this:

```
message Failure {
    ...
    Failure cause = 4;
    ...
}
```

Unfortunately it seems like betterproto attempts to build an object graph for this and this results in the following error:

```
RecursionError: maximum recursion depth exceeded while calling a Python object
```

I was wondering whether the project is planning on supporting this. 

My plan in the short run is to just fork betterproto for my project and not set a default value for sub-messages in `_get_field_default`. 

If the maintainers have a better approach to fixing this I would gladly offer my time in submitting a PR for it.

chris-chambers:
Changes message initialization (`__post_init__`) so that default values are no longer eagerly created to prevent infinite recursion when initializing recursive messages.

As a result, `PLACEHOLDER` will be present in the message for any uninitialized fields.  So, an implementation of `__get_attribute__` is added that checks for `PLACEHOLDER` and lazily creates and stores default field values.

And, because `PLACEHOLDER` values don't compare equal with zero values, a custom implementation of `__eq__` is provided, and the code generation template is updated so that messages generate with `@dataclass(eq=False, repr=False)`.

Fixes #13 and fixes #74.

---

This needs careful review!  I'm not sure if this is the desired approach.  It's just the least-friction way I could think of.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -530,7 +530,7 @@ class Message(ABC):
         return super().__getattribute__(name)
 
     def __eq__(self, other):
-        if type(self) != type(other):
+        if type(self) is not type(other):
             return False
 
         equal = True
@@ -558,7 +558,7 @@ class Message(ABC):
             if value is PLACEHOLDER:
                 continue
             found = True
-            parts.extend([field_name, "=", repr(value), ","])
+            parts.extend([field_name, "=", repr(value), ", "])
 
         if found:
             parts.pop()
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -550,20 +550,15 @@ class Message(ABC):
 
         return equal
 
-    def __repr__(self):
-        parts = [self.__class__.__name__, "("]
-        found = False
+    def __repr__(self) -> str:
+        parts = []
         for field_name in self._betterproto.meta_by_field_name:
             value = self.__raw_get(field_name)
             if value is PLACEHOLDER:
                 continue
-            found = True
-            parts.extend([field_name, "=", repr(value), ", "])
+            parts.append(f"{field_name}={value!r}")
 
-        if found:
-            parts.pop()
-        parts.append(")")
-        return "".join(parts)
+        return f"{self.__class__.__name__}({', '.join(parts)})"
 
     def __getattribute__(self, name: str) -> Any:
         value = super().__getattribute__(name)
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -428,6 +428,7 @@ class ProtoClassMetadata:
         "cls_by_field",
         "field_name_by_number",
         "meta_by_field_name",
+        "sorted_field_names",
     )
 
     def __init__(self, cls: Type["Message"]):
@@ -453,6 +454,9 @@ class ProtoClassMetadata:
         self.oneof_field_by_group = by_group
         self.field_name_by_number = by_field_number
         self.meta_by_field_name = by_field_name
+        self.sorted_field_names = tuple(
+            by_field_number[number] for number in sorted(by_field_number.keys())
+        )
 
         self.default_gen = self._get_default_gen(cls, fields)
         self.cls_by_field = self._get_cls_by_field(cls, fields)
@@ -529,17 +533,16 @@ class Message(ABC):
     def __raw_get(self, name: str) -> Any:
         return super().__getattribute__(name)
 
-    def __eq__(self, other: T) -> bool:
+    def __eq__(self, other) -> bool:
         if type(self) is not type(other):
             return False
 
         for field_name in self._betterproto.meta_by_field_name:
             self_val = self.__raw_get(field_name)
             other_val = other.__raw_get(field_name)
-            if self_val is PLACEHOLDER and other_val is PLACEHOLDER:
-                continue
-
             if self_val is PLACEHOLDER:
+                if other_val is PLACEHOLDER:
+                    continue
                 self_val = self._get_field_default(field_name)
             elif other_val is PLACEHOLDER:
                 other_val = other._get_field_default(field_name)
@@ -550,16 +553,19 @@ class Message(ABC):
         return True
 
     def __repr__(self) -> str:
-        parts = []
-        for field_name in self._betterproto.meta_by_field_name:
-            value = self.__raw_get(field_name)
-            if value is PLACEHOLDER:
-                continue
-            parts.append(f"{field_name}={value!r}")
-
+        parts = [
+            f"{field_name}={value!r}"
+            for field_name in self._betterproto.sorted_field_names
+            for value in (self.__raw_get(field_name),)
+            if value is not PLACEHOLDER
+        ]
         return f"{self.__class__.__name__}({', '.join(parts)})"
 
     def __getattribute__(self, name: str) -> Any:
+        """
+        Lazily initialize default values to avoid infinite recursion for recursive
+        message types
+        """
         value = super().__getattribute__(name)
         if value is not PLACEHOLDER:
             return value
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -529,26 +529,25 @@ class Message(ABC):
     def __raw_get(self, name: str) -> Any:
         return super().__getattribute__(name)
 
-    def __eq__(self, other):
+    def __eq__(self, other: T) -> bool:
         if type(self) is not type(other):
             return False
 
-        equal = True
         for field_name in self._betterproto.meta_by_field_name:
             self_val = self.__raw_get(field_name)
             other_val = other.__raw_get(field_name)
             if self_val is PLACEHOLDER and other_val is PLACEHOLDER:
                 continue
-            elif self_val is PLACEHOLDER:
+
+            if self_val is PLACEHOLDER:
                 self_val = self._get_field_default(field_name)
             elif other_val is PLACEHOLDER:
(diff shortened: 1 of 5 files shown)


chris-chambers:
It looks like this may also fix #63.  But, since I wasn't trying to do that, it should get some extra review.

src/betterproto/__init__.py
```diff
@@ -519,23 +519,67 @@ def __post_init__(self) -> None:
...
         # Now that all the defaults are set, reset it!
         self.__dict__["_serialized_on_wire"] = not all_sentinel
         self.__dict__["_unknown_fields"] = b""
         self.__dict__["_group_current"] = group_current
 
+    def __raw_get(self, name: str) -> Any:
+        return super().__getattribute__(name)
+
+    def __eq__(self, other):
+        if type(self) != type(other):
```
Gobot1234, line 539: You shouldn't compare types using `__ne__`

src/betterproto/__init__.py
```diff
@@ -519,23 +519,67 @@ def __post_init__(self) -> None:
...
+
+    def __repr__(self):
+        parts = [self.__class__.__name__, "("]
+        found = False
+        for field_name in self._betterproto.meta_by_field_name:
+            value = self.__raw_get(field_name)
+            if value is PLACEHOLDER:
+                continue
+            found = True
+            parts.extend([field_name, "=", repr(value), ","])
```
Gobot1234, line 567: There should be a space after this comma, to match PEP8

Gobot1234: I think it would be better to join them using ", " to avoid this.

On another note, since this is for constructing the class from the repr you should probably also check if the field is an Enum and then use `Enum.__name__.value.name`

chris-chambers: I'm not sure I follow this:

> On another note, since this is for constructing the class from the repr you should probably also check if the field is an Enum and then use Enum.__name__.value.name

`repr` isn't generally meant to be round-trippable.  For example:

```python
from enum import Enum

class Color(Enum):
    RED = 1
    BLUE = 2

@dataclass
class Foo(object):
    color: Color

repr(Foo(Color.RED)) # => "Foo(color=<Color.RED: 1>)"
```

This implementation of `__repr__` gives a similar result to what the implementation provided by `@dataclass` gives.

Gobot1234: According to the python docs (https://docs.python.org/3/reference/datamodel.html#object.__repr__) where possible an object's `__repr__` should allow for `x = eval(repr(x))` where possible, with the current implementation it would not be possible to do this, so I was suggesting that Enum fields should be converted to there Enum class name and member name to allow for this.

chris-chambers: I agree with your interpretation of the docs, but already the built in `enum.Enum` does not follow this convention.  I'm just delegating to each field's existing `__repr__` implementation, which is as much as we should do, in my opinion.  I don't think `Message`'s implementation of `__repr__` should inspect its field's types and change it's behavior.

src/betterproto/templates/template.py.j2
```diff
@@ -67,7 +67,7 @@ class {{ service.py_name }}Stub(betterproto.ServiceStub):
                                             Optional[{{ field.annotation }}]
                                          {%- else -%}
                                             {{ field.annotation }}
-                                         {%- endif -%} =
```
chris-chambers, line 70: Just noticed this accidental whitespace change.  Repairing now.

src/betterproto/__init__.py
chris-chambers: I just realized that `betterproto` generates enums that inherit from `betterproto.Enum`, not `enum.Enum`.  I apologize for my confusion.

Given that, it would be possible to implement `betterproto.Enum.__repr__` and get the behavior you're talking about.  And it seems like a perfectly good thing to do.  It's independent of this PR, though, and would have the same benefit whether it lands before or after this, so I still wouldn't want to tie the two changes together.

boukeversteegh: also in favor of keeping general improvements out of scope of bugfix PR's, unless they are essential

Gobot1234: I wasn't suggesting altering the `betterproto.Enum.__repr__` but altering `Message.__repr__` to use `enum.Enum.__str__` instead to follow the standard.

chris-chambers:
Sorry, while these PRs have been in review, I discovered that the encoding and decoding latency of betterproto is too high for my use-case, so I need to put my efforts into another solution.

Please, feel free to make whatever changes you wish, if you want to incorporate these changes.  I enabled direct editing of the PR by contributors, so it should be easy to make changes.  Or, if you prefer to close this and take another approach to solving this problem, that works for me.

aleneum:
Hi @chris-chambers,

thank you for your effort nevertheless! I am in need of this feature. Is there something that could be done by me before this PR is ready to be merged (@Gobot1234)?

Gobot1234:
It LGTM.

Gobot1234:
I can also fix the conflicts at some point soon-ish

chris-chambers:
I've got time this afternoon, so I'll rebase and fix the merge conflict.  So no need, @Gobot1234

aleneum:
just fyi: I forked `master`, rebased and then merged the PR. There was one issue in the template where ` {% for message in description.messages %} ` wasnt correctly updated (to `output_file`) but this seems to be fixed already. Parsing a rather complex message with a redundant item composition no longer caused recursion errors for me.

src/betterproto/__init__.py
```diff
@@ -519,23 +519,67 @@ def __post_init__(self) -> None:
...
+            elif self_val is PLACEHOLDER:
+                self_val = self._get_field_default(field_name)
+            elif other_val is PLACEHOLDER:
+                other_val = other._get_field_default(field_name)
+
+            if self_val != other_val:
+                equal = False
+                break
+
+        return equal
```
Gobot1234, line 557:
```suggestion
    def __eq__(self, other: T) -> bool:
        if type(self) is not type(other):
            return False

        for field_name in self._betterproto.meta_by_field_name:
            self_val = self.__raw_get(field_name)
            other_val = other.__raw_get(field_name)
            if self_val is PLACEHOLDER and other_val is PLACEHOLDER:
                continue

            if self_val is PLACEHOLDER:
                self_val = self._get_field_default(field_name)
            elif other_val is PLACEHOLDER:
                other_val = other._get_field_default(field_name)

            if self_val != other_val:
                return False

        return True
```
is IMO cleaner and is a bit easier to follow

src/betterproto/__init__.py
```diff
@@ -519,23 +519,67 @@ def __post_init__(self) -> None:
...
+            value = self.__raw_get(field_name)
+            if value is PLACEHOLDER:
+                continue
+            found = True
+            parts.extend([field_name, "=", repr(value), ", "])
+
+        if found:
+            parts.pop()
+        parts.append(")")
+        return "".join(parts)
```
Gobot1234, line 572:
```suggestion
    def __repr__(self) -> str:
        parts = []
        for field_name in self._betterproto.meta_by_field_name:
            value = self.__raw_get(field_name)
            if value is PLACEHOLDER:
                continue
            parts.append(f"{field_name}={value!r}")

        return f"{self.__class__.__name__}({', '.join(parts)})"
```
Is a lot more readable

Gobot1234:
This https://github.com/danielgtaylor/python-betterproto/blob/558b04ca169a1bbb9e1184b7bb81f5e3199be416/src/betterproto/__init__.py#L485-L492 is also going to need to be updated

src/betterproto/__init__.py
nat-n: Yes this is an improvement.

nat-n:
Thanks to @chris-chambers et al. for working on this. Sorry for the slow turn around.

I pushed some final changes:
- Rebase onto master
- Add more test cases using the standard test pattern
- Tweak some of the new code
- Ensure repr keys are consistently ordered by field number

Ready to merge I'd say 👍

nat-n (approved):

PR #134, opened 2020-08-04
Add __version__ attribute to package

nat-n:
This allows users to check which version of betterproto they're using be evaluating `betterproto.__version__`.

Includes a test to prevent the value from being incorrect in a release.

### DIFF
--- a/src/betterproto/_version.py
+++ b/src/betterproto/_version.py
@@ -1,0 +1,3 @@
+from pkg_resources import get_distribution
+
+__version__ = get_distribution("betterproto").version
--- a/poetry.lock
+++ b/poetry.lock
@@ -62,14 +62,6 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 pytz = ">=2015.7"
 
 [[package]]
-name = "backports-datetime-fromisoformat"
-version = "1.0.0"
-description = "Backport of Python 3.7's datetime.fromisoformat"
-category = "main"
-optional = false
-python-versions = "*"
-
-[[package]]
 name = "black"
 version = "20.8b1"
 description = "The uncompromising code formatter."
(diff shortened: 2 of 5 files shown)


boukeversteegh (dismissed):

abn:
I this really required? Is this information placed in the generated dataclasses and do we expect something along the lines of min version required for compatibility between runtime vs compile time versions?  I just feel this adds more things to keep in sync if we do not have a clear use case.  If someone really wanted wouldn't something like this be sufficient? 

```py
__import__('pkg_resources').get_distribution("betterproto").version
```

nat-n:
> I this really required? Is this information placed in the generated dataclasses and do we expect something along the lines of min version required for compatibility between runtime vs compile time versions? I just feel this adds more things to keep in sync if we do not have a clear use case. If someone really wanted wouldn't something like this be sufficient?
> 
> ```python
> __import__('pkg_resources').get_distribution("betterproto").version
> ```

It's not really required. It's a convenience, and I think it's quite a common convention, to be able to do "import some_lib; some_lib.__version__".

It adds an extra step to the process of updating the version, but at least the test means you can't forget.

src/betterproto/_version.py
```diff
@@ -0,0 +1 @@
+__version__ = "2.0.0b1"
```
cetanu, line 1: You can get this programmatically, I think?

```
from pkg_resources import get_distribution
get_distribution('betterproto').version
```

cetanu (commented):
Personally I am 100% okay with the `__import__('pkg_resources').get_distribution("betterproto").version` implementation.

I don't think it adds any harm or unnecessary overheads/maintenance to the library.

cetanu (dismissed):
I made changes to derive the version from the current distribution.

No point in not merging this. It gives users a shortcut that is very obviously seen from perusing the source.

cetanu (approved):

cetanu (approved):

PR #135, opened 2020-08-04
MAIN: Small improvements to models.py

adriangb:
Two misc improvements I noticed were possible during other dev.

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -151,6 +151,7 @@ class ProtoContentBase:
 
     path: List[int]
     comment_indent: int = 4
+    parent: Union["Messsage", "OutputTemplate"]
 
     def __post_init__(self):
         """Checks that no fake default fields were left as placeholders."""
@@ -503,16 +504,13 @@ class MapEntryCompiler(FieldCompiler):
                     self.proto_v_type = self.proto_obj.Type.Name(nested.field[1].type)
         super().__post_init__()  # call FieldCompiler-> MessageCompiler __post_init__
 
-    def get_field_string(self, indent: int = 4) -> str:
-        """Construct string representation of this field."""
-        name = f"{self.py_name}"
-        annotations = f": {self.annotation}"
-        betterproto_field_type = (
-            f"betterproto.map_field("
-            f"{self.proto_obj.number}, betterproto.{self.proto_k_type}, "
-            f"betterproto.{self.proto_v_type})"
-        )
-        return name + annotations + " = " + betterproto_field_type
+    @property
+    def betterproto_field_args(self):
+        return f", betterproto.{self.proto_k_type}, betterproto.{self.proto_v_type}"
+
+    @property
+    def field_type(self) -> str:
+        return "map"
 
     @property
     def annotation(self):


src/betterproto/plugin/models.py
```diff
@@ -503,16 +504,13 @@ def __post_init__(self):
...
-        name = f"{self.py_name}"
-        annotations = f": {self.annotation}"
-        betterproto_field_type = (
-            f"betterproto.map_field("
-            f"{self.proto_obj.number}, betterproto.{self.proto_k_type}, "
-            f"betterproto.{self.proto_v_type})"
-        )
-        return name + annotations + " = " + betterproto_field_type
+    @property
+    def betterproto_field_args(self):
```
Gobot1234, line 508:
```suggestion
    def betterproto_field_args(self) -> str:
```

src/betterproto/plugin/models.py
```diff
@@ -503,16 +504,13 @@ def __post_init__(self):
                     self.proto_v_type = self.proto_obj.Type.Name(nested.field[1].type)
         super().__post_init__()  # call FieldCompiler-> MessageCompiler __post_init__
 
-    def get_field_string(self, indent: int = 4) -> str:
```
nat-n, line 506: Could you explain a bit what this change is about? I'm not seeing the full picture here.

adriangb: I realized that we could easily re-use the `get_field_string` machinery in the base class instead of overwriting it.

src/betterproto/plugin/models.py
```diff
@@ -503,16 +504,13 @@ def __post_init__(self):
...
-        annotations = f": {self.annotation}"
-        betterproto_field_type = (
-            f"betterproto.map_field("
-            f"{self.proto_obj.number}, betterproto.{self.proto_k_type}, "
-            f"betterproto.{self.proto_v_type})"
-        )
-        return name + annotations + " = " + betterproto_field_type
+    @property
+    def betterproto_field_args(self) -> str:
+        return f", betterproto.{self.proto_k_type}, betterproto.{self.proto_v_type}"
```
adriangb, line 509: As a side note (for another PR maybe) I think this should return a list that `get_field_string` uses in `",".join(..)`.

src/betterproto/plugin/models.py
```diff
@@ -503,16 +504,13 @@ def __post_init__(self):
...
+
+    @property
+    def field_type(self) -> str:
+        return "map"
```
adriangb, line 513: Might be worth checking what the value returned from the base class' method is before overwriting it.

adriangb: I checked, it returns `message`.

src/betterproto/plugin/models.py
adriangb: Implemented, worked great

nat-n:
black says no :(
https://github.com/danielgtaylor/python-betterproto/pull/135/checks?check_run_id=981360800

adriangb:
@nat-n  black is now saying _yes_

nat-n (approved):

PR #136, opened 2020-08-04
REF: Spin off tree -> code compilation into a separate module

adriangb:
This small edit splits off the actual code generation based on the `models.py` classes into its own module. The hope is that it would be possible to swap out the new `compiler.py` for another module that implements the `outputfile_compiler` function with the same signature, thus allowing for generation of arbitrary code based on the `models.py` tree structure.


src/betterproto/plugin/parser.py
```diff
@@ -70,16 +69,6 @@ def generate_code(
 ) -> None:
     plugin_options = request.parameter.split(",") if request.parameter else []
```
adriangb, line 70: Does anyone know how to encode an alternative template/compiler (i.e. to swap out `betterproto.plugin.compiler.py`) as part of the command line options?

adriangb: The goal here is that _at the very least_ someone could copy `compiler.py` to `my_compiler.py` and edit the template (or the rendering method altogether) and pass an option to protoc that would tell `plugin.py` to import `my_compiler.outputfile_compiler` and use that instead of the default one.

nat-n: I guess you have in mind being able to do something like:

```bash
protoc --python_betterproto_out=lib --custom_opt=TYPE_VALIDATION,SILLY_WALKS,WHATEVER -I ...
```

I like the idea of stuff being modular and potentially configurable in this way. But I wouldn't race to provide an API to support fully support this without a real use case.

adriangb: I think this change stands on it's own, and I agree with you that more discussion is needed before providing an API to do the above, but I do want to make sure that this change is at least compatible with that approach.

PR #137, opened 2020-08-06
Code with keywords fixes

Gobot1234:
Some future proofing and fixes for Enum code generation.


src/betterproto/casing.py
```diff
@@ -120,3 +86,8 @@ def camel_case(value: str, strict: bool = True):
 
 def lowercase_first(value: str):
     return value[0:1].lower() + value[1:]
+
+
+def safen(value: str) -> str:
```
nat-n, line 91: this is an Improvement. Though could we have a clearer name? 

Maybe `make_safe` or `sanitize_name`

PR #138, opened 2020-08-14
Implement Message.__bool__

Gobot1234:
This allows you to check if a message has had a non-default value set for it.

```py
>>> bool(MyMessage())
False
>>> bool(MyMessage(my_value='a value'))
True
>>> bool(MyMessage(my_value=''))
False
```
Proposed use case
```py
def parse_body(self, body: MyMessage):
    for item in body.items:
        if item.type == 'some cool value':  
            # do stuff
    if body:
        acknowledge_body()
```


src/betterproto/__init__.py
```diff
@@ -551,6 +551,12 @@ def __setattr__(self, attr: str, value: Any) -> None:
 
         super().__setattr__(attr, value)
 
+    def __bool__(self):
+        for field_name in self._betterproto.meta_by_field_name:
```
nat-n, line 555: how about:
```suggestion
        return any(getattr(self, field_name) for field_name in self._betterproto.meta_by_field_name)
```

Gobot1234: LGTM

Gobot1234:
As things go I think it would be better if you could detect if any value was passed to init but I wasn't too sure how that could be implemented with dataclasses as, following on from the collections note, `bool([""])` would evaluate to True, which I think would make a bit more sense.

cetanu:
So the message is true if any of the fields are populated with a truthy value?
I'm okay with it, but I'm curious what the exact use-case is

Gobot1234:
If #130 was merged however this could be done

src/betterproto/__init__.py
```diff
@@ -551,6 +551,12 @@ def __setattr__(self, attr: str, value: Any) -> None:
 
         super().__setattr__(attr, value)
 
+    def __bool__(self) -> bool:
+        return any(
+            getattr(self, field_name)
```
nat-n, line 556: would it make more sense to say
```suggestion
            getattr(self, field_name) is not None
```
?

nat-n:
I should have said this earlier, but an API change like this should really have a couple tests.

src/betterproto/__init__.py
Gobot1234: I'm just gonna assume at this point #130 will be merged so I'm updating for that

Gobot1234:
I think I'm gonna have to close and reopen this.

PR #139, opened 2020-08-15
Speedups

Gobot1234:
Some speedups/memory improvements to the internals

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -749,9 +749,9 @@ class Message(ABC):
         """
         # Got some data over the wire
         self._serialized_on_wire = True
-
+        proto_meta = self._betterproto
         for parsed in parse_fields(data):
-            proto_meta = self._betterproto
+
             field_name = proto_meta.field_name_by_number.get(parsed.number)
             if not field_name:
                 self._unknown_fields += parsed.raw
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -548,12 +548,6 @@ class Message(ABC):
 
         super().__setattr__(attr, value)
 
-    def __bool__(self) -> bool:
-        return any(
-            self.__raw_get(field_name) is not PLACEHOLDER
-            for field_name in self._betterproto.meta_by_field_name
-        )
-
     @property
     def _betterproto(self):
         """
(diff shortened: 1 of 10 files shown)


nat-n:
Some reasonable micro/-optimisations. This should get the code a little closer to being lintable without much trauma.

src/betterproto/__init__.py
```diff
@@ -756,12 +751,13 @@ def parse(self: T, data: bytes) -> T:
         self._serialized_on_wire = True
 
         for parsed in parse_fields(data):
-            field_name = self._betterproto.field_name_by_number.get(parsed.number)
+            proto_meta = self._betterproto
```
cetanu, line 754: Is it possible to move this outside the for loop?

Gobot1234: I was gonna remove it with the same rational as #140

Gobot1234: Actually maybe here it is ok

cetanu: Anywhere that attributes are being traversed inside loops is an opportunity for a small, easy performance optimization

If it was not inside a loop then I wouldn't bother

src/betterproto/__init__.py
```diff
@@ -754,14 +749,15 @@ def parse(self: T, data: bytes) -> T:
         """
         # Got some data over the wire
         self._serialized_on_wire = True
-
+        proto_meta = self._betterproto
         for parsed in parse_fields(data):
-            field_name = self._betterproto.field_name_by_number.get(parsed.number)
+
```
nat-n, line 754: empty line

src/betterproto/__init__.py
```diff
@@ -261,7 +255,7 @@ class Enum(enum.IntEnum):
     def from_string(cls, name: str) -> int:
         """Return the value which corresponds to the string name."""
         try:
-            return cls.__members__[name]
+            return cls._member_map_[name]
```
nat-n, line 258: what's this change about?

Gobot1234: Accessing `enum. _member_map_ ` is a more direct call https://github.com/python/cpython/blob/master/Lib/enum.py#L365 since read-only isn't important

nat-n:
Looks good, just need to rebase?

Gobot1234:
Righto

nat-n:
> Do I?

I'm guessing that would fix the CI run.

Gobot1234:
It's just from the black PR and the format changing.

nat-n (approved):

PR #140, opened 2020-08-15
Fixes for Python 3.9

Gobot1234:
`type(field_type)` can now be `<class 'list'>` breaking the old method of checking.


src/betterproto/__init__.py
```diff
@@ -818,9 +818,10 @@ def to_dict(
         """
         output: Dict[str, Any] = {}
         field_types = self._type_hints()
-        for field_name, meta in self._betterproto.meta_by_field_name.items():
+        proto_meta = self._betterproto
```
nat-n, line 821: I don't think this it worth it here. I'm guessing the point is that for the cost of adding an assignment in scope, you save a lookup step for the value later?

However the perf value of this here is certainly below the level of noise for even a contrived test case, and I feel it makes the code less clear by adding a level of indirection without really clarifying anything in the process.

nat-n:
Looks like the new CI config doesn't work?

Gobot1234:
Yeah I'm not too sure what's up with it

src/betterproto/__init__.py
Gobot1234: Ok

Gobot1234:
Looks like it's been fixed

Gobot1234:
Not a clue what to do about this

Gobot1234:
Can we cherry pick the actual fix to merged and leave the tests as a separate PR to be merged once 3.9 plays better?

nat-n:
> Can we cherry pick the actual fix to merged and leave the tests as a separate PR to be merged once 3.9 plays better?

That would break the pipeline for other PRs.

Gobot1234:
Building wheels for dependencies seems like an impossible task in actions.

src/betterproto/__init__.py
```diff
@@ -926,9 +926,10 @@ def to_dict(
         """
         output: Dict[str, Any] = {}
         field_types = self._type_hints()
+        defaults = self._betterproto.default_gen
         for field_name, meta in self._betterproto.meta_by_field_name.items():
             field_type = field_types[field_name]
```
nat-n, line 931: doesn't look like this variable has good reason to be here anymore.

nat-n (approved):
Why is the lockfile changed here?

Gobot1234:
It was due to me experimenting with checking system platform specific dependencies.

src/betterproto/__init__.py
Gobot1234: I thought it would be better here to be consistent with `field_types`

Gobot1234:
```py
>>> import typing
>>> def func(t: typing.List[str]):
...     pass
... 
>>> typing.get_type_hints(func)["t"]
typing.List[str]
>>> type(typing.get_type_hints(func)["t"])
<class 'typing._GenericAlias'>
>>> type(typing.List)
<class 'typing._SpecialGenericAlias'>
>>> type(typing.List) is type(typing.get_type_hints(func)["t"])
False
```
I think this is the actual bug.

src/betterproto/__init__.py
nat-n: Given the the value is no longer needed till near the end of the function, if at all, I think it's best to remove this assignment here now.

nat-n (changes requested):
> Why is the lockfile changed here?

> It was due to me experimenting with checking system platform specific dependencies.

Better not to update the lockfile in this change if there's no good reason to do so.

src/betterproto/__init__.py
Gobot1234: It's used on https://github.com/Gobot1234/python-betterproto/blob/fixes/src/betterproto/__init__.py#L932 🤔

nat-n: Just to make sure we're on the same page, I'm referring to `field_type`, not `defaults`.

Gobot1234: I didn't add that

nat-n: Indeed. Though you changed how it's used, so it would be good to clean it up in the process.

Gobot1234: But wouldn't that mean the method was called multiple times in the loop when it could be called just once?

nat-n: I'm talking about `field_type = field_types[field_name] ` which as it stands after you change is defined *inside* the loop and then maybe used once (though not used at all on most code paths).

nat-n (approved):

PR #141, opened 2020-08-16
QOL fixes

Gobot1234:
Some QOL fixes/improvements/simplifications of code, also adds in some missing type annotations.

General checklist:
- Use `ImportError.name` to get missing imports for compiler dependencies.
- Make plugin imports as relative as possible.
- Fix some docstrings/comments.
- Add and correct some type hints.
- Simplify some overly complex conditionals.
- Use f-strings where I've noticed.
- Use top level constants where applicable.


src/betterproto/plugin/main.py
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env python
-import sys
+
 import os
+import sys
 
 from google.protobuf.compiler import plugin_pb2 as plugin
 
-from betterproto.plugin.parser import generate_code
+from .parser import generate_code
```
nat-n, line 8: I'm not certain but I think switching to relative imports within the plugin might break invoking this module as a script which is otherwise supported.

src/betterproto/plugin/parser.py
```diff
@@ -1,7 +1,9 @@
 import itertools
 import pathlib
 import sys
-from typing import List, Iterator
+from typing import Iterator, List, Tuple, Union
+
+from google.protobuf.descriptor import Descriptor
```
nat-n, line 6: if only used for type annoation then this import should be under `if TYPE_CHECKING:`

nat-n (commented):
Thanks for this. There are some nice improvements here, also a few details worth iterating on.

Gobot1234:
I'll give these a look over I think I also have a few more to add

src/betterproto/__init__.py
```diff
@@ -430,12 +429,6 @@ def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
 
 
 class ProtoClassMetadata:
-    oneof_group_by_field: Dict[str, str]
```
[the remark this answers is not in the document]
Gobot1234: IMO it saves on duplication and looks nicer.

src/betterproto/grpc/grpclib_client.py
```diff
@@ -1,19 +1,21 @@
-from abc import ABC
 import asyncio
-import grpclib.const
+from abc import ABC
```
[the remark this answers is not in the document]
Gobot1234: This was isort.

src/betterproto/plugin/models.py
```diff
@@ -155,7 +143,7 @@ class ProtoContentBase:
     comment_indent: int = 4
     parent: Union["Messsage", "OutputTemplate"]
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
```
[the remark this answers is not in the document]
Gobot1234: Mypy isn't a fan of functions apart from `__init__` that don't define return types.

src/betterproto/plugin/main.py
Gobot1234: Seems like a good call

Gobot1234: It should be fine as tests seem to run right?

src/betterproto/__init__.py
```diff
@@ -1191,15 +1179,15 @@ def timestamp_to_json(dt: datetime) -> str:
         if (nanos % 1e9) == 0:
             # If there are 0 fractional digits, the fractional
             # point '.' should be omitted when serializing.
-            return result + "Z"
+            return f"{result}Z"
         if (nanos % 1e6) == 0:
             # Serialize 3 fractional digits.
-            return result + ".%03dZ" % (nanos / 1e6)
+            return f"{result}.%03dZ" % (nanos / 1e6)
```
Gobot1234, line 1185: I'm not too sure how to make this completely and f-string if anyone knows how to do it feel free to commit it.

src/betterproto/plugin/main.py
nat-n: I don't think there are any feature tests for that invoke the plugin in that way. It's generally not the most convenient way to do it but still worth supporting.

src/betterproto/plugin/models.py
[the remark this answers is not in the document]
nat-n: Really? Usually it's pretty good at inferring that if a function doesn't even have a return statement then it's return value is implicitly None.

src/betterproto/grpc/grpclib_client.py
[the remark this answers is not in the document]
nat-n: Odd that abc comes after asyncio no?

src/betterproto/__init__.py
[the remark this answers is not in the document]
nat-n: Having it on the class serves as a single reference point for documenting the class rather than having the types mixed in the the logic. I'm not sure why you'd need to duplicate it anywhere.

src/betterproto/grpc/grpclib_client.py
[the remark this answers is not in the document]
Gobot1234: It's cause its a `from import`

src/betterproto/__init__.py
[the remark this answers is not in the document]
Gobot1234: Well mixing it in with the logic can again in my opinion make it clearer what the attribute is

src/betterproto/__init__.py
nat-n: If I'm not mistaken `f"{result}.{ int(nanos // 1e6) :03d}Z"` is what you're looking for.

asv.conf.json
```diff
@@ -1,157 +1,12 @@
 {
-    // The version of the config file format.  Do not change, unless
```
nat-n, line 2: Why strip down this config file? I'd leave this change out. Having a documented base config is kind of useful for this sort of thing.

nat-n (commented):
Looks good, except the one comment.

asv.conf.json
Gobot1234: Ok

src/betterproto/__init__.py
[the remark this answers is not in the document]
nat-n: I think it's more important to document the class, since that's conventionally how types work, being able to get the full picture of a class' content at a glance without scanning through the logic is important. Honestly I find the logic ends up harder to read jammed full of types annotations, whereas referencing the class as needed doesn't add much burden to reading.

Gobot1234:
Should be done now 👍

.gitignore
```diff
@@ -14,5 +14,5 @@ output
 .idea
 .DS_Store
 .tox
-.venv
+venv
```
nat-n, line 17: .venv is a commonly used location for virtualenvs. Maybe it's worth ignoring `venv` as well, but not instead of.

nat-n (changes requested):
One more thing.

.gitignore
Gobot1234: Must have slipped in.

Gobot1234:
Should benchmarks also be added to the black tests or is that outside the scope of this PR?

nat-n:
> Should benchmarks also be added to the black tests or is that outside the scope of this PR?

sure, why not.

nat-n (approved):
Alright, good work.

PR #142, opened 2020-08-24
Implement Message.__bool__

Gobot1234:
Carries on from #138, I messed up the git tree.


danielgtaylor:
This look interesting. I like the general idea for sure.

Given proto's use of default zero values I wonder if it's more useful to know:

1. Has anything ever been set on this?
2. Is the message equivalent to zero (i.e. all fields are their zero value)?

This PR seems to implement 1 unless I'm mistaken. Thoughts?

Gobot1234:
Am I mistaken in thinking that returning self._serialized_on_wire would achieve this?

Gobot1234:
I should have this working as a collection would now, so if a value is set to anything and then has its value set to its default it will return `False`

nat-n:
@Gobot1234 I think we can merge this in principal, but it's significant an API feature that it should really be documented. Would you mind?

Gobot1234:
Ah yes of course.

Gobot1234:
Sorry just realised I haven't added it to the Sphinx special members directive yet

src/betterproto/__init__.py
```diff
@@ -516,6 +516,10 @@ class Message(ABC):
         .. describe:: bytes(x)
 
             Calls :meth:`__bytes__`.
+
+        .. describe:: bool(x)
```
nat-n, line 520: Maybe it would be appropriate to explain somewhere around here that the truthiness of a message object is similar to a collection in that it depends on whether it contain any (non-default) values.

Gobot1234: Should be done.

nat-n (approved):

PR #144, opened 2020-08-27
Remove @dataclass, add slots

### LINKED ISSUE #50: Optimize dataclasses with __slots__
[Adding __slots__ to a class](https://stackoverflow.com/questions/472000/usage-of-slots) improves memory usage and performance of attribute access, at the expense of not being able to set arbitrary attributes on the resulting objects.

[This project](https://github.com/starhel/dataslots/blob/master/dataslots/__init__.py) demonstrates how dataclasses can be augmented to use __slots__, whilst still having class attributes set (normally adding the __slots__ attribute to a class definition would cause an error with dataclassess that declare default values).

Making generated dataclasses use slots in this way would be a small breaking change, though I don't see how it could be too inconvenient to work with, and I expect the performance improvement would be worthwhile.

Gobot1234:
Adds a custom metaclass for Message to quickly create dataclasses for Messages whilst also slotting them. Internally should still be basically the same as a dataclass.

Benefits:

- This will almost certainly be faster than the current approach, I will need to run speed tests on it before I am 100% sure.
- Removes an import from all generated messages, and improves visual clarity.

This is incorporates 2 breaking changes:

- Removing `@dataclass` from all generated outputs, since this is for V2 however I think it is fine, there are a couple of options if you want to give warning like monkey patching the dataclass decorator to do nothing for Message subclasses.
- Fields cannot now be access outside of instances of themselves, pretty minor change, this shouldn't really affect anyone as they (I think) would always be `betterproto.Placeholder`, but if necessary it could be remedied by implementing `MessageMeta.__getattribute__`.

Closes #50


abn:
I am not sure if optimising for performance over usability by default is a good thing. While, I can understand the motivation for this change, I feel that dropping dataclasses means you are loosing a lot of niceties that it brings to the table. I am also unclear on how much of a gain we are talking here. We might also want to talk about making it optional, ie. dataclasses vs slots at compile time.

As an additional note, the alternative is to move from dataclasses to something that already fascilitates slots. The [attrs](https://www.attrs.org/en/stable/examples.html#slots) project is a good option for that. Obviously, everything has trade-offs. :)

Gobot1234:
It isn't dropping anything internally, it just removes the decorator.

abn:
> It isn't dropping anything internally, it just removes the decorator.

Ah, I might have spoke too soon then. Will go through the changes. Ignore my reactionary comment. :)

Gobot1234:
Some tests

# Results

Tabled results are for a Message of unless specified otherwise:
```py
class Message(betterproto.Message):
    foo: int = betterproto.uint32_field(0)
    bar: int = betterproto.uint32_field(1)
    baz: int = betterproto.uint32_field(2)
```

## Important results:

|  | Old implementation |New slotted implementation |
|--|--|--|
| Size of [1](#size-of-instance) | 576 bytes | 256 bytes |
| Runtime Creation [2](#runtime-overhead) | 741 usec | 38.4 usec |
| Attribute Access [3](#size-of-instance) | 71.4 nsec |  43.8 nsec |


### Size of Instance

Tested with 0, 3, 5, 10, 50 field large message instances.

| Size | 0 | 3 | 5 | 10 | 50 | 
|--|--|--|--|--|--|
| Old | 152 | 576 | 728 | 1096 | 3848 |
| New | 208 | 256 | 272 | 312 | 632 |

Slots make messages significantly more linear in the amount of memory usage they end up using 

### Runtime overhead

```
Old raw times: 74.1 sec, 90.5 sec, 74.1 sec, 82 sec, 87.8 sec

100000 loops, best of 5: 741 usec per loop
```
```
Slotted raw times: 4.96 sec, 4.72 sec, 4.17 sec, 3.84 sec, 4.1 sec

100000 loops, best of 5: 38.4 usec per loop
```
This is ridiculously faster with MessageMeta at around 20x

### Attribute Access

```
Old raw times: 13.2 msec, 8.81 msec, 7.48 msec, 7.28 msec, 7.14 msec

100000 loops, best of 5: 71.4 nsec per loop
```
```
Slotted raw times: 4.38 msec, 6.24 msec, 5.9 msec, 4.46 msec, 5.67 msec

100000 loops, best of 5: 43.8 nsec per loop
```
So it's about 60% faster with slots

### Instantiation 
```
Old raw times: 2.34 sec, 1.03 sec, 1 sec, 1.61 sec, 2.55 sec

100000 loops, best of 5: 10 usec per loop
```
```
Slotted raw times: 1.25 sec, 1.1 sec, 1.14 sec, 1.12 sec, 1.2 sec

100000 loops, best of 5: 11 usec per loop
```
Currently instantiation with the current implementation is faster, however only marginally and I think the instantiation can be improved on the new model, or if not we can just use the dataclasses implementation, if it doesn't hurt runtime overhead particularly.

Gobot1234:
I've optimized init more and it's now faster (by a couple of usec) than the dataclasses approach.

Gobot1234:
Tests again will fail until #130 is merged

src/betterproto/__init__.py
```diff
@@ -577,16 +644,18 @@ def __getattribute__(self, name: str) -> Any:
     def __setattr__(self, attr: str, value: Any) -> None:
         if attr != "_serialized_on_wire":
             # Track when a field has been set.
-            self.__dict__["_serialized_on_wire"] = True
+            super().__setattr__("_serialized_on_wire", True)
 
-        if hasattr(self, "_group_current"):  # __post_init__ had already run
+        if hasattr(self, "_group_current"):
```
Gobot1234, line 649: This can be removed

src/betterproto/__init__.py
```diff
@@ -577,16 +644,18 @@ def __getattribute__(self, name: str) -> Any:
...
             if attr in self._betterproto.oneof_group_by_field:
                 group = self._betterproto.oneof_group_by_field[attr]
                 for field in self._betterproto.oneof_field_by_group[group]:
                     if field.name == attr:
                         self._group_current[group] = field.name
                     else:
-                        super().__setattr__(field.name, PLACEHOLDER)
+                        super().__setattr__(
+                            field.name, self._get_field_default(field.name)
+                        )
```
Gobot1234, line 658: This needs to be reverted

adriangb:
Have you evaluated [Pydantic](https://pydantic-docs.helpmanual.io/) instead? I'm not sure how it would work out speed-wise, but making `Message` inherit from Pydantic's `BaseModel` could be interesting.

Gobot1234:
What benefits would using pydantic bring over this implementation, I really don't see how it could get much better in terms of performance which is what this is aiming for.

adriangb:
I don't *think* it will be better in terms of performance (although it might be worth trying) but Pydantic does have a very rich feature set that IMO could be very useful.

Just curious, I see that this comes from #50. Has there been any use case studied where the performance was an issue and this level of performance improvement resolved things?

Gobot1234:
I don't think performance is a particularly large issue but I see no harm in optimising this kind of stuff.

adriangb:
> I don't think performance at runtime is a particularly large issue but I see no harm in optimising this kind of stuff.

There is a **lot** of harm in implementing features (performance optimization is a feature) for the sake of implementing features. Without concrete real-world use cases for implementing a feature, it should not be implemented.

Gobot1234:
I don't agree with that in this case as this **does not** change anything internally. In my eyes better performance is better performance. I thing my previous comment might have slightly down played the benefits this adds in terms of attribute access and memory usage.

adriangb:
> this **does not** change anything internally.

I see several hundred lines of pretty complex code being changed (without counting the removal of `@dataclass`).

> the benefits this adds in terms of attribute access and memory usage.

Do you have a real world use case where that performance difference is actually important? Does this level of performance improvement resolve or enable those use cases? We could get even better performance by re-writing this in C if we wanted to, but we shouldn't do that just because we can.

adriangb:
Hey @Gobot1234, would you be able to post your benchmarks? I'd be happy to turn them into [asv](https://github.com/airspeed-velocity/asv) benchmarks and submit that as a separate PR so that it's easy for anyone to evaluate the impact their proposed changes have on performance.

Gobot1234:
Sure

Gobot1234:
https://repl.it/@Gobot1234/TriangularSuperiorReentrant#main.py

adriangb:
I added these benchmarks into #148

adriangb:
Another option to consider: replace the `dataclasses` dependency with [dataclassy](https://github.com/biqqles/dataclassy) which already has the ability to use `__slots__` to speed things up _and_ is compatible with Python 3.6 _and_ has other nice features.

adriangb:
> We could get even better performance by re-writing this in C if we wanted to

I tested the output from the built in protoc python plugin, and it is an order of magnitude faster than betterproto (because it's ugly, inflexible and has C/C++ portions).

Gobot1234:
Yes I would have assumed so, that's why I wanted to write an accelerator in Rust

Gobot1234:
There are also still places I can see room for optimisations, however with how polarising this PR has been I wouldn't expect them to be particularly likely to be merged.

src/betterproto/templates/template.py.j2
```diff
@@ -54,11 +52,11 @@ class {{ message.py_name }}(betterproto.Message):
     {% endif %}
 
     {% if message.deprecated or message.deprecated_fields %}
-    def __post_init__(self) -> None:
+    def __init__(self, *args, **kwargs) -> None:
```
nat-n, line 55: the `__init__` should have the correct argument list.

src/betterproto/__init__.py
```diff
@@ -505,30 +545,54 @@ class Message(ABC):
     _serialized_on_wire: bool
     _unknown_fields: bytes
     _group_current: Dict[str, str]
+    __dataclass_fields__: Mapping[str, dataclasses.Field]  # Technically should be Final
```
nat-n, line 548: Don't think this comment is useful. Betterproto won't be dropping 3.7 support for a long time.

src/betterproto/__init__.py
```diff
@@ -505,30 +545,54 @@ class Message(ABC):
...
+            # We have to set these here to allow for them to be editable
+            set_attribute(field.name, field.default)
 
-        # Set current field of each group after `__init__` has already been run.
-        group_current: Dict[str, Optional[str]] = {}
         for field_name, meta in self._betterproto.meta_by_field_name.items():
+            if args or kwargs:
+                try:
+                    set_attribute(
+                        field_name, (args or kwargs).pop(0 if args else field_name)
```
nat-n, line 565: this is a bit cryptic. Maybe having more distinct code paths for args vs kwargs would be better. In fact I think the control flow within this loop is worth revisiting more generally.

nat-n (changes requested):
Thanks for spending time on this. I think performance is important, which is why I created #50, and I suspect a metaclass broadly along these lines is the best approach. However it's also a significant architectural change, and not worth rushing, which is part of why I've been slow to weigh in. As others have pointed out I also think it's important to weigh optimisation against other concerns, particularly correctness and maintainability, which are two areas in which this project more urgently needs to improve.

I think the implementation here is generally good. The biggest issue I see is the loss of args in `__init__` which is easy to fix.

Wrt optimisation there are some good gains here, although bear in mind what I think it probably most important to optimise is actually initialisation & setting attributes, since this is always in the critical path of real use-cases.

What do you mean by "Fields cannot now be access outside of instances of themselves"?

Overall I'd be happy to merge this with a little more work, although because of the nature of this change I'd rather wait till after the next beta release which I'll hopefully get around to soon.

src/betterproto/__init__.py
Gobot1234: I was thinking of PRing support for typing_extensions at some point to stop relying on class dunders, which supports `Final`, but I'll remove it for now.

src/betterproto/__init__.py
```diff
@@ -1056,6 +1122,9 @@ def delta_to_json(delta: timedelta) -> str:
 
 
 class _Timestamp(Timestamp):
+    seconds: int = int64_field(1)
```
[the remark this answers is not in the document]
Gobot1234: They aren't picked up and added to the class's fields iirc?

src/betterproto/templates/template.py.j2
Gobot1234: That's gonna be a challenge. I'll see what I can do.

src/betterproto/__init__.py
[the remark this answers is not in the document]
nat-n: Normally inheritance works with dataclasses

src/betterproto/__init__.py
Gobot1234: This was the fastest way I could think of to make the `__init__` run, if you aren't a fan of this I'll do some exec grossness to do in effect what happens in the code block in https://docs.python.org/3/library/dataclasses.html

Gobot1234:
So when I said fields can only be accessed outside of instances of themselves the fields when not set are `types.MemberDescriptor` so field info cannot be accessed.

src/betterproto/__init__.py
Gobot1234: I think it would be better rather than doing what I'm doing ATM to make generated protos have a generated `__init__`?

src/betterproto/__init__.py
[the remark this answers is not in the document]
Gobot1234: Protobufs however don't support this behaviour from the research I've done, if this is actually part of the spec I would consider it but it would slow stuff down.

src/betterproto/__init__.py
Gobot1234: As currently the code is really really gross 
```py
    def __new__(
        mcs, name: str, bases: Tuple[type, ...], attrs: Dict[str, Any]
    ) -> Type["Message"]:
        annotations = attrs.get("__annotations__", {})
        attrs["__slots__"] = tuple(annotations)
        # Slot the class

        fields = {}
        for annotation, type in annotations.items():
            field: Optional[dataclasses.Field] = attrs.pop(annotation, None)
            # Remove field class variables from the class namespace
            if field is not None:
                field.name = annotation
                field.type = type
                field._field_type = dataclasses._FIELD
                fields[annotation] = field

        message_class = super().__new__(mcs, name, bases, attrs)
        message_class.__dataclass_fields__ = fields
        body = "\n".join(
            f"    self.{field_name} = {field_name}" for field_name in fields
        )
        exec(
            f"""
def __init__(self, {", ".join(f"{field_name} = PLACEHOLDER" for field_name in fields)}):
    object.__setattr__(self, "_group_current", {'{}'})
    object.__setattr__(self, "_init_running", True)
{body}
    # Now that all the defaults are set, reset it!
    object.__setattr__(self, "_unknown_fields", b"")
    object.__setattr__(self, "_init_running", False)
    object.__setattr__(self, "_serialized_on_wire", False)
message_class.__init__ = __init__
"""
        )
        return message_class
```

Gobot1234: Although having the init be per class kind of defeats the purpose of originally having these as dataclasses so maybe this is the only way to do it

src/betterproto/templates/template.py.j2
nat-n: Unless I'm missing something I think it should be fairly straight forward. If you provide a get_kwargs method on the relevant compiler class that formats the fields and their default values as a kwargs list, and then call it here and on line 59. Does that sound feasible?

Gobot1234: I've implemented it, it wasn't anywhere near as bad as I expected.

src/betterproto/__init__.py
nat-n: I'm not sure what the best solution is just by looking at it, there are clearly a few reasonable approaches. My issue with the existing implementation is just that the control flow could be cleaner.

src/betterproto/__init__.py
```diff
@@ -563,7 +563,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         all_sentinel = True  # Keep track of whether every field was default
         group_current: Dict[str, str] = {}
 
-        for field in self.__raw_get(dataclasses._FIELDS).values():
```
nat-n: is this intentional?

Gobot1234: I was trying to fix the git tree so just added something to remove later

Gobot1234:
poe bench returns 
```
All benchmarks:

       before           after         ratio
     [f10bec47]       [222a105b]
     <master>         <dataslots>
      1.99±0.09μs      1.93±0.2μs      0.96 benchmarks.BenchMessage.time_attribute_access
      10.5±0.4μs       9.01±0.2μs      0.86 benchmarks.BenchMessage.time_attribute_setting
      17.3±1μs         13.1±0.2μs      0.76 benchmarks.BenchMessage.time_init_with_values
      17.4±0.5μs       7.96±0.6μs      0.46 benchmarks.BenchMessage.time_instantiation
      722±30μs          293±9μs        0.41 benchmarks.BenchMessage.time_overhead
      22.5±2μs          22.2±1μs       0.99 benchmarks.BenchMessage.time_serialize
	    928               328          0.35 benchmarks.MemSuite.mem_instance
```

Gobot1234:
Not too sure why the getattr call isn't much of an improvement as it should be the same just `__slot__` boosted?

Gobot1234:
I'm going to close this for now. I'm sure this can be done better but it requires a lot of work.

PR #149, opened 2020-09-17
Added example of calling protoc from python

spacecowboy:
Took me a couple of minutes to figure out that this project didn't generate a binary called `protoc` when installed via pip so I added an example to the readme which would have helped me in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,14 @@ message Greeting {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nat-n:
Good idea. thanks :)

nat-n (approved):

PR #152, opened 2020-09-26
Custom options test

adriangb:
@nat-n concrete example from slack discussion


src/betterproto/plugin/main.py
```diff
@@ -13,6 +13,7 @@ def main():
     data = sys.stdin.buffer.read()
 
     # Parse request
+    # import extension_pb2
```
adriangb, line 16: If you uncomment this, you get:

```console
Generated output for 'custom_options'
[extension] {
  max_size: 40
}
```

When running `python -m tests.generate -v`. Otherwise you don't (because protoc does not pick up the custom options)

nat-n:
I pushed another commit that demonstrates reading a custom option by using betterproto classes to load the protoc message (instead of relying on the classes generated by the default plugin). A few additional steps would be required for a full solution.

Running `poe generate -v custom_options` should output a dump of the CodeGeneratorRequest in which you can find `MyOptions(max_size=40)`.

It's a bit hacky, and broken besides that, but I think this general approach is ideal going forward.

src/betterproto/__init__.py
```diff
@@ -802,7 +803,12 @@ def _postprocess_single(
                 elif meta.wraps:
                     # This is a Google wrapper value message around a single
                     # scalar type.
-                    value = _get_wrapper(meta.wraps)().parse(value).value
+                    wrapper = _get_wrapper(meta.wraps)().parse(value)
+                    if meta.wraps == "enum":
```
Gobot1234, line 807: Shouldn't this use TYPE_ENUM?

nat-n: yea, that would make sense. Feel free to make commits on top of this one if it's useful to do so.

Gobot1234: I do not have the brains for that lol.

adriangb:
I took a first look at this @nat-n 

It looks like you created the backend to store the custom options, right?

Regarding the monkey patching, it can probably be made easier by making a factory-function like `factory_function(raw_object: whatever_we_load) -> MyOptions` classes.

I'm more concerned about being able to parse out the custom-options from the `CodeGeneratorRequest` object.
 - Do you think we'll be able to pick these out easily? Will we be getting entire files or messages?
 - Do we need to worry about mixed files (i.e. have extension messages but also normal messages, which themselves could depend on other extensions)?
 - Other strange situations?
I'll try to take a stab at this sometime next week, but I'll be diving into `CodeGeneratorRequest` for the first time really, so let me know if you have any ideas.

I'm also not sure what you mean by:
```
    - monkey patching FieldOptions:
        e.g. FieldOptions.__init__ = ...
             FieldOptions.__dataclass_fields__['extension'] = field(...)
             FieldOptions.__annotations__['extension'] = MyOptions
```

src/betterproto/plugin/main.py
```diff
@@ -12,16 +14,19 @@ def main():
...
+    # TODO: scan request to find Option extensions,
+    #       and if found then monkey patch the relevant messages and parse data again
 
     dump_file = os.getenv("BETTERPROTO_DUMP")
     if dump_file:
         dump_request(dump_file, request)
 
     # Create response
-    response = plugin.CodeGeneratorResponse()
+    response = CodeGeneratorResponse()
```
adriangb, line 29: @nat-n I'm inspecting the `response` option here with line 775 of `src/betterproto/lib/google/protobuf/__init__.py` commented out, and I'm seeing:
`request.proto_file` -> 3 files, first one is google types, second is `extension.proto` and 3rd is `double.proto`.
`request.proto_file[2].message_type[0].field[0]` -> this is the `email` field
```
FieldDescriptorProto(name='email', number=3, label=1, type=9, type_name='', extendee='', default_value='', oneof_index=0, json_name='email', options=FieldOptions(ctype=0, packed=False, jstype=0, lazy=False, deprecated=False, weak=False, uninterpreted_option=[]))
```

I unfortunately do not see any information in `FieldOptions` that would suggest that we need to monkey-patch it with the extension reference. My understanding is we would have to have some indication (via the `uninterpreted_option` or something) that we need to monkey patch the extension.

And how would this monkey patching take place? Are we doing:
```python3
request = CodeGeneratorRequest().parse(data)
# somehow figure out that request.proto_file[2].message_type[0].field[0] requires a monkey patch
extension = generate_extenion_dataclass(request.proto_file[1].message_type[0])  # assuming this encapsulates 
request.proto_file[2].message_type[0].field[0].options.load_extension(extension)  # assuming load_extension calls self.__dataclass_fields__...
# and now go through a bytes -> parse loop?
data = bytes(request)
request = CodeGeneratorRequest().parse(data) 
```

nat-n:
> It looks like you created the backend to store the custom options, right?

You mean creating `betterproto.lib.google.protobuf.compiler`? Yup I ran betterproto on [this proto file](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/plugin.proto) and made some tweaks for the PoC. Not 100% sure it's the correct approach since it appears to miss some details about enums but that doesn't matter for this specific case.

> Regarding the monkey patching, it can probably be made easier by making a factory-function like `factory_function(raw_object: whatever_we_load) -> MyOptions` classes.

Yes, encapsulating programatic creation of classes referenced by the extension (not really any monkey patching involved at this point) with something like that would make sense.

> I'm more concerned about being able to parse out the custom-options from the `CodeGeneratorRequest` object.
> 
> * Do you think we'll be able to pick these out easily? Will we be getting entire files or messages?
> * Do we need to worry about mixed files (i.e. have extension messages but also normal messages, which themselves could depend on other extensions)?
> * Other strange situations?
>   I'll try to take a stab at this sometime next week, but I'll be diving into `CodeGeneratorRequest` for the first time really, so let me know if you have any ideas.

From my commit here running `poe generate -v custom_options` generates the custom_options python module as usual and also dumps the betterproto object for the CodeGeneratorRequest in which you can see that all all field descriptor options not include your new "extension" property with type MyOptions, and the in case of the _email_should_have_options_ field (at `request.proto_file[2].message_type[0].field[0]`) with the value of max_size set to 40. From there you just need to ensure the options are passed to the python code generation logic to interpret. 

For me the relevant descriptor comes out looking like (minus the formatting):

```python
FieldDescriptorProto(
    name="email_should_have_options",
    number=3,
    label=1,
    type=9,
    type_name="",
    extendee="",
    default_value="",
    oneof_index=0,
    json_name="emailShouldHaveOptions",
    options=FieldOptions(
        ctype=0,
        packed=False,
        jstype=0,
        lazy=False,
        deprecated=False,
        weak=False,
        uninterpreted_option=[],
        extension=MyOptions(max_size=40),
    ),
)
```

Does this make sense?

> 
> I'm also not sure what you mean by:
> 
> ```
>     - monkey patching FieldOptions:
>         e.g. FieldOptions.__init__ = ...
>              FieldOptions.__dataclass_fields__['extension'] = field(...)
>              FieldOptions.__annotations__['extension'] = MyOptions
> ```

So for the above to be true, the `extend google.protobuf.FieldOptions { ... }` clause in your proto input has to actually extend the FieldOptions type used by the compiler, or else when reading the message from protoc it'll ignore your custom extension values as an unknown field (as with any other backwards compatible update to a protobuf message).

So concretely this means that in the main.py of the plugin we need to:

1. walk the input to find extensions for FieldOptions and other message types from the `google.protobuf` package.
2. Monkey patch the extensions onto the relevant betterproto generated message dataclasses (from `betterproto.lib.google.protobuf.compiler`), vie the procedure you quoted. From my minimal experimentation it appears that you can extend a dataclass (at least sufficiently for our purposes) just by updating those three attributes of the class. This would also require programatically creating any new message classes referenced from the extension (such as MyOptions in this example) as discussed above.
3. Parse the input again. Since the betterproto parser is using dataclasses that have now been updated to know about your custom options it will be able to parse the values of those custom options from the input (instead of ignoring them).
4. The custom options should then be available to the rest of the plugin – code generation etc – as usual.

Is any of this unclear?

What I've done validates the key assumptions about how the parser/monkey-patching etc should work, the rest should be relatively straight-forward (if not simple). I think the hard part the remains is to fix the remaining issues with the plugin itself using betterproto generated classes instead of the default ones.

I'm interested in working on this, so I'm happy to help, but I'm not sure when I'll have bits of time to commit to it so don't wait up on me if you have time pressure to arrive at a solution.

adriangb:
Thank you for the extensive walkthrough. I think we're generally on the same page. Comments inline below.

> * walk the input to find extensions for FieldOptions and other message types from the `google.protobuf` package.
> * Monkey patch the extensions onto the relevant betterproto generated message dataclasses (from `betterproto.lib.google.protobuf.compiler`), vie the procedure you quoted.
What's unclear to me here is: are we editing the central `betterproto.lib.google.protobuf.compiler` (like you did here), or are we editing/replacing the individual instances attached to fields? It sounds like it's the central one. In that case, is this limited to 1 extension?

So basically, I'm confused about how we would handle multiple different `FieldOptions` extensions using this approach. I must be missing something.

nat-n:
> So basically, I'm confused about how we would handle multiple different `FieldOptions` extensions using this approach. I must be missing something.

Yes,  we're essentially patching the plugin at runtime. My understanding of the [relevant docs](https://developers.google.com/protocol-buffers/docs/proto#customoptions) is that custom options are a bit of a hack that emerges from protobuf's self-referential nature. You chose to extend `google.protobuf.FieldOptions` with a new field called `extension` at `1010`, which can co-existing with an extension of a different name/number. You could add another extension as another location on the message. But if you wanted to run the compiler including another proto with another custom extension with the same name or number then you'd have a problem. I don't think there's any way around this, other than trying/hoping to avoid such collisions. This probably means that using a high number and namespacing the field name is a good idea.

adriangb:
How would you set multiple `extension` attributes?

```python3
class FieldOptions(betterproto.Message):
    extension: "MyOptions" = betterproto.message_field(1010)
    extension: "MyOptions2" = betterproto.message_field(1011)  # this is just overwriting the first one
```
That is, how would be add multiple values for the same attribute?

Or is this a oneof type thing?

nat-n:
I image something more like:

```python
class FieldOptions(betterproto.Message):
    ....
    acmecorp_validation: "AcmeValidationRules" = betterproto.message_field(4876234)
    obscurelang_hints: "ObscureLangHints" = betterproto.message_field(9823743)
```

except of course we don't define a new FieldOptions dataclass, we extend the existing one in place.

makes sense?

adriangb:
I see! I was thinking `extension` was a keyword. The example you give corresponds to:

```protobuf
import "acmecorp_validation.proto";

message Person {
    string email_should_have_options = 3 [(acmecorp_validation).max_size = 40];
} 
```

(or something like that)

Right?

PR #161, opened 2020-10-18
Use betterproto in the plugin

nat-n:
This builds on work started for #152, and will make it easier to support custom options. @adriangb 

---

Make plugin use betterproto generated classes internally

This means the betterproto plugin no longer needs to depend directly on
protobuf.

This requires a small runtime hack to monkey patch some google types to
get around the fact that the compiler uses proto2, but betterproto
expects proto3.

Also:
- regenerate google.protobuf package
- fix a regex bug in the logic for determining whether to use a google
  wrapper type.
- fix a bug causing comments to get mixed up when multiple proto files
  generate code into a single python module

--- 
Update deps & add generate_lib task

- Remove plugin dependency on protobuf since it's no longer required.
- Update poethepoet to for better pyproject toml syntax support
- Add handy generate_lib poe task for maintaining generated libs


src/betterproto/plugin/models.py
```diff
@@ -311,8 +339,20 @@ def is_map(
 
 
 def is_oneof(proto_field_obj: FieldDescriptorProto) -> bool:
-    """True if proto_field_obj is a OneOf, otherwise False."""
-    return proto_field_obj.HasField("oneof_index")
+    """
+    True if proto_field_obj is a OneOf, otherwise False.
+
+    .. warning::
```
nat-n, line 345: this is referring to monkey_patch_oneof_index

src/betterproto/plugin/models.py
```diff
@@ -139,6 +165,7 @@ def get_comment(
 class ProtoContentBase:
     """Methods common to MessageCompiler, ServiceCompiler and ServiceMethodCompiler."""
 
+    source_file: FileDescriptorProto
```
nat-n, line 168: turns out tracking the source_file is necessary to ensure we look in the right file to find comments, in case there are multiple source proto files in scope.

src/betterproto/plugin/models.py
```diff
@@ -360,7 +400,7 @@ def betterproto_field_args(self) -> List[str]:
     def field_wraps(self) -> Optional[str]:
         """Returns betterproto wrapped field type or None."""
         match_wrapper = re.match(
-            r"\.google\.protobuf\.(.+)Value", self.proto_obj.type_name
+            r"\.google\.protobuf\.(.+)Value$", self.proto_obj.type_name
```
nat-n, line 403: this was causing chaos by matching `EnumValueDescriptorProto` as if it were `EnumValue`.

src/betterproto/plugin/models.py
```diff
@@ -44,81 +69,82 @@
...
+            "betterproto"
+        ],
+        "group",
+        "oneof_index",
+    )
+    object.__setattr__(
+        Field.__dataclass_fields__["oneof_index"].metadata["betterproto"],
+        "group",
+        "oneof_index",
+    )
```
cetanu, line 137: Can you give me a rundown on why it needs to be done like this exactly? (I am trying to understand)

I looked up the help for this method

```
>>> help(object.__setattr__)
Help on wrapper_descriptor:

__setattr__(self, name, value, /)
    Implement setattr(self, name, value).
```

So my patchy current understanding is that the attribute `group` is being set to the string `"oneof_index"`

But I have no idea what is being passed to `self`

cetanu:
I've gone through, this all looks fine to me except for the one bit that I left a comment on, but even then it's simply me not understanding how it works.

If you want me to go through this more thoroughly I could checkout this branch and play with it for a bit, but if not I'd be keen to get this out the door and uncover in the beta release and uncover any small patches that need to be made.

cetanu (approved):

src/betterproto/plugin/models.py
Gobot1234: Because it's being passed as the instance to the unbound method `__setattr__`

cetanu:
Fixed the merge conflicts

src/betterproto/plugin/models.py
nat-n: @cetanu My attention was elsewhere so I only just noticed that you'd looked at this. thanks!

The [protoc messages](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L62) are proto2 (and we only support proto3), so the [FieldDescriptorProto](https://github.com/danielgtaylor/python-betterproto/blob/master/src/betterproto/lib/google/protobuf/__init__.py#L468) and Field message types rely on proto2 `optional` semantics to know whether a these specific fields have been set.

Although proto3 doesn't support knowing whether a field has been set in this way, the binary for a proto2 optional field is compatible with a proto3 one_of with a single possible type!

Therefore we can replicate the intended proto2 semantics here with a betterproto message type if we force it to treat this field as if it were in a one_of group of its own, by setting the "group" value to something locally unique (like `"oneof_index"`).

This hack is nasty enough to call into question whether this whole endeavour of making betterproto use it's own message types internally is worthwhile given that it's aimed at proto3. However I think it's not that big a deal given that the blast radius is quite well contained, to the plugin process which I think is quite well exercised by the unit tests.

nat-n:
@cetanu I rebased again and got CI tests to pass. I think it would be awesome to get this merged. There a a few other smaller quality improvement PRs waiting behind it. In my mind this is a minor milestone for betterproto's maturity, to not depend on protobuf to compile.

It's probably worth checking out and doing whatever is necessary satisfy yourself that it's good. I appreciate you taking the time.

src/betterproto/plugin/models.py
cetanu: The former is the part I didn't understand but now I do, the latter is interesting.
If we have some tests (doesn't necessarily have to be in this PR) that cover the proto2 and proto3 behavior that we want to support, then I see no issue with making this type of hack.

I'm okay to merge this

cetanu (approved):

Issue #19, opened 2020-02-10
Generate Servicer?


jyggen:
Currently it seems like this package is intended for gRPC _clients_ written in Python, meaning that people who also write their gRPC _servers_ in Python will have to generate their Python bindings the old fashioned way as well as better protos for their clients. Is it on the roadmap to generate a servicer as well? Would be nice to be able to use this package on both sides.

zidik:
I was just about to write the same question. I was looking for the ````GreeterBase```` class to extend like in this server implementation.
https://grpclib.readthedocs.io/en/latest/#server

nat-n:
I've had success using betterproto with grpcio. It required a bit of boiler plate, and manually coding the package name which betterproto doesn't include in the generated data classes,  but that can be hidden away with a bit of meta-programming.

githorse:
@nat-n Could you elaborate a bit on what you mean by "using betterproto with grpcio"? If I'm building a Python servicer, there's no getting around the fact that I have to compile all my protos twice (once with standard tools for server side, and once with betterproto for client side), right? Or is there some way I can return betterproto-compiled messages from my standard servicer? E.g.

```
import my_standard_grpc_module as standard
import my_betterproto_module as better

class EchoServicer(standard.EchoServicer):
    def Echo(self, request: better.EchoRequest, context) -> better.EchoResponse:
        return better.EchoResponse(message="boo")
```

(The above approach fails with `"Failed to serialize response!"` error.)

nat-n:
@githorse IMHO the idea of generating case classes for servicers doesn't really translate well into python where the type system doesn't naturally play the same role as in Java & co.

The point is that since python lends itself better to meta-programming and it all works with duck typing there's nothing to stop you from writing your own generic glue code to shove betterproto types into the grpcio service API (as I did).

Of course it would be useful if betterproto provided this glue code, and I'd be happy to clean up and contribute a PR for my own implementation if I had confidence the PR would get accepted.

I've created [a gist demonstrating the basics](https://gist.github.com/nat-n/e90097ebfb861cbb25e20b68bec0e39c) in a way that you can probably copy easily enough. Let me know of it works for you.

w4rum:
[...]

PR #170, opened 2020-11-06
service implementation stub generation

w4rum:
This implements #19. An example compiler output can be found [here](https://gist.github.com/w4rum/1f749149cafeec11c863bd625e62a5e9).

When compiling protobuf definitions, a service implementation stub is generated along the existing client stub:
```python
class ExampleServiceImplementation(betterproto.ServiceImplementation):
    async def example_unary_unary(
        self, example_string: str, example_integer: int
    ) -> "ExampleResponse":
        raise grpclib.GRPCError(grpclib.const.Status.UNIMPLEMENTED)

    async def __rpc_example_unary_unary(self, stream):
        request = await stream.recv_message()

        request_kwargs = {
            "example_string": request.example_string,
            "example_integer": request.example_integer,
        }

        await self._call_rpc_handler_server_unary(
            self.example_unary_unary,
            stream,
            request_kwargs,
        )

    def __mapping__(self) -> Dict[str, grpclib.const.Handler]:
        return {
            "/example_service.ExampleService/ExampleUnaryUnary": grpclib.const.Handler(
                self.__rpc_example_unary_unary,
                grpclib.const.Cardinality.UNARY_UNARY,
                ExampleRequest,
                ExampleResponse,
            ),
        }

```
The user is then supposed to create their own class that inherits from and overrides the methods of this generated implementation stub:
```python
class ExampleService(ExampleServiceImplementation):
    async def example_unary_unary(
        self, example_string: str, example_integer: int
    ) -> "ExampleResponse":
        return ExampleResponse(
            example_string=example_string,
            example_integer=example_integer,
        )
```
Note that the generated implementation stub includes a wrapper for each defined service (`__rpc_example_unary_unary`). This wrapper and the utility functions in the new abstract base class `ServiceImplementation` take care of translating `grpclib.server.Stream`'s `recv_message` and `send_message` from and to `yield` and `return` statements and parameters. I tried to match this to how the generated client stub is used.

Finally, the generated implementation stub can be used as a `grpclib` service handler because of the `__mapping__` function. An example of this is included as a new test case (`example_service`). I'm not sure if it's good practice to bind ports during tests but I didn't know how else to link the server and client. I also think it's not ideal to test the betterproto server against the betterproto client. Maybe it makes sense to include gRPC in `output_reference` so that we can create a reference client.

EDIT: Updated gist and snippets to reflect changes made in 207233b.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -25,7 +25,6 @@ from typing import (
 from ._types import T
 from .casing import camel_case, safe_snake_case, snake_case
 from .grpc.grpclib_client import ServiceStub
-from .grpc.grpclib_server import ServiceImplementation
 
 if sys.version_info[:2] < (3, 7):
     # Apply backport of datetime.fromisoformat from 3.7
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -25,6 +25,7 @@ from typing import (
 from ._types import T
 from .casing import camel_case, safe_snake_case, snake_case
 from .grpc.grpclib_client import ServiceStub
+from .grpc.grpclib_server import ServiceImplementation
 
 if sys.version_info[:2] < (3, 7):
     # Apply backport of datetime.fromisoformat from 3.7
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -1,4 +1,6 @@
 from abc import ABC
+from collections import AsyncIterable
+from typing import Callable, Any, Dict
 
 import grpclib
 import grpclib.server
@@ -9,21 +11,30 @@ class ServiceImplementation(ABC):
     Base class for async gRPC servers.
     """
 
-    __service_name__: str
+    async def __call_rpc_handler_server_unary(
+        self,
+        handler: Callable,
+        stream: grpclib.server.Stream,
+        request_kwargs: Dict[str, Any],
+    ) -> None:
 
-    def __rpc_methods__(self):
-        pass
+        response = await handler(**request_kwargs)
+        await stream.send_message(response)
 
-    def __mapping__(self):
-        mapping = {}
-        for (
-            method,
-            proto_name,
-            cardinality,
-            request_type,
-            response_type,
-        ) in self.__rpc_methods__():
-            mapping[f"/{self.__service_name__}/{proto_name}"] = grpclib.const.Handler(
-                method, cardinality, request_type, response_type
-            )
-        return mapping
+    async def __call_rpc_handler_server_stream(
+        self,
+        handler: Callable,
+        stream: grpclib.server.Stream,
+        request_kwargs: Dict[str, Any],
+    ) -> None:
+
+        response_iter = handler(**request_kwargs)
+        # check if response is actually an AsyncIterator
+        # this might be false if the method just returns without
+        # yielding at least once
+        # in that case, we just interpret it as an empty iterator
+        if isinstance(response_iter, AsyncIterable):
+            async for response_message in response_iter:
+                await stream.send_message(response_message)
+        else:
+            response_iter.close()
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -6,7 +6,7 @@ import grpclib
 import grpclib.server
 
 
-class ServiceImplementation(ABC):
+class ServiceBase(ABC):
     """
     Base class for async gRPC servers.
     """
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -11,16 +11,6 @@ class ServiceBase(ABC):
     Base class for async gRPC servers.
     """
 
-    async def _call_rpc_handler_server_unary(
-        self,
-        handler: Callable,
-        stream: grpclib.server.Stream,
-        request_kwargs: Dict[str, Any],
-    ) -> None:
-
-        response = await handler(**request_kwargs)
-        await stream.send_message(response)
-
     async def _call_rpc_handler_server_stream(
         self,
         handler: Callable,
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -1,0 +1,29 @@
+from abc import ABC
+
+import grpclib
+import grpclib.server
+
+
+class ServiceImplementation(ABC):
+    """
+    Base class for async gRPC servers.
+    """
+
+    __service_name__: str
+
+    def __rpc_methods__(self):
+        pass
+
+    def __mapping__(self):
+        mapping = {}
+        for (
+            method,
+            proto_name,
+            cardinality,
+            request_type,
+            response_type,
+        ) in self.__rpc_methods__():
+            mapping[f"/{self.__service_name__}/{proto_name}"] = grpclib.const.Handler(
+                method, cardinality, request_type, response_type
+            )
+        return mapping
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -11,7 +11,7 @@ class ServiceImplementation(ABC):
     Base class for async gRPC servers.
     """
 
-    async def __call_rpc_handler_server_unary(
+    async def _call_rpc_handler_server_unary(
         self,
         handler: Callable,
         stream: grpclib.server.Stream,
@@ -21,7 +21,7 @@ class ServiceImplementation(ABC):
         response = await handler(**request_kwargs)
         await stream.send_message(response)
 
-    async def __call_rpc_handler_server_stream(
+    async def _call_rpc_handler_server_stream(
         self,
         handler: Callable,
         stream: grpclib.server.Stream,
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -565,10 +565,6 @@ class ServiceCompiler(ProtoContentBase):
         return self.proto_obj.name
 
     @property
-    def full_proto_name(self) -> str:
-        return f"{self.parent.package_proto_obj.package}.{self.proto_obj.name}"
-
-    @property
     def py_name(self) -> str:
         return pythonize_class_name(self.proto_name)
 
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -280,10 +280,6 @@ class MessageCompiler(ProtoContentBase):
         return pythonize_class_name(self.proto_name)
 
     @property
-    def py_name_as_field(self) -> str:
-        return pythonize_field_name(self.proto_name)
-
-    @property
     def annotation(self) -> str:
         if self.repeated:
             return f"List[{self.py_name}]"
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -280,6 +280,10 @@ class MessageCompiler(ProtoContentBase):
         return pythonize_class_name(self.proto_name)
 
     @property
+    def py_name_as_field(self) -> str:
+        return pythonize_field_name(self.proto_name)
+
+    @property
     def annotation(self) -> str:
         if self.repeated:
             return f"List[{self.py_name}]"
@@ -558,6 +562,10 @@ class ServiceCompiler(ProtoContentBase):
     @property
     def proto_name(self) -> str:
         return self.proto_obj.name
+
+    @property
+    def full_proto_name(self) -> str:
+        return f"{self.parent.package_proto_obj.package}.{self.proto_obj.name}"
 
     @property
     def py_name(self) -> str:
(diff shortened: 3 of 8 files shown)


w4rum:
Whoops. I seem to have misunderstood the test framework. My test doesn't work. I'll get on fixing that.

w4rum:
Hmm, I don't know how my test could be incorporated into the existing test framework. My test doesn't have a message or service called `Test` and renaming my service to `Test` doesn't help because the generated class is called `TestImplementation`. I could write a custom test but I don't see how I could still have the test framework compile my protobuf sources then.

src/betterproto/grpc/grpclib_server.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Base class for async gRPC servers.
+    """
+
+    __service_name__: str
+
+    def __rpc_methods__(self):
+        pass
+
+    def __mapping__(self):
```
cetanu, line 17: are return type annotations possible for this method and the above?

cetanu:
For me, this looks good, but I will wait for others to weigh in.
I think opening a socket during tests is not the worst thing ever, but there could be a proper way to do it.

src/betterproto/grpc/grpclib_server.py
w4rum: I removed `__rpc_methods__` in my latest commit. `__mapping__` now has typing hints.

w4rum:
I removed `__service_name__` and `__rpc_methods__`. I realized that that was an unnecessary indirection. The `__mapping__` method necessary to register the implementation stub as a grpclib handler is now part of the generated class.

I also cut down on some of the duplicated code in the generated class. The code for handling the grpclib is now part of the ABC.

It appears that the CI is broken. Correct me if I'm wrong.

tests/inputs/example_service/test_example_service.py
```diff
@@ -0,0 +1,108 @@
...
+
+from tests.output_betterproto.example_service.example_service import (
+    ExampleServiceImplementation,
+    ExampleServiceStub,
+    ExampleRequest,
+    ExampleResponse,
+)
+
+
+class ExampleService(ExampleServiceImplementation):
```
nat-n, line 15: This seems backwards to me. I'd usually use _Implementation_ to refer to the class that implements the abstract class for the service. 

How about: 

```suggestion
class ExampleServiceImplementation(ExampleServiceBase):
```

tests/inputs/example_service/test_example_service.py
```diff
@@ -0,0 +1,108 @@
...
+    assert response.example_string == test_string
+    assert response.example_integer == test_int
+
+    # stream stream
+    async for response in stub.example_stream_stream(request_iterator()):
+        assert response.example_string == test_string
+        assert response.example_integer == test_int
+
+
+def test_server_start():
```
nat-n, line 106: I don't think it should be necessary to run your own event loop. Did you try use `@pytest.mark.asyncio`? 

example: https://github.com/danielgtaylor/python-betterproto/blob/master/tests/grpc/test_grpclib_client.py#L174

tests/inputs/example_service/test_example_service.py
```diff
@@ -0,0 +1,108 @@
...
+async def async_test_server_start():
+    host = "127.0.0.1"
+    port = 13337
+
+    test_string = "test string"
+    test_int = 42
+
+    # start server
+    server = Server([ExampleService()])
+    await server.start(host, port)
```
nat-n, line 63: Do you try use the testing facility of grpclib?
e.g. https://github.com/danielgtaylor/python-betterproto/blob/master/tests/grpc/test_grpclib_client.py#L63
I think this is generally preferable to opening a socket etc.

nat-n (changes requested):
Thanks for picking this up, it'll be good to have a solution.

I still need to properly look at the code generation part when I get some time, but I found a points worth looking at for now.

src/betterproto/__init__.py
```diff
@@ -25,6 +25,7 @@
 from ._types import T
 from .casing import camel_case, safe_snake_case, snake_case
 from .grpc.grpclib_client import ServiceStub
+from .grpc.grpclib_server import ServiceImplementation
```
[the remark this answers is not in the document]
w4rum: Sounds good. I moved the import to `betterproto/grpc/__init__.py`

tests/inputs/example_service/test_example_service.py
w4rum: I initially did this to clearly distinguish between the client-facing `ServiceStub` and the server-facing `ServiceImplementation`. But I see how this can be confusing. I renamed the ABC to `ServiceBase` and the generated class to `ExampleServiceBase`. I'd like to avoid putting "Implementation" in the example implementation name as I feel it's unnecessarily long.

How about:
```python
class ExampleService(ExampleServiceBase):
```

tests/inputs/example_service/test_example_service.py
w4rum: I didn't know that was a thing. Thanks for the pointer!

tests/inputs/example_service/test_example_service.py
w4rum: Also didn't know that was a thing. This makes the code a lot cleaner.

w4rum:
On a side note: Why does the CI suddenly work again? This branch does not have changes from #172. Is it using the action config from `master` because the PR is targeting `master`? If so, why did the initial CI run on #172 succeed even though it was targeting `master` with the old action config?

src/betterproto/__init__.py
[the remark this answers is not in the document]
nat-n: Actually lets not do that either. The idea is to be able to generate grpcio based client/server as well (according to compile time options), so best to keep them isolated for the same reason. It doesn't matter if the resulting import paths are long since. it'll primarily be used from generated code.

nat-n:
> On a side note: Why does the CI suddenly work again? This branch does not have changes from #172. Is it using the action config from `master` because the PR is targeting `master`? If so, why did the initial CI run on #172 succeed even though it was targeting `master` with the old action config?

I'm a surprised as you 🤷

src/betterproto/__init__.py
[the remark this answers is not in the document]
w4rum: Ah, so you want to allow the user to generate *just* the client or *just* the server? That makes sense. I'll remove the import entirely and just use the long path in the generated file.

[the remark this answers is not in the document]
w4rum: I don't see how I can make the generated class work without adding an extra import now:
```python
from betterproto.grpc.grpclib_server import ServiceBase
```
since just using
```python
class ExampleServiceBase(betterproto.grpc.grpclib_server.ServiceBase):
```
does not work. Is it okay for me to add the above import to the generated class? I figure the template can later be adjusted so that this import can be disabled via compiler options.

src/betterproto/templates/template.py.j2
```diff
@@ -154,6 +155,92 @@ class {{ service.py_name }}Stub(betterproto.ServiceStub):
...
+                                            Optional[{{ field.annotation }}]
+                                         {%- else -%}
+                                            {{ field.annotation }}
+                                         {%- endif -%}
+                    {%- if not loop.last %}, {% endif -%}
+                {%- endfor -%}
+            {%- endif -%}
+        {%- else -%}
+            {# Client streaming: need a request iterator instead #}
+            , {{ method.py_input_message.py_name_as_field }}_iterator: AsyncIterable["{{ method.py_input_message_type }}"]
```
nat-n, line 180: what about just calling the argument request_iterator? I figure consistency is worth something here and the type annotation is already more descriptive.

nat-n: Also is the type not more specifically an AsyncIterator?

src/betterproto/templates/template.py.j2
```diff
@@ -154,6 +155,92 @@ class {{ service.py_name }}Stub(betterproto.ServiceStub):
...
+        request = await stream.recv_message()
+
+        request_kwargs = {
+        {% for field in method.py_input_message.fields %}
+            "{{ field.py_name }}": request.{{ field.py_name }},
+        {% endfor %}
+        }
+
+        {% else %}
+        request_kwargs = {"{{ method.py_input_message.py_name_as_field }}_iterator": stream.__aiter__()}
```
nat-n, line 203: why call `__aiter__` here instead of just passing the stream object?

src/betterproto/grpc/grpclib_server.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    async def _call_rpc_handler_server_unary(
+        self,
+        handler: Callable,
+        stream: grpclib.server.Stream,
+        request_kwargs: Dict[str, Any],
+    ) -> None:
+
+        response = await handler(**request_kwargs)
```
nat-n, line 21: this logic seems simple enough that you might as well just inline it in the generate code?

src/betterproto/templates/template.py.j2
w4rum: > what about just calling the argument request_stream?
It's called `request_iterator` on the client-facing side. I suggest that we name it that if we want more consistency.

w4rum: > Also is the type not more specifically an AsyncIterator?

You're right, I'm specifically calling `stream.__aiter__`. I'll change this.

tests/inputs/example_service/example_service.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package example_service;
+
+service ExampleService {
```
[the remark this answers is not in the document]
w4rum: Changed it. Now that I think about it, `example_service` might not be too good a test case name since it's supposed to be an actual test instead of just something for the README, like the `example` case.

What do you think about integrating this test into the existing `service` test case? It seems pretty similar.

src/betterproto/grpc/grpclib_server.py
w4rum: I did this for consistency. But yeah, I suppose 5 lines of code + an abstract method is a bit excessive for something that's a one-liner.

w4rum:
I'm not sure what the issue is with the test framework. Specifically, I don't see why `example_service` fails but `service` doesn't. This error occurred after I removed the `Test` message and named the service `Test`.

nat-n:
> I'm not sure what the issue with the test framework is. Specifically, I don't see why `example_service` fails but `service` doesn't. This error occurred after I removed the `Test` message and named the service `Test`.

I think maybe you need to register the service in `tests/inputs/config.py`

w4rum:
Thank you, that seems to have fixed it.

tests/inputs/example_service/test_example_service.py
```diff
@@ -0,0 +1,95 @@
...
+        response = ExampleResponse(
+            example_string=example_string,
+            example_integer=example_integer,
+        )
+        yield response
+        yield response
+        yield response
+
+    async def example_stream_unary(
+        self, example_request_iterator: AsyncIterable["ExampleRequest"]
```
nat-n, line 35: Is this argument name still correct, I suspect mypy would complain.

nat-n (changes requested):
Almost good to go, I just spotted a couple more minor issue.

tests/inputs/example_service/test_example_service.py
w4rum: Fixed.

I don't know if mypy complains about this specifically, since mypy highlights these entries with
```
syntax error in type comment
```
for me anyway. Since this is also true for the client-facing methods that I didn't change, I just assumed that my mypy config is somewhat broken and paid it no further attention.

src/betterproto/plugin/models.py
```diff
@@ -553,12 +557,17 @@ class ServiceCompiler(ProtoContentBase):
...
+        self.output_file.typing_imports.add("Dict")
         super().__post_init__()  # check for unset fields
 
     @property
     def proto_name(self) -> str:
         return self.proto_obj.name
 
+    @property
+    def full_proto_name(self) -> str:
+        return f"{self.parent.package_proto_obj.package}.{self.proto_obj.name}"
```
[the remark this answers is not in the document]
w4rum: That was actually a leftover. I had switched to using the already-existing `method.route` in a previous commit. I have now deleted this function.

tests/inputs/example_service/test_example_service.py
nat-n: Yea, there's a bit of work to get the whole project properly typed, but every little helps.

src/betterproto/plugin/models.py
```diff
@@ -279,6 +279,10 @@ def proto_name(self) -> str:
     def py_name(self) -> str:
         return pythonize_class_name(self.proto_name)
 
+    @property
+    def py_name_as_field(self) -> str:
```
nat-n, line 283: looks like this isn't used either actually?

w4rum: Not anymore since 5bbe19ab. Nice catch. Good thing you're looking over this. I completely forgot.

nat-n (approved):
Ship it!

PR #172, opened 2020-11-21
Fix CI: replace now-disabled set-env command

w4rum:
[GitHub recently disabled the `set-env` command.](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) This is a fix sent to me by @abn that resolves the issue.


nat-n:
Thanks for taking the initiative on this :+1:

nat-n (approved):

PR #173, opened 2020-11-24
Update dependencies and add ci checks for python 3.9

abn:
Updating locked dependencies so that python 3.9 development environments works (eg: grpcio compile issue). This change set also adds python 3.9 as a github action runtime for tests.


nat-n (approved):
thanks :)

PR #176, opened 2020-11-24
Properly serialize zero-value messages in a oneof group

nat-n:
This PR is a continuation/replacement for #129, just with additional tests.


tests/util.py
```diff
@@ -47,15 +47,27 @@ async def protoc(
     return stdout, stderr, proc.returncode
 
 
-def get_test_case_json_data(test_case_name: str, json_file_name: Optional[str] = None):
-    test_data_file_name = json_file_name or f"{test_case_name}.json"
-    test_data_file_path = inputs_path.joinpath(test_case_name, test_data_file_name)
+def get_test_case_json_data(test_case_name: str, *json_file_names: str) -> List[str]:
```
nat-n, line 50: this change makes it easier to manage multiple sample message json files in the test inputs.

src/betterproto/__init__.py
```diff
@@ -704,7 +704,7 @@ def __bytes__(self) -> bytes:
                     meta.number,
                     meta.proto_type,
                     value,
-                    serialize_empty=serialize_empty,
+                    serialize_empty=serialize_empty or selected_in_group,
```
nat-n, line 707: this is the fix.

pyproject.toml
```diff
@@ -49,10 +49,20 @@ generate    = { script = "tests.generate:main", help = "Generate test cases (do
 test        = { cmd = "pytest --cov src", help = "Run tests" }
 types       = { cmd = "mypy src --ignore-missing-imports", help = "Check types with mypy" }
 format      = { cmd = "black . --exclude tests/output_",   help = "Apply black formatting to source code" }
-clean       = { cmd = "rm -rf .coverage .mypy_cache .pytest_cache dist betterproto.egg-info **/__pycache__ tests/output_*", help = "Clean out generated files from the workspace" }
 docs        = { cmd = "sphinx-build docs docs/build", help = "Build the sphinx docs"}
 bench       = { shell = "asv run master^! && asv run HEAD^! && asv compare master HEAD", help = "Benchmark current commit vs. master branch"}
 
+clean.cmd = """
+rm -rf .asv
```
nat-n, line 56: added .asv to the list of files & directories to clean up in the workspace.

SoerenBusse:
Hey there,

any update on having this fix merged? I've just ran into the same issue and where debugging the issue until I find your pull-request. I came up with the same idea of adding `selected_in_group` to the `serialize_empty` parameter.

I would really appreciate a merge.

pyproject.toml
SoerenBusse: This change doesn't seem to be valid toml syntax. When running pip install directly from this branch we're getting an `pytoml.core.TomlError: /tmp/pip-req-build-ebkn5z74/pyproject.toml(55, 1):` error.

nat-n: It is valid toml syntax. But pytoml doesn't know that because it's buggy and unmaintained :( 

I guess you're using an older version of pip that depends on it? Let me know if that's not the case.

SoerenBusse: Yes, you're right. 20.0.2 is broken while 20.3.3 is working fine. However I think it would be good to have a bit of backwards compatibility, especially because 20.0.2 is currently shipped with Ubuntu 20.04, so I think multiple users will stumple into this issue. It might also be more consistent when using the same (pytoml compatible) syntax for the clean command, which is also used for docs, bench ...

cetanu (approved):

PR #177, opened 2020-11-25
Fix incorrect routes in generated client when service is not in a package

### LINKED ISSUE #171: Mismatch of method route names with grpclib server
Hey there,

we're trying to use the "Betterproto" library with a grpc server based on grpclib. However the generated routes in the stub file of betterproto doesn't match the routes in the service from grpclib, which results in a very vacuous error message, so the method call will fail:

Betterproto:
```
    "/.Frontend/SampleMethod", request, SampleMethodReply
)
```

grpclib:
```
return {
            '/Frontend/SampleMethod': grpclib.const.Handler(
                ...
            )
}
```

Why does betterproto generate different route-names than grpclib? Is there a specification for naming the route-names?

I think the error is comming from here:
https://github.com/danielgtaylor/python-betterproto/blob/master/src/betterproto/plugin/models.py#L641
Is there a check missing which replaces a dot with an empty string?

We're generating the proto files using the following command:
`python3 -m grpc_tools.protoc -I grpc-interface-definitions/proto/ --python_betterproto_out=lib/grpc-interface_definitions grpc-interface-definitions/proto/*`

<hr>

The vacuous error message we got from betterproto is the following one:
```
GRPCError(<Status.UNKNOWN: 2>, 'Missing content-type header', None)
```
Debugging the server shows that the request is canceled due to a not found method name. It sends a `Status.UNIMPLEMENTED` error back to the client, which seems to have an issue parsing these early errors. I think this error message might be related to a bug in the underlying grpclib and isn't directly associated to betterproto. However I will mention it here if anybody is searching for this error message like me a few hours ago ;)

nat-n:
Fixes #171

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -638,7 +638,10 @@ class ServiceMethodCompiler(ProtoContentBase):
 
     @property
     def route(self) -> str:
-        return f"/{self.output_file.package}.{self.parent.proto_name}/{self.proto_name}"
+        package_part = (
+            f"{self.output_file.package}." if self.output_file.package else ""
+        )
+        return f"/{package_part}{self.parent.proto_name}/{self.proto_name}"
 
     @property
     def py_input_message(self) -> Optional[MessageCompiler]:


cetanu (approved):

src/betterproto/plugin/models.py
```diff
@@ -638,7 +638,10 @@ def proto_name(self) -> str:
 
     @property
     def route(self) -> str:
-        return f"/{self.output_file.package}.{self.parent.proto_name}/{self.proto_name}"
+        package_part = (
+            f"{self.output_file.package}." if self.output_file.package else ""
```
cetanu, line 642: I would personally use

```python
package_part = '.'.join([self.output_file.package, self.parent.proto_name])
return f"/{package_part}/{self.proto_name}"
```

But maybe the intention of yours is more clear, since my example almost uses a side effect to achieve the result

Issue #179, opened 2020-11-30
An empty message in a repeated field will be serialized to empty bytes


livelxw:
Given
```protobuf
syntax = "proto3";

message MessageA {
  repeated float field_a = 1;
}

message MessageB {
  repeated MessageA field_b = 1;
}
```
When 
```python3
s = MessageB(field_b=[MessageA()]).SerializeToString()
```
Then `s` will be `b''` and if call `MessageB.FromString(s)` the result will be `MessageB()` (different from `MessageB(field_b=[MessageA()])`).
But in some other protobuf libraries like [Scalapb](https://github.com/scalapb/ScalaPB) the serialized string will be `\n\000`. If call `MessageB.FromString(b'\n\000')` the result will be `MessageB(field_b=[MessageA()])`(the expected result).
Is this a bug?

nat-n:
It looks like we differ from the google implementation in this respect, so yes I'd say it is a bug. Thanks for reporting.

Gobot1234:
Fixed in #180

PR #180, opened 2020-11-30
Fix serialization of repeated fields with empty messages

nat-n:
This addresses #179 

There's a failing test because the sample JSON for the binary compatibility test does not re-serialise to the same json (because it has deeply nested empty values). I need to amend the testing utils to allow for cases like this but would rather wait for #176  to be merged first to avoid conflicting refactors.


mydatamodels:
Hello there,
Any news?
I'm sorry to chase you but it's a pretty serious bug. What can we do to help you?
Regards.

cetanu:
Oops, I pressed ready for review not knowing what it does. If this is still WIP feel free to revert that.

nat-n:
Rebased and finished refactoring some testing logic to allow json examples to be used for testing binary compatibility but skipped in the test for to_dict/from_dict symmetry.

cetanu (approved):

cetanu (approved):

PR #187, opened 2020-12-18
Remove enum prefixes

### LINKED ISSUE #174: Enum Prefix is not stripped
According to the [Style Guide](https://developers.google.com/protocol-buffers/docs/style#enums), protobuf enums should have the enum name as a prefix for all enum values, to avoid naming conflicts in languages like C, where all values from all enums are in the global namespace. Protobuf compilers for languages with scoped enums like C# will [strip these prefixes](https://developers.google.com/protocol-buffers/docs/reference/csharp-generated#enum) if they match the enum name. Since enums in betterproto are scoped as well, these prefixes should be stripped.

Example:
```
enum ArithmeticOperator {
    ARITHMETIC_OPERATOR_NONE = 0;
    ARITHMETIC_OPERATOR_PLUS = 1;
    ARITHMETIC_OPERATOR_MINUS = 2;
}
```
should become
```
class ArithmeticOperator(betterproto.Enum):
    NONE = 0
    PLUS = 1
    MINUS = 2
```
so that we can use it as 
```
if something.operator == ArithmeticOperator.PLUS:
     do_addition()
```
instead of
```
if something.operator == ArithmeticOperator.ARITHMETIC_OPERATOR_PLUS:
    do_addition()
```

### LINKED ISSUE #302: Enum with reserved word (`None`) clashes with python
Hiya,

I've just started working with gather.town which provides a version 3 proto file.  I had to remove the optional keywords (as best I can tell, they're somewhat unnecessary in proto3 files) and it all started generating fine.  I found, however, that when importing the file that the following enum:

```
message EmoteEnum {
  enum ENUM {
    None = 0;
    Wave = 1;
    Heart = 2;
    PartyPopper = 3;
    ThumbsUp = 4;
    QuestionMark = 5;
    RaisedHand = 6;
  }
}
```

generates the following dataclass:

```
class EmoteEnumENUM(betterproto.Enum):
    None = 0
    Wave = 1
    Heart = 2
    PartyPopper = 3
    ThumbsUp = 4
    QuestionMark = 5
    RaisedHand = 6
```

which tries to assign to None.  There doesn't appear to be a great mechanism for working around this automatically.  The best I've found would be to append an `_` and documenting that this would happen to any field name found in [keyword.kwlist](https://docs.python.org/3/library/keyword.html#keyword.kwlist).  There is a [workaround](https://stackoverflow.com/questions/60074344/reserved-word-as-an-attribute-name-in-a-dataclass-when-parsing-a-json-object) for `dataclasses-json`, but obviously that's not  a default package with python and doesn't look like a suitable solution to the issue...

Gobot1234:
## Summary

Removes any enum class name prefixes from enum members.

NONE -> NONE
ARITHMETIC_OPERATOR_NONE -> NONE

## Checklist

- [X] This PR fixes an issue.
  - [x] This change has an associated test
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed)

Closes #174 
Closes #302


Gobot1234:
I might actually just update this to remove any prefix on an enum that is the same for all the members.

src/betterproto/compile/naming.py
```diff
@@ -11,3 +13,19 @@ def pythonize_field_name(name: str) -> str:
 
 def pythonize_method_name(name: str) -> str:
     return casing.safe_snake_case(name)
+
+
+def pythonize_enum_member_names(names: Iterable[str]) -> List[str]:
+    """Removes any of the same characters from an Enum member's names."""
```
nat-n, line 19: I'm not sure this is the quite right thing to do. Surely what we really want is to specifically check if all the enum names are prefixed with exactly the capitalised snake case name of the enum (as per the convention), and if so then remove it.

Removing ANY common prefix could be quite surprising... especially if the enum only has one value ;)

nat-n (changes requested):

src/betterproto/compile/naming.py
Gobot1234: I'll revert this then

src/betterproto/compile/naming.py
```diff
@@ -11,3 +13,11 @@ def pythonize_field_name(name: str) -> str:
 
 def pythonize_method_name(name: str) -> str:
     return casing.safe_snake_case(name)
+
+
+def pythonize_enum_member_name(name: str, enum_name: str) -> str:
+    enum_name = casing.snake_case(enum_name).upper()
+    find = name.find(enum_name)
```
nat-n, line 20: I don't think removing the enum name from just anywhere in the key name would be a good idea, especially if each key is evaluated independently.

```suggestion
    prefix = casing.snake_case(enum_name).upper() + "_"
    if name.startswith(prefix):
        return name[len(prefix):]
    return name
```

I wonder if it's better to be all or nothing here... i.e. only do the renaming if the convention is perfectly followed (by all keys). Do you know of a precedent or documented rule to follow one way or the other?

tests/inputs/enum/test_enum.py
```diff
@@ -82,3 +79,8 @@ def enum_generator():
         yield Choice.THREE
 
     assert Test(choices=enum_generator()).to_dict()["choices"] == ["ONE", "THREE"]
+
+
+def test_renamed_enum_members():
+    for member in ArithmeticOperator:
```
nat-n, line 85:
```suggestion
    assert set(ArithmeticOperator) == set(("NONE", "PLUS", "MINUS")):
```

Gobot1234: That will fail due to sets sorting keys but I get what you're going for

src/betterproto/compile/naming.py
Gobot1234: I used find because sometimes people in C use k_ to prefix an enum.

cetanu: +1 for find, although a loosely held opinion

cetanu: Why would a prefix of `k_` necessitate `str.find`? Just curious

Do we just need a small test for this case (and any others)?

Gobot1234: Because the name doesn't start with k_

cetanu:
Needs to be relocked, but apart from that looks fine to me

cetanu (approved):

Gobot1234:
Failure looks unrelated

cetanu (approved):
I assume you will squash the commits for this 😅

src/betterproto/casing.py
```diff
@@ -136,4 +136,8 @@ def lowercase_first(value: str) -> str:
 
 def sanitize_name(value: str) -> str:
     # https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles
-    return f"{value}_" if keyword.iskeyword(value) else value
+    if keyword.iskeyword(value):
+        return f"{value}_"
+    if not value.isidentifier():
```
nat-n, line 141: What's the reasoning here? There are many ways that a string could fail to be a valid identifier, I'm not sure prefixing with an underscore is a treatment well matched to many of them. If your intention is more specific (like just to add a prefix to names that don't start with a letter?) then you should implement that logic.

src/betterproto/compile/naming.py
nat-n: I still think this implementation is unclear and error prone. For example imagine the chaos given an enum type with a single letter name! If the intention is to remove a "prefix", then just remove the prefix.

If you feel it's necessary to make it more clever than that (which I doubt) then a docstring explanation and test coverage are essential.

nat-n (changes requested):
Hi, just dropping by again. You can get someone else to approve and merge this, but if you want my opinion it needs a little more thought.

src/betterproto/casing.py
Gobot1234: The protobuf compiler only accepts ASCII alphabetical identifiers as field names, so the only inputs to this function that isn't a valid python identifier is an integer prefixed string and so to turn it into a valid python identifier you must prefix it with an underscore.

nat-n: Good to know the problem is more constrained that I thought it might be. Although if your intention here is to add a prefix to any string that starts with a digit, then wouldn't it be better (clearer, less surprising) to explicitly test for the first character being a digit?

Gobot1234: AFAIK the set of invalid identifiers has no intersection with _ + the invalid identifier, this way I just though was more forwards compatible

cetanu (approved):
I've looked over the files again, I am okay with this change. If some complication arises we can deal with it later.

src/betterproto/casing.py
cetanu: I think this is not quite a one-way door, and we can maybe add a comment or TODO to refactor later, so we can prioritize a 2.0 release. If nobody disagrees I will resolve this conversation

cetanu (approved):

snikch:
FYI this has broken the ability to use enums in JSON payloads cross platform as the JSON value has been stripped of the prefix too.

From the linked [C# docs](https://protobuf.dev/reference/csharp/csharp-generated/#enum) in #174:

> This name transformation does not affect the text used within the JSON representation of messages.

PR #192, opened 2020-12-30
Bug fix in casing.sanitize_name for names starting with an int

Gobot1234:


nat-n:
Nice catch. Would be good to have a test for this.

Gobot1234:
Only found this due to #187 and code using an int would be invalid
```proto
message Test {
    bool value = 1;
    string 23 = 2;
}
```

nat-n:
> Only found this due to #187 and code using an int would be invalid
> 
> ```protobuf
> message Test {
>     bool value = 1;
>     string 23 = 2;
> }
> ```

Do you mean that there isn't actually a real scenario that would trigger the bug caused here?

Gobot1234:
It

> > Only found this due to #187 and code using an int would be invalid
> > ```protobuf
> > message Test {
> >     bool value = 1;
> >     string 23 = 2;
> > }
> > ```
> 
> Do you mean that there isn't actually a real scenario that would trigger the bug caused here?

It could be if #187 was merged.

Gobot1234:
I'm gonna merge this with #187

PR #193, opened 2021-01-01
Fix compilation of fields named 'bytes'

SpencerSharkey:
If you have a bytes field named "bytes", a recursion error occurs when importing the compiled stubs.

### DIFF
--- a/src/betterproto/casing.py
+++ b/src/betterproto/casing.py
@@ -132,7 +132,15 @@ def lowercase_first(value: str) -> str:
     """
     return value[0:1].lower() + value[1:]
 
+def is_keyword(value: str) -> bool:
+    if keyword.iskeyword(value):
+        return True
+
+    if value == "bytes":
+        return True
+
+    return False
 
 def sanitize_name(value: str) -> str:
     # https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles
-    return f"{value}_" if keyword.iskeyword(value) else value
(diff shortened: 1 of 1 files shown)


Gobot1234:
Why is `bytes` a special case, should it not happen for all builtin types?

nat-n:
@SpencerSharkey thanks for doing this. 

The solution looks reasonable, however as @Gobot1234 it may be that you've scratched the surface of a larger issue here.

That said it's not immediately obvious to me exactly why this error is happening. I don't have a lot of time to look into this now, but if you could post an (abridged) stack trace of the recursion error then that would help :)

Also it would be good to have a regression test for this, would you be willing to add an example to the test suite following the pattern described here https://github.com/danielgtaylor/python-betterproto/tree/master/tests ?

Gobot1234:
I can't reproduce this
```proto
syntax = "proto3";

message Test {
    bytes bytes = 1;  // or string bytes = 1;
}
```both compile fine for me. Can you show an example where this doesn't work?

giantryansaul:
Commented on the related issue #210, I'm seeing a recursive issue from a definition we have that uses `bytes`. I tested out this PR locally and it fixed my runtime issue.

nat-n:
@SpencerSharkey Thanks again for your contributiong.

I'm closing in favour of #226 because I made some tweaks to extend the fix and enable a test that has already been written to cover it, but didn't have write access to push to the fork (there's a checkbox to allow this when creating the PR).

PR #196, opened 2021-01-16
Remove imports if possible

Gobot1234:
Does a bit of work over #170 to make generated code more consistent with betterproto.ServiceStub


nat-n:
I'm sorry to say so but I don't think this change is desirable, because it will make it harder (that is an even more breaking change) to allow grpclib to be an optional dependency.

The reason this is an issue is that currently most of the weight of betterproto (in terms of memory usage and import time) is due to the root package depending on grpclib. This is unfortunate given that many users don't benefit from this dependency.

Gobot1234:
Makes sense, but from what I can remember the import is put in by default, so I'll have to have update the template to fix this I would assume?

Issue #199, opened 2021-01-19
nested empty types not in to_dict


DK99:
I am not sure if I am wrong here but when executing this:

```python
from dataclasses import dataclass

import betterproto

@dataclass
class Inner(betterproto.Message):
    inner1: str = betterproto.string_field(1)
    inner2: str = betterproto.string_field(2)


@dataclass
class Test(betterproto.Message):
    name: str = betterproto.string_field(1)
    inner: "Inner" = betterproto.message_field(2)

test = Test(name='123', inner=Inner())
print(test.to_json())
```

I receive the output `{"name": "123"}`. I would expect the output to be`{"name": "123", inner={}}`.
Of course there is a similar issue with the `to_dict` function

Gobot1234:
Is this also the default behaviour for the google implementation?

DK99:
Yes it is here the example generated with the google implementation

<details>
<summary>Click to expand!</summary>

```python
# -*- coding: utf-8 -*-
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# source: test.proto
"""Generated protocol buffer code."""
from google.protobuf import descriptor as _descriptor
from google.protobuf import message as _message
from google.protobuf import reflection as _reflection
from google.protobuf import symbol_database as _symbol_database
# @@protoc_insertion_point(imports)

_sym_db = _symbol_database.Default()




DESCRIPTOR = _descriptor.FileDescriptor(
  name='test.proto',
  package='',
  syntax='proto3',
  serialized_options=None,
  create_key=_descriptor._internal_create_key,
  serialized_pb=b'\n\ntest.proto\"\'\n\x05Inner\x12\x0e\n\x06inner1\x18\x01 \x01(\t\x12\x0e\n\x06inner2\x18\x02 \x01(\t\"+\n\x04Test\x12\x0c\n\x04name\x18\x01 \x01(\t\x12\x15\n\x05inner\x18\x02 \x01(\x0b\x32\x06.Innerb\x06proto3'
)




_INNER = _descriptor.Descriptor(
  name='Inner',
  full_name='Inner',
  filename=None,
  file=DESCRIPTOR,
  containing_type=None,
  create_key=_descriptor._internal_create_key,
  fields=[
    _descriptor.FieldDescriptor(
      name='inner1', full_name='Inner.inner1', index=0,
      number=1, type=9, cpp_type=9, label=1,
      has_default_value=False, default_value=b"".decode('utf-8'),
[...]

PR #200, opened 2021-01-19
to_dict on empty nested types

DK99:
The issue is described [here #199](https://github.com/danielgtaylor/python-betterproto/issues/199).

It was a very simple fix. I don't think a test case is neccesary for this bug. If you think otherwise I can add one.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -972,6 +972,8 @@ class Message(ABC):
                     )
                 ):
                     output[cased_name] = value.to_dict(casing, include_default_values)
+                else:
+                    output[cased_name] = {}
             elif meta.proto_type == TYPE_MAP:
                 for k in value:
                     if hasattr(value[k], "to_dict"):


DK99:
I analyzed the code a bit more. The problem why so many test fail is that currently the library doesn't know the difference between an uninitialized field and a field which is initialized with an empty object.

src/betterproto/__init__.py
```diff
@@ -875,6 +874,10 @@ def parse(self: T, data: bytes) -> T:
                 )
 
             current = getattr(self, field_name)
```
Gobot1234, line 876: This should be moved below that if statement

src/betterproto/__init__.py
```diff
@@ -875,6 +874,10 @@ def parse(self: T, data: bytes) -> T:
...
+
+            if self.__raw_get(field_name) == PLACEHOLDER:
```
Gobot1234, line 878: This should use `is` not `==`

src/betterproto/__init__.py
```diff
@@ -972,6 +975,8 @@ def to_dict(
                     )
                 ):
                     output[cased_name] = value.to_dict(casing, include_default_values)
+                elif self.__raw_get(field_name) != PLACEHOLDER:
```
Gobot1234, line 978: Similar thing here, should be `is not`

src/betterproto/__init__.py
Gobot1234: And maybe there should be a return on this

src/betterproto/__init__.py
DK99: The line 879 uses the `current` attribute, so putting that below doesn't make any sense

src/betterproto/__init__.py
DK99: I was thinking about that but I saw line 540 doing the same comparison

src/betterproto/__init__.py
DK99: Why should there be a return here? The for loop has to continue

Gobot1234: Sorry I meant continue

tests/test_features.py
```diff
@@ -17,7 +17,7 @@ class Foo(betterproto.Message):
     assert betterproto.serialized_on_wire(foo.bar) is False
 
     # Serialized after setting something
-    foo.bar.baz = 1
+    foo.bar = Bar(baz=1)
```
Gobot1234, line 20: Changing all this is a regression if this PR breaks this

DK99: I know that isn't great. My PR would break lazy initialization. If you want to keep that, I have to think of another implementation. But lazy initialization also isn't in the google implementation, so I thought it wouldn't be important.

Gobot1234: Well breaking things isn't good. I know how this can be fixed it's been discussed before but it involves every field having a serialized attribute or similar.

nat-n:
Hi @DK99, thanks for working on this.

> I analyzed the code a bit more. The problem why so many test fail is that currently the library doesn't know the difference between an uninitialized field and a field which is initialized with an empty object.

I believe this is by design in order to be in line with proto3 semantics, whereby primitive type fields are interpreted as their zero value if unset, and vice versa.

More generally I think the fix for this bug should be isolated to the to_dict structuring logic and shouldn't have any implications for the parsing or internal representation of object. Unless I'm missing something?

> I know that isn't great. My PR would break lazy initialization. If you want to keep that, I have to think of another implementation. But lazy initialization also isn't in the google implementation, so I thought it wouldn't be important.

As I mentioned in slack channel, as I recall lazy initialisation is required for recursive message types to work, so I suspect the google implementation actually does support it.

As for testing, ideally we should have a test case for this in line with the pattern documented here: https://github.com/danielgtaylor/python-betterproto/tree/master/tests Let me know if this doesn't turn out to be trivial (the test setup needs some work which has been done in other pending PRs).

Issue #20, opened 2020-02-12
Option for generating code without async?


adisunw:
I have some infrastructure developed around the previously gen'd grpc code, however I would really like to adopt betterproto for all its extra pythonic features. 

The only hurdle i'm facing is that the async implementation is turned on by default. Is there a way to turn this off and not have the async gen code?

githorse:
Just to clarify beyond all doubt here -- for the moment the answer here is no, correct? We can't turn off async/await in the generated stub?

nat-n:
@githorse your assessment is correct.

hdutra:
This library does not work for Cloud Run due to the lack of this functionality. If it worked for sure betterproto would be the best option.

zbentley:
What about using native Python's async-to-sync functionality. That is, given an async RPC method called via `result = await ServiceStub.some_method()`, you could do either:

```python
result = asyncio.run(ServiceStub.some_method())
```

...if you wanted to start and stop the event loop just for that method. If that incurs too much overhead, you could call into an existing event loop via something like:

```python
result = asyncio.run_coroutine_threadsafe(ServiceStub.some_method(), loop=asyncio.get_event_loop()).result()
```

whs:
I tried that as a workaround. The problem was that

- grpclib channel constructor requires a running working loop. Constructor can't be marked as async so using those run in async functions wouldn't work. So you'd have to set a main loop before calling that. I'm not sure about the impact of doing so in asyncio-based web servers like gunicorn w/aiohttp or uvicorn. From running in production for the last quarter I never had any issue.
- That loop is used to maintain connection keepalive. Running RPC and the event loop to RPC completion works until the remote end timeout the connection (because no loop is running to reply to pings). I believe normal gRPC library has automatic retry in this case but grpclib would just stop working entirely.
- One workaround I use is to call `channel.close()` after every RPC to always get a fresh connection, but that's not efficient at all.

whs:
If I may suggest, [protobuf-ts](https://github.com/timostamm/protobuf-ts/tree/master/packages/grpc-transport) has a swappable transport implementation for RPC where you can swap between gRPC, twirp and grpcweb.

theartofdevel:
any news? need sync grpc client...

wpdonders:
[...]

Issue #56, opened 2020-05-23
Spilt up the protoc plugin to make gRPC content configurable


nat-n:
Currently there is a single plugin with limited support for generating gRPC client code, and there is demand to cover servicer generation #19, with grpclib or grpcio, or none  #44 #20 .

Since AFAIK there is not standard way to provide arguments to protoc plugins, I propose that the plugin be exposed as multiple distinct executables, one to generate just the protobuf dataclasses, one to generate the dataclasses and grpclib client/server and one to generate dataclasses and grpcio client/server.

Are there other approaches worth considering?

roblabla:
> Since AFAIK there is not standard way to provide arguments to protoc plugins, I propose that the plugin be exposed as multiple distinct executables

this is wrong. We can pass arguments to protoc just fine, e.g. `protoc --python_betterproto_out=out --python_betterproto_opt=enable-grpc` will put `enable-grpc` in the CodeGeneratorRequest's `parameter` field.

Gobot1234:
Yeah this is an old issue, we actually use custom options https://github.com/danielgtaylor/python-betterproto/blob/master/src/betterproto/plugin/parser.py#L75

PR #201, opened 2021-01-21
Grpcio support

### LINKED ISSUE #44: Generation without gRPC
I would like to use betterproto for the generation of code from proto messages but without the usage of gRPC. However, there are still services defined in my .proto files for which I will be generating code with a different plugin. Thus, I need to disable the service generation of the betterproto plugin which, as far as I can tell, is not possible at the moment.

roblabla:
Fixes #44 

This PR does two thing: it introduces a `grpc-kind` flag that allows specifying which grpc backend we want, if any (none, grpcio, grpclib). It defaults to grpclib for backward compatibility.

Secondly, it adds a new gRPC codegen backend for grpcio. This codegen backend generates base classes that can be registered against an aio grpcio server.

It's fairly early, I literally just whipped this up in a couple hours, but wanted some feedback on the design. I'm not very familiar with python! The base class generated by the new backend looks like this:

```python
class ServiceName(better_proto.grpcio.ServiceBase):
    async def method(
        self, request: ty, context: grpc.aio.ServicerContext
    ) -> ty:
    async def stream_stream_method(
        self, request: AsyncIterator[ty], context: grpc.aio.ServicerContext
    ) -> AsyncGenerator[ty, None]:

    __proto_path__ = ""

    @property
    def __rpc_methods__(self):
```

There are three significant differences compared to the `grpclib` backend:

- The methods take the request object as a single argument. I find this easier to deal with - in my app, I have some fairly large objects going over the network, and getting a bunch of arguments can be bothersome.
- The returned type of server-streaming requests is `AsyncGenerator` instead of `AsyncIterator`. This comes down to a difference between `grpcio` and `grpclib`: `grpcio` expects an `AsyncGenerator` from server-streaming requests, whereas `grpclib` expects an `AsyncIterator`. It might be possible to create a wrapper to unify both backends, but I question how wise this would be. I personally much prefer `grpcio`'s way of returning an `AsyncGenerator` - yielding within the request is much more natural than returning an iterator, especially for stream_stream requests!
- ~~I added a method called `register_to_server`, which registers a given implementation against the grpcio server. Grpcio's official stubs instead have an `add_{{typename}}_to_server` standalone function. I figured making it a method would be cleaner, as it avoids the necessary dance to turn typename lowercase.~~ We now have a `betterproto.grpc.grpcio_server.register_servicers` standalone function to register multiple servicers to a grpcio server.

TODO:
- Make sure the grpcio backend works (I didn't test it yet).
- Rename the grpcio backed to grpcio-aio, to potentially support a sync grpcio support in the future (see #20).

### DIFF
--- a/src/betterproto/templates/template.py.j2
+++ b/src/betterproto/templates/template.py.j2
@@ -21,6 +21,9 @@ from betterproto.grpc.grpclib_server import ServiceBase
 {% if output_file.services and output_file.parent_request.options.grpc_kind == "grpclib" %}
 import grpclib
 {% endif %}
+{% if output_file.services and output_file.parent_request.options.grpc_kind == "grpcio" %}
+import grpc
+{% endif %}
 
 
 {% if output_file.enums %}{% for enum in output_file.enums %}
@@ -240,6 +243,60 @@ class {{ service.py_name }}Base(ServiceBase):
         ),
         {% endfor %}
         }
+
+{% endfor %}
+{% endif %}
+
+{% if output_file.parent_request.options.grpc_kind == "grpcio" %}
+{% for service in output_file.services %}
+class {{ service.py_name }}Base:
+    {% if service.comment %}
+{{ service.comment }}
+
+    {% endif %}
+
+    {% for method in service.methods %}
+    async def {{ method.py_name }}(self
+        {%- if not method.client_streaming -%}
+            , request: "{{ method.py_input_message_type }}"
+        {%- else -%}
+            {# Client streaming: need a request iterator instead #}
+            , request_iterator: AsyncIterator["{{ method.py_input_message_type }}"]
+        {%- endif -%}
+            , context: grpc.aio.ServicerContext
+            ) -> {% if method.server_streaming %}AsyncGenerator["{{ method.py_output_message_type }}", None]{% else %}"{{ method.py_output_message_type }}"{% endif %}:
+        {% if method.comment %}
+{{ method.comment }}
+
+        {% endif %}
+        context.set_code(grpc.StatusCode.UNIMPLEMENTED)
+        context.set_details('Method not implemented!')
+        raise NotImplementedError('Method not implemented!')
+
+    {% endfor %}
+
+    def register_to_server(self, server: grpc.aio.Server):
+        rpc_method_handlers = {
+        {% for method in service.methods %}
+            "{{ method.proto_name }}":
+            {% if not method.client_streaming and not method.server_streaming %}
+                grpc.unary_unary_rpc_method_handler(
+            {% elif method.client_streaming and method.server_streaming %}
+                grpc.stream_stream_rpc_method_handler(
+            {% elif method.client_streaming %}
+                grpc.stream_unary_rpc_method_handler(
+            {% else %}
+                grpc.unary_stream_rpc_method_handler(
+            {% endif %}
+                    self.{{ method.py_name }},
+                    request_deserializer={{ method.py_input_message_type }}.FromString,
+                    response_serializer={{ method.py_input_message_type }}.SerializeToString,
+                ),
+        {% endfor %}
+        }
+        generic_handler = grpc.method_handlers_generic_handler(
+            "{{ service.proto_path }}", rpc_method_handlers)
+        server.add_generic_rpc_handlers((generic_handler,))
 
 {% endfor %}
 {% endif %}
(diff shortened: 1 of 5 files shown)


src/betterproto/templates/template.py.j2
```diff
@@ -237,6 +245,61 @@ class {{ service.py_name }}Base(ServiceBase):
...
+            {% endif %}
+                    self.{{ method.py_name }},
+                    request_deserializer={{ method.py_input_message_type }}.FromString,
+                    response_serializer={{ method.py_input_message_type }}.SerializeToString,
+                ),
+        {% endfor %}
+        }
+        generic_handler = grpc.method_handlers_generic_handler(
+            "{{ service.proto_path }}", rpc_method_handlers)
+        self.add_generic_rpc_handlers((generic_handler,))
```
nat-n, line 299: should this be?
```suggestion
        server.add_generic_rpc_handlers((generic_handler,))
```

nat-n (commented):
Thanks for working on this. It looks great! I'd been meaning to do it myself but not gotten around to it...

Yes having some unit tests would be a good idea. I'd suggest looking at how the tests are done for the grpclib servicer, hopefully it's possible to avoid spinning up a server over TCP, though I'm not sure if is.

Having something like register_to_server is a good idea, though I wonder if an instance method is the best way to do it. How about instead defining a `__rpc_handlers__` field on the generated class, and then providing a `betterproto.register_servicers(service, "grpc.aio.Server", *servicer: Type)` utility method to finish the job?

related #20 #56 #44 #19

roblabla:
> Having something like register_to_server is a good idea, though I wonder if an instance method is the best way to do it. How about instead defining a __rpc_handlers__ field on the generated class, and then providing a betterproto.register_servicers(service, "grpc.aio.Server", *servicer: Type) utility method to finish the job?

The big downside of this is that it means the betterproto lib will have to depend on `grpcio` - and while depending on grpclib isn't a big problem since it's a pure python dependency, grpcio is a Cython dependency that links to C code, so it can be a very annoying dependency to have.

Is this a problem? And if it is, how do we work around it? Does Python have optional dependencies/modules?

nat-n:
> The big downside of this is that it means the betterproto lib will have to depend on `grpcio` - and while depending on grpclib isn't a big problem since it's a pure python dependency, grpcio is a Cython dependency that links to C code, so it can be a very annoying dependency to have.
> 
> Is this a problem? And if it is, how do we work around it? Does Python have optional dependencies/modules?

grpcio can be made an optional dependency of betterproto. This can be done using [the same mechanism](https://github.com/danielgtaylor/python-betterproto/blob/master/pyproject.toml#L44) by which the plugin dependencies are only installed if protobuf[compiler] is specified.

Fortunately the `register_servicers` method I suggest can be implemented without a hard dependency on grpcio if it does a lazy import within the function. I'm thinking something like:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import grpc
       
def register_servicers(server, "grpc.aio.Server", *servicers: Type):
    from grpc import method_handlers_generic_handler

    server.add_generic_rpc_handlers(
        tuple(
            method_handlers_generic_handler(
                servicer.__proto_path__, 
                servicer.__rpc_handlers__
            )
            for servicer in servicers
        )
    )
```

It's possible that the import could fail with a runtime error, but quite unlikely in practice, since the intended use case implies the user has already imported it elsewhere.

The type checker specific import of grpc is fine because it'll only run in dev, and protobuf is always available as a dev dependency. 

This of course requires `__proto_path__` to also be set on the generated class.

isac322:
I am excited for grpcio supporting. Is there anything I can do to help this get merged (faster) in master ?

roblabla:
Yeah, sorry, I've been meaning to get back to this, but life's been keeping me busy. I'm not going to be able to get back to this for at least a week or two. Hopefully after that, I should have some time to push this PR further towards completion.

Thanks for the link to the testing guide, that will definitely be useful!

jalaziz:
We've recently come across a need for this as `grpclib` is missing certain important features (e.g. `grpc. max_connection_age_ms`) and we'd really like to avoid dealing with the awful developer experience of standard grpc libraries.

@roblabla any chance you're still working on this? I'd be happy to try and help get this over the line if you're unlikely to work on this anytime soon.

roblabla:
Unfortunately, I'm not working on this anymore, and I don't have the bandwidth to bring this to the finish line.

jalaziz:
I've started a new rebased PR in #328.

PR #202, opened 2021-01-25
Fix CI

nat-n:
Something is funny with the CI. It seems that rebuilding the lockfile with the latest version of poetry fixes the issue.


danielgtaylor (approved):
LGTM :+1:

Issue #36, opened 2020-05-07
to_dict() method converts integer field values to strings


ahobsonsayers:
When using the to_dict method of a message, integer fields are converted to strings. This can be see on [this line](https://github.com/danielgtaylor/python-betterproto/blob/ce9f492f506af347e2e4f54c33077b3f579d19bb/betterproto/__init__.py#L763) in the source code.

Why are integers converted to strings, while all other types are left as is? Is this a bug that i can open a PR for, or it deliberate? Any help is much appreciated as this seems like a bit odd behaviour

nat-n:
That seems odd, any insight? @danielgtaylor

jameslan:
[Spec](https://developers.google.com/protocol-buffers/docs/proto3#json_mapping) requires it to be a string.

There was a [discussion](https://github.com/protocolbuffers/protobuf/issues/2679) on protobuf repo.

But I do not quite understand the [`list` handling](https://github.com/danielgtaylor/python-betterproto/blob/a46979c8a68827d4cdb0f20d2d7e345d391f4ddf/betterproto/__init__.py#L836) of int64 types. @danielgtaylor

boukeversteegh:
> But I do not quite understand the [`list` handling](https://github.com/danielgtaylor/python-betterproto/blob/a46979c8a68827d4cdb0f20d2d7e345d391f4ddf/betterproto/__init__.py#L836) of int64 types

As far as I can see, the list implementation is equivalent to the single int64 type. It just converts all ints in the list to a string. Or is something else unclear about it?

jameslan:
Oh that's for `repeated`. I forgot about it. Then there's no question about the code.

samschlegel:
One thing I know we found confusing when looking into betterproto is that to_dict converts everything to the JSON representation, but that isn't really clearly documented. It makes sense with how to_json just json.dumps the output of to_dict, since without this conversion you'd have no way to ensure it matches the spec, but it's confusing when you're expected to_dict to return pythonic values like `datetime` or in this case an `int`, which leads this behavior to be seen as a bug

boukeversteegh:
[...]

PR #203, opened 2021-01-29
add to/from_pydict methods

crflynn:
This PR adds `to_pydict()` and `from_pydict()` to parse and serialize messages into python dictionaries, alternative to the current `to/from_dict()` methods which cater to JSON serialization.

In addition to defining messages, our protobuf definitions double as spark table schemas. Ingesting messages requires declaring the spark schema explicitly, and so we do that by mapping the betterproto message dataclass field types into spark sql types. The ingestion itself is done by decoding into a betterproto message, calling `to_dict()`, and then adding additional business logic to transform the JSON-prepared datetime strings into datetime objects. To do this we have to write a bit of extra code that looks like this:

```python
def get_decoder(message_type: dataclass) -> Callable:
    """Get a decoder used to create a udf."""

    # determine fields which are datetimes so we can convert them from strings
    datetime_fields = [
        name
        for name, field in message_type().__dataclass_fields__.items()
        if field.type == datetime
    ]

    def decoder(s):
        """betterproto converts datetimes to strings on to_dict() so we convert it back
        using dateutil (iso8601) so that it goes into the spark table as a proper datetime
        """
        asdict = message_type().parse(bytes(s)).to_dict(casing=betterproto.Casing.SNAKE)
        for field in datetime_fields:
            if field in asdict:
                asdict[field] = parser.parse(asdict[field])
        return asdict

    return decoder
```

With this change our code would look like this, and the output dictionary value types match those declared in the dataclasses:

```python
def get_decoder(message_type: dataclass) -> Callable:
    """Get a decoder used to create a udf."""

    def decoder(s):
        return message_type().parse(bytes(s)).to_pydict(casing=betterproto.Casing.SNAKE)

    return decoder
```

An example copied from the test:

```python
@dataclass
class TestDatetimeMessage(betterproto.Message):
    bar: datetime = betterproto.message_field(1)
    baz: timedelta = betterproto.message_field(2)

test = TestDatetimeMessage().from_dict(
    {"bar": "2020-01-01T00:00:00Z", "baz": "86400.000s"}
)

print(test.to_pydict())
# {'bar': datetime.datetime(2020, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), 'baz': datetime.timedelta(days=1)}
```

This relates to some of the discussion happening in #36 with respect to datetimes. Personally this is the behavior that I expected for the `to_dict()` function, expecting it to act more like `dataclasses.asdict` but omitting defaults and handling casing.

Thanks a ton for `betterproto`. I'm happy to iterate on this if there are any requested changes.


guysz:
Voted +1
I can assist if needed

Gobot1234:
> Voted +1 I can assist if needed

If you'd be willing to pick this up with the feedback I've suggested, I'd be very grateful.

tcamise-gpsw:
This does in fact satisfy my request from #316 (after the addition of a `meta.proto_type == TYPE_ENUM` case in the `to_pydict` method).

@Gobot1234, in regards to minimizing the code duplication... I took a stab at your generator idea. I don't think it was as clean as you imagined (or maybe I'm missing something). I also tried another approach of merging the two to/from_dict methods together into one and conditionally doing things depending on whether or not it is a to/from_pydict. 

The generator approach is here: https://github.com/tcamise-gpsw/python-betterproto/commit/e1cb903f882446124a68bb7e54f234dda614504d
The merging approach is here: https://github.com/tcamise-gpsw/python-betterproto/commit/b749ad4cbf4910915ce3344fa24ae1473fd29643

Do either of these look acceptable? If so, i can open a PR and try to set up the poetry environment for more testing.

Gobot1234:
> This does in fact satisfy my request from #316 (after the addition of a `meta.proto_type == TYPE_ENUM` case in the `to_pydict` method).
> 
> @Gobot1234, in regards to minimizing the code duplication... I took a stab at your generator idea. I don't think it was as clean as you imagined (or maybe I'm missing something). I also tried another approach of merging the two to/from_dict methods together into one and conditionally doing things depending on whether or not it is a to/from_pydict.
> 
> The generator approach is here: [tcamise-gpsw@e1cb903](https://github.com/tcamise-gpsw/python-betterproto/commit/e1cb903f882446124a68bb7e54f234dda614504d) The merging approach is here: [tcamise-gpsw@b749ad4](https://github.com/tcamise-gpsw/python-betterproto/commit/b749ad4cbf4910915ce3344fa24ae1473fd29643)
> 
> Do either of these look acceptable? If so, i can open a PR and try to set up the poetry environment for more testing.

Hi, sorry for the late reply, either looks fine to me. Thank you for looking into this

einarjohnson:
Any ETA on when this might be released?

Gobot1234:
> Any ETA on when this might be released?

I'm going to try and get this into b5, but no promises.

Gobot1234 (approved):
Assuming #293 is merged, which should make the Message instances have enum fields always be enums,
> This does in fact satisfy my request from https://github.com/danielgtaylor/python-betterproto/issues/316 (after the addition of a meta.proto_type == TYPE_ENUM case in the to_pydict method).

special casing here shouldn't be necessary

Gobot1234:
Thanks for this!

PR #211, opened 2021-02-15
Added `Message.to_dict` and `from_dict` support for repeated timestamps and durations

### LINKED ISSUE #185: Protobuf to JSON/Dict Fails for repeated datetime
I have the following protobuf definition

```
message DateAxis { repeated google.protobuf.Timestamp values = 1; }
```

image

Because of the snippet on line 964, & since native `datetime` objects don't have a `to_dict` method associated with them, the protobuf serialization fails. Need to apply the same `if isinstance(value, datetime):` chaining logic as before to avoid bugs

badge:
- Added support for repeated Timestamp and Duration types in `Message.to_dict` and `from_dict`, which previously raised an error as these use `datetime` and `timedelta` respectively
- Added test which passes in this branch and fails in master
- Original ignored list elements in the same way as the element types are checked in the `to_dict` method, but of course these are important to preserve order

Resolves #185.


nat-n (approved):
Looks good! Thanks for the contribution :)

PR #213, opened 2021-02-19
Use dateutil parser

robinaly:
This PR replaces the datetime parsing from using datetime's fromisoformat (which isn't iso compliant) to dateutil, which is.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -8,6 +8,7 @@ import typing
 from abc import ABC
 from base64 import b64decode, b64encode
 from datetime import datetime, timedelta, timezone
+from dateutil.parser import isoparse
 from typing import (
     Any,
     Callable,
@@ -1044,7 +1045,7 @@ class Message(ABC):
             meta = self._betterproto.meta_by_field_name.get(field_name)
             if not meta:
                 continue
-
+            
             if value[key] is not None:
                 if meta.proto_type == TYPE_MESSAGE:
                     v = getattr(self, field_name)
@@ -1052,7 +1053,7 @@ class Message(ABC):
                         cls = self._betterproto.cls_by_field[field_name]
                         if cls == datetime:
                             v = [
-                                datetime.fromisoformat(item.replace("Z", "+00:00"))
+                                isoparse(item)
                                 for item in value[key]
                             ]
                         elif cls == timedelta:
@@ -1063,10 +1064,12 @@ class Message(ABC):
                         else:
                             v = [cls().from_dict(item) for item in value[key]]
                     elif isinstance(v, datetime):
-                        v = datetime.fromisoformat(value[key].replace("Z", "+00:00"))
+                        v = isoparse(value[key])
                         setattr(self, field_name, v)
                     elif isinstance(v, timedelta):
                         v = timedelta(seconds=float(value[key][:-1]))
+                        setattr(self, field_name, v)
+                    elif isinstance(v, bool):   
                         setattr(self, field_name, v)
                     elif meta.wraps:
                         setattr(self, field_name, value[key])
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1045,17 +1045,14 @@ class Message(ABC):
             meta = self._betterproto.meta_by_field_name.get(field_name)
             if not meta:
                 continue
-            
+
             if value[key] is not None:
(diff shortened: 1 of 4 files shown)


src/betterproto/__init__.py
```diff
@@ -1063,11 +1061,13 @@ def from_dict(self: T, value: Dict[str, Any]) -> T:
                         else:
                             v = [cls().from_dict(item) for item in value[key]]
                     elif isinstance(v, datetime):
-                        v = datetime.fromisoformat(value[key].replace("Z", "+00:00"))
+                        v = isoparse(value[key])
                         setattr(self, field_name, v)
                     elif isinstance(v, timedelta):
                         v = timedelta(seconds=float(value[key][:-1]))
                         setattr(self, field_name, v)
+                    elif isinstance(v, bool):
```
nat-n, line 1069: Is this another sneaky bugfix? Could you add a test case to clarify?

tests/test_features.py
```diff
@@ -395,3 +396,56 @@ class Truthy(betterproto.Message):
...
+
+
+def test_iso_datetime():
+    @dataclass
+    class Envelop(betterproto.Message):
+        ts: datetime = betterproto.message_field(1)
+
+    e = Envelop()
+    # valid ISO datetimes according to https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/
+    candidates = """2009-12-12T12:34
```
nat-n, line 408: This set of test cases looks convincing but I'm curious what case you encountered that didn't behave as expected before?

tests/test_features.py
```diff
@@ -395,3 +396,56 @@ class Truthy(betterproto.Message):
     assert t
     t.bar = 0
     assert not t
+
+
+def test_iso_datetime():
+    @dataclass
+    class Envelop(betterproto.Message):
```
nat-n, line 403:
```suggestion
    class Envelope(betterproto.Message):
```

nat-n (changes requested):
Thanks for this.

Once nice thing about this change is that it removes the need to apply the `backports-datetime-fromisoformat` monkey patch for 3.6 support. Would you mind cleaning that up from pyproject.toml and the top of the root module while you're at it?

tests/test_features.py
```diff
@@ -395,3 +396,56 @@ class Truthy(betterproto.Message):
...
+        ts: datetime = betterproto.message_field(1)
+
+    e = Envelop()
```
nat-n, line 406:
```suggestion
    msg = Envelope()
```

tests/test_features.py
```diff
@@ -395,3 +396,56 @@ class Truthy(betterproto.Message):
...
+2010-02-18T16:23:00.33+0600
+2010-02-18T16:00:00.23334444
+2010-02-18T16:00:00,2283
+2009-05-19 143922
+2009-05-19 1439""".split(
+        "\n"
+    )
+
+    for i, candidate in enumerate(candidates):
+        e.from_dict({"ts": candidate})
```
nat-n, line 451:
```suggestion
        assert isinstance(e.from_dict({"ts": candidate}).ts, datetime)
```

tests/test_features.py
```diff
@@ -395,3 +396,56 @@ class Truthy(betterproto.Message):
     assert t
     t.bar = 0
     assert not t
+
+
+def test_iso_datetime():
+    @dataclass
+    class Envelop(betterproto.Message):
+        ts: datetime = betterproto.message_field(1)
```
nat-n, line 404: Could easily extend this test to cover the repeated case as well, probably like:

```
    tss: List[datetime] = betterproto.message_field(2)
```

tests/test_features.py
robinaly: I didn't check all of them. I got to it because it didn't work with nanoseconds ("2010-02-18T16:00:00.23334444")

src/betterproto/__init__.py
robinaly: Removed this off topic addition

tests/test_features.py
robinaly: Added a separate test

robinaly:
Github states that there is still one change requested... But I don't find which it is referring to. Help anyone?

Gobot1234:
@robinaly it's just waiting for re-review

nat-n (approved):

PR #215, opened 2021-02-23
Added support for infinite and nan floats/doubles

### LINKED ISSUE #214: Missing support for infinite and nan floats/doubles
The [protobuf spec](https://developers.google.com/protocol-buffers/docs/proto3#json) supports the special double values:

- `"Infinity"`
- `"-Infinity"`
- `"Nan"`

Note that these are strings, and not custom types as supported by Python's `json` package.

These are supported by the [Python protobuf package](https://github.com/protocolbuffers/protobuf/blob/5911099659dc1d15dd6b1967a6b5afbdc8563db8/python/google/protobuf/json_format.py#L814).

badge:
Resolves #214.

This pull request adds support for the custom double values defined in the [protobuf spec](https://developers.google.com/protocol-buffers/docs/proto3#json):

- `"Infinity"`
- `"-Infinity"`
- `"Nan"`

Support is added by updates to the `Message.to_dict` and `Message.from_dict` methods, and not via `json.loads`, to ensure that only `double` and `float` fields will have their values casted to `float("inf")`, etc.

Furthermore,

- `Message.__eq__` has been updated to consider NaN values equal, otherwise a message may not be equal to itself
- The `test_message_json` and `test_binary_compatibility` test functions have been updated to use the new `dict_replace_nans ` function that replaces NaN float values with the string `"NaN"` (because `assert float("nan") == float("nan")` always fails).
- Added a `infinite_floats` test to check correctness; this works with the `protobuf` package and fails in the current `master` branch

### Warning

For the sake of clarity, `json.loads(message.to_json())` is not equal to itself in the case that a message contains NaN values.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -2,6 +2,7 @@ import dataclasses
 import enum
 import inspect
 import json
+import math
 import struct
 import sys
 import typing
@@ -111,6 +112,12 @@ def datetime_default_gen() -> datetime:
 
 
 DATETIME_ZERO = datetime_default_gen()
+
+
+# Special protobuf json doubles
+INFINITY = "Infinity"
+NEG_INFINITY = "-Infinity"
+NAN = "NaN"
 
 
 class Casing(enum.Enum):
@@ -369,6 +376,51 @@ def _serialize_single(
     return bytes(output)
 
 
+def _parse_float(value: Any) -> float:
+    """Parse the given value to a float
+
+    Parameters
+    ----------
+    value : Any
+        Value to parse
+
+    Returns
+    -------
+    float
+        Parsed value
+    """
+    if value == INFINITY:
+        return float("inf")
+    if value == NEG_INFINITY:
+        return -float("inf")
+    if value == NAN:
+        return float("nan")
+    return float(value)
+
+
+def _dump_float(value: float) -> Union[float, str]:
+    """Dump the given float to JSON
+
+    Parameters
+    ----------
+    value : float
+        Value to dump
+
+    Returns
+    -------
+    Union[float, str]
+        Dumped valid, either a float or the strings
+        "Infinity" or "-Infinity"
+    """
+    if value == float("inf"):
+        return INFINITY
+    if value == -float("inf"):
+        return NEG_INFINITY
+    if value == float("nan"):
+        return NAN
+    return value
+
+
 def decode_varint(buffer: bytes, pos: int) -> Tuple[int, int]:
     """
     Decode a single varint value from a byte buffer. Returns the value and the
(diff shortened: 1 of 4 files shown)


cetanu (approved):
Looks good to me 👍

src/betterproto/__init__.py
```diff
@@ -373,6 +380,51 @@ def _serialize_single(
     return bytes(output)
 
 
+def _parse_float(value: Any) -> float:
+    """Parse the given value to a float
+
+    Parameters
+    ----------
+    value : Any
```
Gobot1234, line 388: I wouldn't document this. I also think value shouldn't be Any and rather str

src/betterproto/__init__.py
```diff
@@ -373,6 +380,51 @@ def _serialize_single(
...
+        return float("nan")
+    return float(value)
+
+
+def _dump_float(value: float) -> Union[float, str]:
+    """Dump the given float to JSON
+
+    Parameters
+    ----------
+    value : float
```
Gobot1234, line 410: I again wouldn't document this

tests/test_inputs.py
```diff
@@ -69,6 +70,56 @@ def module_has_entry_point(module: ModuleType):
     return any(hasattr(module, attr) for attr in ["Test", "TestStub"])
 
 
+def list_replace_nans(l: List) -> List[Any]:
+    """Replace float("nan") in a list with the string "NaN"
+
+    Parameters
+    ----------
+    l : List
```
Gobot1234, line 78:
```suggestion
    l: List
```

tests/test_inputs.py
```diff
@@ -69,6 +70,56 @@ def module_has_entry_point(module: ModuleType):
...
+        x.append(e)
+    return x
+
+
+def dict_replace_nans(d: Dict[Any, Any]) -> Dict[Any, Any]:
+    """Replace float("nan") in a dictionary with the string "NaN"
+
+    Parameters
+    ----------
+    l : Dict[Any, Any]
```
Gobot1234, line 103:
```suggestion
    l: Dict[Any, Any]
```

Gobot1234 (changes requested):
I'm still unsure about documenting utils functions as none of the others are

badge:
> I'm still unsure about documenting utils functions as none of the others are

What’s the downside? Documenting anything makes it easier for new people to contribute to the package; removing committed comments unnecessarily  increases the barrier to entry.

src/betterproto/__init__.py
badge: I think it can be a float too; will check and update.

tests/test_inputs.py
badge: This is in the NumPy docstring spec: https://numpydoc.readthedocs.io/en/latest/format.html#sections

Gobot1234:
It's more to maintain

tests/test_inputs.py
Gobot1234: It's not how the other functions are documented

nat-n (dismissed):
Looks good. Thanks for the contribution :)

nat-n:
Rebased to resolve conflicts and enhanced the tests.

nat-n (approved):

PR #216, opened 2021-02-23
Added docstrings

### LINKED ISSUE #209: ProtoContentBase.proto_file is always the first file of a package
I have a large protobuf package whose messages, etc. are defined over multiple files. However, `ProtoContentBase.proto_file` always refers to the *first* file of the package passed to `protoc`. As a result, when I run `protoc` the comments attached to each class and field are from the first file.

I hacked a solution to this by adding `proto_input_file` to `ProtoContentBase` and passing this as an argument all the way from `generate_code` through `read_protobuf_type`/`read_protobuf_service`, then adding:

- To `OutputTempate`:
```python
    @property
    def input_files_dict(self) -> Dict[str, FileDescriptorProto]:
        """Dictionary mapping filenames to their FileDescriptorProto objects

        Returns
        -------
        Dict[str, FileDescriptorProto]
            Dictionary mapping filenames to their FileDescriptorProto objects
        """
        return {f.name: f for f in self.input_files}
```

- to `ProtoContentBase`
```python
    @property
    def proto_file(self) -> FieldDescriptorProto:
        template = self.output_file
        return template.input_files_dict[self.proto_input_file]
```

I'd be delighted to contribute a proper fix and open a PR for it, but it'd be good to hear if there's a better approach to solving the problem.

badge:
The first commit, [39b8f6f](https://github.com/danielgtaylor/python-betterproto/commit/39b8f6fcbafe4a85f4c191e9a7d5211c3d2042a1), resolves #209 by keeping track of the input filename for each object (though it will not deal with extended messages in different source files). Everything else is somewhat more opinionated—it adds NumPy-style docstrings in place of the existing comments.

Take this example `.proto` source file:

```protobuf
syntax = "proto3";

enum Casing { // Message casing
  UNCHANGED = 0; // Don't change the message
  LOWERCASE = 1; // Convert to lowercase
  UPPERCASE = 2; // Convert to uppercase
  TITLECASE = 3; // Convert to titlecase
}

message Test { // A request to the server
  // String to echo
  string value = 1;
  // Number of extra times to echo
  uint32 extra_times = 2;
  Casing casing = 3; // Message case
}
```

Currently, the only comment captured from this source are the leading comments in `EchoRequest`.

```python
    # String to echo
    value: str = betterproto.string_field(1)
    # Number of extra times to echo
    extra_times: int = betterproto.uint32_field(2)
    casing: "Casing" = betterproto.enum_field(3)

        super().__post_init__()
```

Commits [d608d80](https://github.com/danielgtaylor/python-betterproto/commit/d608d80b4c5e3decaf9f07ee19f3cfc688b1284d) and [/26a6d8e0](https://github.com/danielgtaylor/python-betterproto/commit/26a6d8e01d1aef9f367164f3be0c1141d0c17255) replace the existing `comment` properties of the `ProtoContentBase` class and its subclasses with `docstring` properties specific to their type. For instance, the `docstring ` property of a `FieldCompiler` includes its name and annotation. `get_comment` has also been updated to include trailing comments. `MessageCompiler` and its subclasses now have NumPy-style docstrings, and the output from the above proto source file now looks like this:

```python
class Casing(betterproto.Enum):
    """Message casing

    Attributes
    ----------
    UNCHANGED : int = 0
        Don't change the message
    LOWERCASE : int = 1
        Convert to lowercase
    UPPERCASE : int = 2
        Convert to uppercase
    TITLECASE : int = 3
        Convert to titlecase
    """

    UNCHANGED = 0
    LOWERCASE = 1
    UPPERCASE = 2
    TITLECASE = 3


    """A request to the server

    Parameters
    ----------
    value : str
        String to echo
    extra_times : int
        Number of extra times to echo
    casing : Casing
        Message case
    """

    value: str = betterproto.string_field(1)
    extra_times: int = betterproto.uint32_field(2)
    casing: "Casing" = betterproto.enum_field(3)

        super().__post_init__()
```

These output docstrings can then be used by e.g. Sphinx autodoc with napoleon to generate Python package documentation directly from the .proto source files

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -143,7 +143,7 @@ def get_comment(
             if sci.leading_comments:
                 comments.append(
                     sci.leading_comments.replace("\n", " ").replace("  ", " ").strip()
-            )
+                )
             if sci.trailing_comments:
                 comments.append(
                     sci.trailing_comments.replace("\n", " ").replace("  ", " ").strip()
@@ -155,7 +155,7 @@ def get_comment(
         return ""
     if len(comments) == 1:
         comment = comments[0]
-                else:
+    else:
         comment = " ".join(comments)
 
     if not wrap and not comment_out:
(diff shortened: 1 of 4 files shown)


Gobot1234:
I feel like as this makes your code a lot longer if you've added comments the old way, it would be nice if this was an option for compilation with a flag.

Gobot1234:
Closing this in favour of https://github.com/danielgtaylor/python-betterproto/commit/9b5594adbed39b0f7180f55ff2aee996cafcb08e

Gobot1234:
Oh I will cherry pick https://github.com/danielgtaylor/python-betterproto/pull/216/commits/39b8f6fcbafe4a85f4c191e9a7d5211c3d2042a1 as a separate PR.

Issue #220, opened 2021-03-07
Betterproto fails to compile empty service without any comments.


qria:
Betterproto fails to compile when a service is empty without any comments.

example.proto: 
```
syntax = "proto3";

package hello;

service HelloService {}
```

Output of command `python -m grpc_tools.protoc -I . --python_betterproto_out=. example.proto`

```

Traceback (most recent call last):
  File "/Users/qria/.virtualenvs/inference/bin/protoc-gen-python_betterproto", line 8, in <module>
    sys.exit(main())
  File "/Users/qria/.virtualenvs/inference/lib/python3.8/site-packages/betterproto/plugin/main.py", line 28, in main
    generate_code(request, response)
  File "/Users/qria/.virtualenvs/inference/lib/python3.8/site-packages/betterproto/plugin/parser.py", line 123, in generate_code
    f.content = outputfile_compiler(output_file=output_package)
  File "/Users/qria/.virtualenvs/inference/lib/python3.8/site-packages/betterproto/plugin/compiler.py", line 34, in outputfile_compiler
    return black.format_str(
  File "/Users/qria/.virtualenvs/inference/lib/python3.8/site-packages/black.py", line 725, in format_str
    src_node = lib2to3_parse(src_contents.lstrip(), mode.target_versions)
  File "/Users/qria/.virtualenvs/inference/lib/python3.8/site-packages/black.py", line 836, in lib2to3_parse
    raise exc from None
black.InvalidInput: Cannot parse: 14:0: class HelloServiceBase(ServiceBase):
--python_betterproto_out: protoc-gen-python_betterproto: Plugin failed with status code 1.
```

This is because without comments, betterproto generates code like this:

```
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: example.proto
# plugin: python-betterproto
from dataclasses import dataclass
from typing import Dict

import betterproto
from betterproto.grpc.grpclib_server import ServiceBase
import grpclib


class HelloServiceStub(betterproto.ServiceStub):

class HelloServiceBase(ServiceBase):



    def __mapping__(self) -> Dict[str, grpclib.const.Handler]:
        return {
        }

```

...which is invalid.

Currently using version 2.0.0b3.

We're probably just going to add comments to it but seems like it's a bug worth mentioning.

nat-n:
I believe this issue should be resolved now in 2.0.0b3. Please comment back here if that's not the case.

Issue #221, opened 2021-03-07
Services with same package gets its comment overwritten


qria:
Say you have two files `service1.proto` and `service2.proto` with same package assigned:

service1.proto: 

```protobuf
syntax = "proto3";

package service;

/* Service1 */
service Service1 {}

```

service2.proto:

```protobuf
syntax = "proto3";

package service;

/* Service2 */
service Service2 {}
```

Compiling these together with the command `python3 -m grpc_tools.protoc -I . --python_betterproto_out=./ service1.proto service2.proto` results in this:

service/__init__.py:

```py
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: service1.proto, service2.proto
# plugin: python-betterproto
from dataclasses import dataclass
from typing import Dict

import betterproto
from betterproto.grpc.grpclib_server import ServiceBase
import grpclib


class Service1Stub(betterproto.ServiceStub):
    """Service1"""


class Service2Stub(betterproto.ServiceStub):
    """Service1"""


class Service1Base(ServiceBase):
    """Service1"""

    def __mapping__(self) -> Dict[str, grpclib.const.Handler]:
        return {}


class Service2Base(ServiceBase):
    """Service1"""

    def __mapping__(self) -> Dict[str, grpclib.const.Handler]:
        return {}
```

As you can see the comment for Service2 got overwritten.

I have reproduced this with current master branch, which I believe is version 2.0.0b2 with commit hash 6c1c41e9ccf7d020641e87f82e6419c3393a3841 .

PR #222, opened 2021-03-09
Allow empty services.

qria:
Allow empty service without comment to be compiled.

While you may think that this issue can be easily circumvented by adding comments, it was very hard to do so with issue #221 present, where it deletes comments somewhere in the codebase.

Fixes issue #220


nat-n (approved):
Thanks for the contribution :)

Issue #224, opened 2021-04-02
Enhanchement: Add "troubleshooting" section to the documentation


ShayNehmad-RecoLabs:
After a discussion on Slack with an issue I had, @nat-n and I discussed adding a troubleshooting section to the docs:

![image](https://user-images.githubusercontent.com/70973922/113434516-c82d3100-93e9-11eb-85ce-32cfd72c4305.png)

[Slack link](https://betterproto.slack.com/archives/C014C74KU7N/p1617365727009000)

So I'm opening this issue to correlate to the PR I'm about to open :smile_cat:

nat-n:
Closing this issue in favour of the PR.

PR #225, opened 2021-04-02
Troubleshooting docs 244

ShayNehmad-RecoLabs:
Resolves https://github.com/danielgtaylor/python-betterproto/issues/224


nat-n:
Hi @ShayNehmad-RecoLabs, thanks for working on this.

Now that I see it explained like this, I'm curious if you tried doing `poetry uninstall betterproto; poetry install betterproto`, before just rebuilding the env? I guess you did try running poetry install again after updating the prproject.toml?

I'm also still unsure exactly what caused the issue you experienced. It's surprising given that AFAIK neither betterproto nor poetry are doing anything at all unusual with how the entry_point console_script is created/installed into the venv.

I just want to be sure that whatever goes into a troubleshooting section like this represents the problem well and gives the simplest solution.

Also poetry now has a smoother workflow for finding/deleting envs: https://python-poetry.org/docs/managing-environments/#listing-the-environments-associated-with-the-project

ShayNehmad-RecoLabs:
I can't really take the time to reproduce - this worked for me, so :shrug:

PR #234, opened 2021-05-12
add hooks for accessing the stream/metadata

iamnoah:
This is just an extra layer in the generated code for accessing the underlying stream to read metadata, etc. Users can implement this instead of the method without suffix for a slightly lower level experience.

The stack becomes:
`__rpc_<method>` handle recv & send -> `<method>__with_stream` default implementation spreads the request -> `<method>` same semantics

### DIFF
--- a/docs/migrating.rst
+++ b/docs/migrating.rst
@@ -43,6 +43,27 @@ You cannot check if a scalar was sent on the wire.
     True
 
 
+Accessing the Stream/Metadata
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you need access to the ``grpclib.server.Stream`` object, either to access the call metadata
+or for other reasons, you can implement ``<method>__with_stream`` instead:
+
+.. code-block:: python
+
+    # this method also receives the unspread request object
+    async def example_unary_unary__with_stream(
+        self, stream: grpclib.server.Stream, request: "ExampleRequest"
+    ) -> "ExampleResponse":
+        # metadata is a MultiDict
+        auth = stream.metadata.get("authorization")
+        # grpclib.events.listen is probably a better way to check auth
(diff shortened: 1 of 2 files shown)


iamnoah:
I am not actually using this. I ended up using [grpclib listeners](https://grpclib.readthedocs.io/en/latest/events.html#grpclib.events.RecvMessage) to access the raw stream and metadata when needed. Worst case, you can stick it in a contextvar, so no need to make the generated code more confusing.

dchebakov:
Can you please give an example how you can inject tokens into metadata using any async framework if you don't have access to the request object?

jbkoh:
Hi @iamnoah, I'm highly interested in your solution using native grpclib for this. I forked betterproto and implemented something like this PR but I'd love to learn your solutions. Any advice would be appreciated. Thanks!

iamnoah:
I've used two workarounds. 

One is to override `__mapping__` in your service class and do something like:

```python
 return super().__mapping__() | {
            "/MyService/Get": grpclib.const.Handler(...)
```

The other is to create a context var, add a [RecvMessage listener](https://grpclib.readthedocs.io/en/latest/events.html#grpclib.events.RecvMessage), and use it to decorate the method, setting the context var:

```python
method_stream: contextvars.ContextVar[Stream] = contextvars.ContextVar("grpclib_stream")

async def stream_listener(event: RecvRequest):
    fn = event.method_func

    @functools.wraps(fn)
    async def wrapper(stream: Stream) -> None:
        method_stream.set(stream)
        await fn(stream)

    event.method_func = wrapper
```

Either way, you're basically bypassing betterproto to use the grpclib API directly. Since the generated code doesn't do a lot, that's NBD.

jbkoh:
Thanks. This is super helpful!

PR #242, opened 2021-05-20
error message when None appears in a repeated value [was] not helpful

iamnoah:
If a `None` slips into a list to be serialized, the resulting error message doesn't provide any information about where the problem was. This preemptively detects `None` values in a collection and raises a message that should make it easy to track down.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -739,6 +739,8 @@ class Message(ABC):
                     output += _serialize_single(meta.number, TYPE_BYTES, buf)
                 else:
                     for item in value:
+                        if item is None:
+                            raise ValueError(f"None value in list field {field_name} of {self}")
                         output += (
                             _serialize_single(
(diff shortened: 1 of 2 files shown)


src/betterproto/__init__.py
```diff
@@ -739,6 +739,8 @@ def __bytes__(self) -> bytes:
                     output += _serialize_single(meta.number, TYPE_BYTES, buf)
                 else:
                     for item in value:
+                        if item is None:
```
nat-n, line 742: For performance it would be preferable to put the loop in a try block, and then in the except block iterate to find the None. It makes the error case a bit more complex and less efficient, but avoid adding the extra to the happy path.

iamnoah:
I'm skeptical that suggestion would make any meaningful performance impact given that serialization will account for 99% of the time in this function. If getting this merged requires dancing around arbitrary demands like this, I will pass.

PR #254, opened 2021-07-07
fix: Map field edge-case

Luminaar:
Hello! I've hit upon this weird edge-case when using `betterproto` in our project.
It took me a while to find the cause (especially realizing that the field name caused the problem)
but the solution is quite simple. I've tried to find a better way to check for the map field type
but was unable to find one (I'm not very familiar with the protobuf internals).


Gobot1234 (approved):

Luminaar:
Found the issue. The package directive was missing in my test proto file.

Gobot1234 (approved):

Issue #261, opened 2021-08-05
Typing Import missing in streaming API


ashat1701:
When you create something like this Proto: 
```  rpc StreamVideoFrames(stream VideoFrame) returns (Ack);```
You get Stub with missing import
```python
    async def stream_video_frames(
        self, request_iterator: AsyncIterator["VideoFrame"]
    ) -> "Ack":
        raise grpclib.GRPCError(grpclib.const.Status.UNIMPLEMENTED)
```
import line:
```python
from typing import AsyncIterable, Dict, Iterable, Union
```
AsyncIterator is missing(
Version: 2.0.0b3

Gobot1234:
Please use betterproto==2.0.0b3

Issue #198, opened 2021-01-18
Service generation for client streaming  missing a `AsyncIterator` import


Lugoues:
The template for client streaming uses AsyncIterator - [template.py#L180](https://github.com/danielgtaylor/python-betterproto/blob/1d54ef8f9936899f86b2fe79da3e301f3096dc4d/src/betterproto/templates/template.py.j2#L180)
But the import code generation doesn't include it - [models.py#590](https://github.com/danielgtaylor/python-betterproto/blob/1d54ef8f9936899f86b2fe79da3e301f3096dc4d/src/betterproto/plugin/models.py#L590)

Gobot1234:
#183 fixes this.

Lugoues:
Oh awesome! Sorry about the duplicate. I didn't see that when I was searching.

PR #264, opened 2021-08-24
Include AsyncIterator import for both clients and servers

lazytype:
resolves https://github.com/danielgtaylor/python-betterproto/issues/198
resolves https://github.com/danielgtaylor/python-betterproto/issues/261

Any guidance on testing would be appreciated.

### DIFF
--- a/src/betterproto/grpc/grpclib_server.py
+++ b/src/betterproto/grpc/grpclib_server.py
@@ -1,5 +1,5 @@
 from abc import ABC
-from collections import AsyncIterable
+from collections.abc import AsyncIterable
 from typing import Callable, Any, Dict
 
 import grpclib
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -655,7 +655,8 @@ class ServiceMethodCompiler(ProtoContentBase):
             self.output_file.typing_imports.add("Union")
 
         # Required by both client and server
-        self.output_file.typing_imports.add("AsyncIterator")
+        if self.client_streaming or self.server_streaming:
+            self.output_file.typing_imports.add("AsyncIterator")
 
         super().__post_init__()  # check for unset fields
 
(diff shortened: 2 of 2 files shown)


src/betterproto/grpc/grpclib_server.py
```diff
@@ -1,5 +1,5 @@
 from abc import ABC
-from collections import AsyncIterable
+from collections.abc import AsyncIterable
```
lazytype, line 2: This fixes the warning:

> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working

TomaszMacudzinski-TomTom (dismissed):

src/betterproto/plugin/models.py
```diff
@@ -653,8 +653,9 @@ def __post_init__(self) -> None:
             self.output_file.typing_imports.add("AsyncIterable")
             self.output_file.typing_imports.add("Iterable")
             self.output_file.typing_imports.add("Union")
-        if self.server_streaming:
-            self.output_file.typing_imports.add("AsyncIterator")
+
+        # Required by both client and server
+        self.output_file.typing_imports.add("AsyncIterator")
```
roblabla, line 658: This will import `AsyncIterator even when both `self.client_streaming` and `self.server_streaming` are both false (e.g. for normal requests of type `rpc X(A) returns B`.

This should really be:

```suggestion
        if self.client_streaming or self.server_streaming:
            self.output_file.typing_imports.add("AsyncIterator")
```

src/betterproto/grpc/grpclib_server.py
kalzoo: 🙏

kalzoo (approved):
Thanks @lazytype!

PR #268, opened 2021-09-21
Proto3 field presence

### LINKED ISSUE #35: support for proto3 `optional` (explicit field presence)
In betterproto it seems like all fields are optional, in that `proto.field = None` is a valid operation. However, there doesn't seem to be any way to indicate a missing field when using the constructor; instead, missing fields get their default values (e.g., 0 for an int).

```
message Pizza {
  string name  = 1;
  int32 radius = 2; 
}

>>> pizza = betterproto.Pizza(name='neapolitan', radius=9)
betterproto.Pizza(name='neapolitan', radius=9)
>>> pizza.radius = None
>>> pizza
betterproto.Pizza(name='neapolitan', radius=None)

>>> pizza = betterproto.Pizza(name='neapolitan')
betterproto.Pizza(name='neapolitan', radius=0)

>>> pizza = betterproto.Pizza(name='neopolitan', radius=None)
*** SyntaxError: invalid syntax
```

If I understand [this announcement](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md) correctly, the `optional` keyword is now available in proto3 in experimental release 3.12. Is there any plan for the betterproto compiler to take advantage of the new proto3 `optional` keyword to differentiate optional and required fields? It would be great if only explicitly `optional` fields permitted assignment of `None`, whether directly or through the constructor.

### LINKED ISSUE #229: proto3 support for optional
Proto3 supports the optional keyword. All libraries supporting this keyword needs to indicate this explicitly. Otherwise the complication will fail with `is a proto3 file that contains optional fields, but code generator protoc-gen-python_betterproto hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.`

Would it be possible to enable proto3 optional?

See this issue for more detail: https://github.com/vmagamedov/grpclib/issues/125

roblabla:
Fixes #229

Fixes #35 

Adds proto3 field presence support. Most of the infrastructure was already present to support it, all I had to do was add some plumbing:

- When a field is specified to be proto3_optional, make its default value None (instead of PLACEHOLDER)
- Tell the protobuf that we're compatible with proto3 field presence

The rest is mostly updating the documentation to take into account the field presence.


kalzoo:
@nat-n, @boukeversteegh , or @danielgtaylor  - mind having a look, or at least approving the GHA workflow?

roblabla:
Superseded by #281 . Thanks @kalzoo :)

PR #281, opened 2021-11-16
Proto3 field presence

kalzoo:
This finishes the work done on https://github.com/danielgtaylor/python-betterproto/pull/268 by @roblabla . See that PR for more information


alfredgunnar:
@kalzoo What's the status on this? My team would benefit alot from having this in place.

vthib:
I am using this PR and have found some issues with it. I have written a PR to fix those here: #292

kalzoo:
@alfredgunnar status is that the tests were caught up in an issue with `black`, I pushed a fix and @Gobot1234 straightened it out but I've been busy since. Now we're on to a new issue in the suite which I'll debug as time permits. Once this pipeline is green, we should be set to merge and release a new version. If you want to accelerate this, feel free to PR a test pipeline fix.

jdlourenco:
Hey @kalzoo are the tests only failing on the pipeline? I ran them successfully locally.

kalzoo:
This is fixed - CI failure was caused by a mistake in merge from `master` and resulting regression - see 3a41992. 

Since I made the final changes, I'd like to ask @Gobot1234 (and others) for review.

Gobot1234:
Everything here looks good to me!

Gobot1234 (approved):

src/betterproto/__init__.py
```diff
@@ -1041,6 +1085,9 @@ def to_dict(
                         ]
                     if value or include_default_values:
                         output[cased_name] = value
+                elif value is None:
+                    if include_default_values:
```
Gobot1234, line 1089: This might be better if you merge these if conditions

kalzoo: Not quite the same, since that would allow cases where `value is None and include_default_values is False` to fall through to later branches. I think I like the clarity of intent here, but is there something I'm missing?

Gobot1234: Oh yeah, you're totally right

PR #284, opened 2021-11-17
Fix Python 3.9 Tests

kalzoo:
1. Constrain Python version to `>=3.6.2,<4.0` (BREAKING CHANGE)
2. Upgrade `black` version in dev-dependencies
3. Don't specify a target version of 3.7 for black; let it infer the target version from the code and environment. See [failure case](https://github.com/danielgtaylor/python-betterproto/runs/4234334625?check_suite_focus=true) with Python 3.6.
3. Improve the output of the test file generation script:

- Don't report success for a task that fails
- Return exit code 1 on failure so that the CI task is marked as such, instead of causing failure in the next task

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ python-dateutil = "^2.8"
 
 [tool.poetry.dev-dependencies]
 asv = "^0.4.2"
-black = "^21.11b0"
+black = "^21.11b1"
 bpython = "^0.19"
 grpcio-tools = "^1.30.0"
 jinja2 = "^2.11.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ python-dateutil = "^2.8"
 
 [tool.poetry.dev-dependencies]
 asv = "^0.4.2"
-black = "^21.11b1"
+black = "^21.11b0"
 bpython = "^0.19"
 grpcio-tools = "^1.30.0"
 jinja2 = "^2.11.2"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ python-dateutil = "^2.8"
 
 [tool.poetry.dev-dependencies]
 asv = "^0.4.2"
-black = "^20.8b1"
+black = "^21.11b0"
```
Gobot1234, line 24: This version of black is bugged iirc

src/betterproto/plugin/compiler.py
```diff
@@ -33,5 +33,5 @@ def outputfile_compiler(output_file: OutputTemplate) -> str:
 
     return black.format_str(
         template.render(output_file=output_file),
-        mode=black.FileMode(target_versions={black.TargetVersion.PY37}),
+        mode=black.FileMode(),
```
Gobot1234, line 36:
```suggestion
        mode=black.Mode(),
```
Might be a good idea to update for this https://github.com/psf/black/blob/main/src/black/__init__.py#L98-L99

Gobot1234:
Apart from a few comments LGTM

pyproject.toml
Gobot1234:
```suggestion
black = "^21.11b1"
```
The old version got yanked

Gobot1234 (approved):

PR #273, opened 2021-10-03
map enum int's into python enums (#157)

ydylla:
This maps deserialized enum integers into real python enum instances.

Closes #157, #169

PR #293, opened 2021-11-29
Map enum int's into Enums redux

### LINKED ISSUE #157: Enum doesn't get mapped on receive
Hi,

Enums seem not to be mapped on init and receive and left as their integer values.

Example proto:
```
message TestRequest {
  Category category = 1;
}

enum Category {
  SOMETHING = 0;
  SOMETHING_ELSE = 1;
}
```

For `TestRequest()`, the `Category` value will be int 0 and not `SOMETHING`.
When receiving `category = 1` over gRPC, it will be also left as int 1.

For deserialization it can be fixed easily by adding yet another branch to `_postprocess_single`. For plain init - I've tried tracing it down, but it seem to go somewhere deep into how dataclasses work

### LINKED ISSUE #169: parse data will change enum value type from enum to int
data = bytes(circuitLine)
                ret = PBCircuitLine().parse(data)  # type: 'PBCircuitLine'

circuitLine:
image
ret:
image

### LINKED ISSUE #291: Accessing Enum name by `str()`
Say I've compiled:
`class Priority(betterproto.Enum):`
`    PRIORITY_LOW = 0`
`    PRIORITY_MEDIUM = 1`
`    PRIORITY_HIGH = 2`

Accessing the string value of each instance (which is defined by `.toString()`) is done similarly to `Enum` class in python, via `Priority.PRIORITY_LOW.name`, for example. 

From usability point of view, I'd like to access this by the pythonic convention of `str(Priority.PRIORITY_LOW)`. (i.e. add `__str__` method to the compiled class). 

Appreciate your consideration.

Gobot1234:
A continuation of #273 containing a reimplementation of betterproto.Enum (with 3.11+ features) as an open set along with some tests for this.

Closes #291
Closes #157
Closes #169


Gobot1234:
This should be ready to go now

joetopp:
+1 to this, for the sake of resolving issue #157

Gobot1234:
Not really sure why the install's failing other than it might just be time to drop 3.6?

Gobot1234:
Actually this is probably just cause poetry dropped 3.6 support

sataset:
@danielgtaylor @kalzoo will this ever be merged or implemented from maintainers?

Current implementation makes no sense. Message enum field should have an appropriate type and not just int.

MicaelJarniac:
Python 3.6 is already at its end of life anyways, and soon 3.7 will be too, so I don't think support for 3.6 should matter that much.

And people using this project, instead of Google's official Python implementation, are likely here because they want newer features, so I don't think most would care for 3.6 anyways.

dingxiong:
Is there any reason this PR is not merged?

Run1e:
If it means anything to a (hopefully) future reviewer, I've been using it in prod for a couple months and it's been working flawlessly.

cetanu:
Just saw this in my review queue, reading now

cetanu (approved):

cetanu:
I think this might have somehow affected the pickling methods that I'm working on

MicaelJarniac:
Do these enums support identity comparison (`color is Color.RED`)?

https://docs.python.org/3/howto/enum.html#comparisons

Gobot1234:
Yes they should

PR #294, opened 2021-11-30
Fix compilation of fields with name identical to their type

guysz:
Defining a dataclass with a field that its name is identical to its type name, and is initialized as a `field` produces `RecursionError`:
```
from dataclasses import dataclass, field
@dataclass
class Foo:
    bool: bool = field()
```
This PR suggests a solution to the problem by using built-in types from the builtins package, so the above code will be changed to:
 ```
from dataclasses import dataclass, field
import builtins
@dataclass
class Foo:
    bool: builtints.bool = field()
```

This PR also reverts #226 because it is solving a more general problem.

### DIFF
--- a/src/betterproto/casing.py
+++ b/src/betterproto/casing.py
@@ -133,16 +133,6 @@ def lowercase_first(value: str) -> str:
     return value[0:1].lower() + value[1:]
 
 
-def is_reserved_name(value: str) -> bool:
-    if keyword.iskeyword(value):
-        return True
-
-    if value in ("bytes", "str"):
-        return True
-
-    return False
-
-
 def sanitize_name(value: str) -> str:
     # https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles
-    return f"{value}_" if is_reserved_name(value) else value
+    return f"{value}_" if keyword.iskeyword(value) else value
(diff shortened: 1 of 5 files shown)


Gobot1234:
I'm not a huge fan of this for two reasons:
1. This makes the code less pythonic by default.
2. This is not an issue that the library by default should have to handle.

Fixing the first one is relatively easy, only add builtins if a name in builtins is a field name, I'll give the second a go so this can also be version conditional in the future.

Gobot1234:
I've opened an issue on the bpo https://bugs.python.org/issue45946 about this.

guysz:
1. Fixed - add `builtins.` prefix only when needed.
2. Why do you think the library should not handle such a case? In practice, the tool generates a broken code from a valid proto file. Specifically, in my use-case, I use a third-party proto definition that falls into this issue which means I can't use this tool without the fix (and I really like to use it :))

Gobot1234:
I'm saying in an ideal world this shouldn't be fixed by us but by cpython.

Gobot1234:
Can you add a test for this?

Gobot1234:
Assuming tests pass this looks much better than the original solution to this. Thank you!

Gobot1234 (approved):

PR #304, opened 2021-12-24
Format field comments also as docstrings

### LINKED ISSUE #303: Format field comments as real docstrings
This line https://github.com/danielgtaylor/python-betterproto/blob/master/src/betterproto/templates/template.py.j2#L49 formats the comments on class attributes like this.

With the example from the readme:
```proto
// This is the class docstring
message EchoRequest {
    string value = 1;
    // This is a field docstring
    uint32 extra_times = 2;
}
```

It generates this Python code:
```python
@dataclass
class EchoRequest(betterproto.Message):
    """This is the class docstring"""

    value: str = betterproto.string_field(1)
    # This is a field docstring
    extra_times: int = betterproto.uint32_field(2)
```

But a slightly different formatting would make it a docstring recognized by IDEs:
```python@dataclass
class EchoRequest(betterproto.Message):
    """This is the class docstring"""

    value: str = betterproto.string_field(1)
    extra_times: int = betterproto.uint32_field(2)
    """This is a field docstring"""
```
image

Would you mind changing the formatting here? Or are there specific reasons why it produces the current output format?

michaelosthege:
Thanks for the pointers @Gobot1234 !

I didn't manage to run the tests (`poe` command not found) and I'm not sure if/how I should generate the code of the project files.
I gotta jump now (christmas dinner..), but please feel free to give additional instructions or take over the branch!

Enjoy the holiday!

PS.: I tested it on my code by generating & installing the Wheel. Looks like in comparison with the current release it puts all classes in `__init__.py` now? Also FYI: The line `from betterproto.grpc.grpclib_server import ServiceBase` was unnecessarily included in the generated file.

Closes #303

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -147,17 +147,13 @@ def get_comment(
                 sci_loc.leading_comments.strip().replace("\n", ""), width=79 - indent
             )
 
-            if path[-2] == 2 and path[-4] != 6:
-                # This is a field
-                return f"{pad}# " + f"\n{pad}# ".join(lines)
+            # This is a field, message, enum, service, or method
+            if len(lines) == 1 and len(lines[0]) < 79 - indent - 6:
+                lines[0] = lines[0].strip('"')
+                return f'{pad}"""{lines[0]}"""'
             else:
-                # This is a message, enum, service, or method
-                if len(lines) == 1 and len(lines[0]) < 79 - indent - 6:
-                    lines[0] = lines[0].strip('"')
-                    return f'{pad}"""{lines[0]}"""'
-                else:
-                    joined = f"\n{pad}".join(lines)
-                    return f'{pad}"""\n{pad}{joined}\n{pad}"""'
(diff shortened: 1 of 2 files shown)


Gobot1234:
Merry christmas,
I have a quick question, does this also apply the same formatting to Enum member comments as I think that would be a good change as well as be consistent with this.

michaelosthege:
Yes - and they were actually not placed correctly.
My edit just takes out the exception that was made for class attributes.

Now I also added line breaks after the docstring to make it easier to read. (I checked with `black`, but it didn't care about that line break either way.)

```python
class SomeEnum(betterproto.Enum):
    VALUE_A = 0
    """Val 1"""

    VALUE_B = 5
    """Val 2"""

    VALUE_C = 1234
    VALUE_D = 6
```

Gobot1234 (approved):

davidmankin:
I came to github to request the same fix and now I see that it's already implemented in a PR! 🎉 

I hope this gets merged before 2.0.0b5 release…

cetanu (approved):

PR #308, opened 2021-12-30
Add a benchmarking script

Gobot1234:
Also includes a change to `Message._betterproto` to speed it up

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,0 +1,50 @@
+name: Run Performance tests
+
+on: [pull_request]
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


cetanu (approved):

.github/workflows/perf.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Install dependencies
+        shell: bash
+        run: |
+          poetry run python -m pip install pip -U
+          poetry install
+
+      - name: Run benchmarks
+        run:
+          poetry run python -m asv continuous master ${{ github.ref }}
```
cetanu, line 50: I guess this needs to be changed slightly so that it can handle pull-requests

Issue #287, opened 2021-11-19
The service methods should be passed the message objects, **not** their fields.


FHTMitchell:
I absolutely love betterproto but some of my protobuf files are very complicated with many fields and nested structures and I need to pass a lot of data around to a lot of different places. The objects I use to pass those around are, naturally, the dataclasses betterproto provides. However, betterproto splits these objects up so I end up with code that looks like 

```python
class VolService(VolServiceBase):
    async def run_vol_pipeline(
        self,
        oc_key: "OcKey",  # nested
        source: str,
        source_topic: str,
        cycle_id: str,
        cycle_time: datetime,
        chain_size: int,
        cycle_size: int,
        priced_count: int,
        unavailable_count: int,
        bad_count: int,
        snapshots: Optional[List["OptionSnapshot"]],  # repeated nested
    ) -> AsyncIterator["PyvolResult"]:
        
        req = PyvolRequest(
        oc_key,
        source,
        source_topic,
        cycle_id,
        cycle_time,
        chain_size,
        cycle_size,
        priced_count,
        unavailable_count,
        bad_count,
        snapshots,
        )
 ```
 
whereas if you pass the object the method signature is cleaner and users can easily extract fields anywhere down the call stack.

```python
class VolService(VolServiceBase):
    async def run_vol_pipeline(self, request: PyvolRequest) -> AsyncIterator["PyvolResult"]:
        ...
```

This is actually how grpclib works! I also feel the same way about the clients. Why not allow us to pass a message object directly instead of having to do something like

```python
async def query(req: fx.FxRequest) -> fx.FxResponse:
    service = fx.FxServiceStub(channel)  # the service stub
    asdict = {f.name: getattr(req, f.name) for f in dataclasses.fields(req)}
    return await service.run_fx_pipeline(**asdict)
```

It gives the user more control about how they want to handle data. It's fully typed so we don't lose anything.

Gobot1234:
This sounds like a good idea to me, interested as to @danielgtaylor's POV on this.

efokschaner:
I also would like this. I found myself having to reconstruct proto messages from the params because I have other functions which are expecting to operate on the defined message.

Gobot1234:
Fixed in https://github.com/danielgtaylor/python-betterproto/commit/d260f071e01657287888e15b73623305b56b4d9f

augi:
Hello, I know I'm late to the party but I would like to express my concerns 🙏 

[...]

PR #311, opened 2022-01-05
Client and Service Stubs take the input type object defined in the proto

efokschaner:
This addresses https://github.com/danielgtaylor/python-betterproto/issues/287

The PR is split in to commits with separated concerns, so take advantage of that for review.

I had a bit of an issue with imports getting sorted automatically. I tried to make sure I was only running black though I may have messed up somewhere. Rather than excluding them all, i stuck them in a dedicated commit: https://github.com/danielgtaylor/python-betterproto/commit/1010b39ceeed79df96d3b45f6d87db601310e61c

Looking forward to the 2.0 release and your feedback on this PR!


.gitignore
```diff
@@ -17,3 +17,4 @@ output
 .venv
 .asv
 venv
+.devcontainer
```
efokschaner, line 20: BTW If you'd like well-setup devcontainer for this project with pyenv python 36 and 310 ready to go, lmk and I can make a separate PR.

docs/migrating.rst
```diff
@@ -155,3 +155,39 @@ Upgrading:
 - Create a new *empty* directory, e.g. ``generated`` or ``lib/generated/proto`` etc.
 - Regenerate your python files into this directory
 - Update import statements, e.g. ``import ExampleMessage from generated``
+
+
+[2.0.0b4] to [2.x.x]
```
efokschaner, line 160: I wasn't quite sure how to adhere to the format here.

tests/test_features.py
```diff
@@ -480,8 +478,10 @@ class Envelope(betterproto.Message):
     assert all([isinstance(item, datetime) for item in msg.timestamps])
 
 
-def test_enum_service_argument__expected_default_value():
-    from tests.output_betterproto.service.service import ThingType, TestStub
+def test_service_argument__expected_parameter():
```
efokschaner, line 481: I completely changed what this test is about but kept it to a similar kind of test, asserting what we can see via inspect.

pyproject.toml
```diff
@@ -111,7 +111,7 @@ omit = ["betterproto/tests/*"]
 legacy_tox_ini = """
 [tox]
 isolated_build = true
-envlist = py36, py37, py38
+envlist = py36, py37, py38, py310
```
efokschaner, line 114: I noticed you're already running 310 in CI so figured it wouldn't hurt to add to tox, which I then used to test on 3.6 and 3.10

tests/test_features.py
Gobot1234: Why exactly have you removed this?

efokschaner: In my mind I was thinking an rpc method wouldn't take an enum, but now I'm wondering if perhaps enum is allowed as the param in a proto file. Assuming that enums are allowed as the parameter to a proto, do we want to provide default value for enum in the function signature? With this change atm there's no longer any defaults in the RPC method signatures.

After some googling though, I don't think direct enums are a valid rpc parameter they seem more like scalars than messages.

Lmk if you know the answer to this, or I'll test it out myself.

Gobot1234: No you're completely correct my bad

docs/migrating.rst
cetanu: Hmm... It's a bit of a strange one since it's from one beta release to another.

Personally I would include it in the changelog but I'd like to know what others think.

cetanu:
I'm happy to approve this but would like other reviewers to weigh in

docs/migrating.rst
Gobot1234: Yeah I'm inclined to agree here

efokschaner: Are you saying move this entire migrating comment in to the changelog, or just delete it? Atm, I put a short line item in the changelog as well.

Gobot1234: I'd suggest just moving most of this to the change log

efokschaner: https://github.com/danielgtaylor/python-betterproto/pull/311/commits/08069babf40d42651fb6d8bf005acb9c2e9923ec moves the migration notes to the changelog

cetanu (approved):

Gobot1234 (approved):

FHTMitchell:
This is great, thanks for doing this!

efokschaner:
@kalzoo @nat-n @boukeversteegh shall we await your review or would you be happy to defer to @cetanu and @Gobot1234 on this change?

boukeversteegh (approved):
Hi @efokschaner !

Looks like a great improvement, as it makes the generator code more simple, and the usage of the clients more straightforward when you already have messages that can be requests. Type checking will also work better in this way.

As I've indicated in some other issues, I'm not currently working on betterproto, and don't have time to review PR's in depth, so I haven't checked out the code or tested it. The implementation looks decent though, I don't see anything unusual in the code.

I would be in favor of merging this and quickly releasing a new version. If that breaks things for people, we should here it soon. In that case, I hope you'd be willing to look into it.

Luminaar:
Hello. I'm surprised that this change was merged without much discussion. The original behavior (passing parameters to methods instead of message objects) was actually one of the features that we really liked and made us start using `betterproto` very early on. In my opinion it is a big improvement of the official library (`grpcio`) and our (internal) users agree.

I suppose this is a done deal now, but I just wanted to add my opinion for posterity.

jbkoh:
@danielgtaylor I just noticed this breaking change and was surprised enough to come here in rush. I think [this](https://github.com/danielgtaylor/python-betterproto/issues/287#issuecomment-1262468031) explains pretty well why the previous design is better. I beg you to reconsider this change.

PR #321, opened 2022-01-18
Ensure any fields that could intersect with Message are renamed

Gobot1234:
Follows from a conversation with @danielgtaylor 

parse -> parse_
serialized_on_wire -> serialized_on_wire_
foo -> foo

This allows us to reserve names for betterproto. I'll need to document this somewhere.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -609,7 +609,12 @@ class Message(ABC):
             Calls :meth:`__bool__`.
     """
 
-    _serialized_on_wire: bool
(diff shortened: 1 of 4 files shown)


cetanu:
Which part of this is breaking? The change overall seems fine to me.

cetanu (approved):

Gobot1234:
I removed the which_one_of and serialised_on_wire methods

cetanu:
I see... and they are attributes on the messages instead... That seems like good value in my opinion

PR #322, opened 2022-01-18
Split up betterproto into separate modules

### LINKED ISSUE #49: Split code up into multiple modules

Gobot1234:
This makes implementing some critical parts of the API to be implemented in a c extension somewhere down the road.

I'm happy to fix up any currently open pull requests to get this merged.

I haven't entirely followed https://github.com/danielgtaylor/python-betterproto/issues/49#issuecomment-636305283 as I think it's good to keep everything as together as possible (this also slightly relies on some of my other PRs being merged). 

I'd like to draw particular attention to https://github.com/Gobot1234/python-betterproto/blob/086da0b6651712226b8382ca0c8477e6519230a9/src/betterproto/lib/google/protobuf/__init__.py as currently it requires editing to actually allow it to be importable without using local imports (which I'd really like to avoid). I'd be interested to hear if anyone can think of a way to avoid this.

TODO
- [ ] Run isort
- [ ] Fix any created merge conflicts

Closes #49


Gobot1234:
I'll remake this one day when it stands more of a chance of being merged

PR #328, opened 2022-01-30
Grpcio support

jalaziz:
This is a rebased version of #201. I still haven't done any testing.

The intention is to continue the work started in #201, add tests, and address any remaining issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,9 +15,11 @@ packages = [
 python = ">=3.6.2,<4.0"
 black = { version = ">=19.3b0", optional = true }
 dataclasses = { version = "^0.7", python = ">=3.6, <3.7" }
+grpcio = { version = "^1.43.0", optional = true }
 grpclib = "^0.4.1"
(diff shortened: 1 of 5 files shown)


cetanu:
Hello, thank you for this contribution

If you could address the conflicts, I will give it a review afterwards and look to merge

JPFrancoia:
Hi all, I'm also very interested in this PR. I am experiencing a bug with grpclib. Switching to grpcio (but dropping betterproto to do the codegen :/) solves the problem. I am trying to debug this, but also, grpclib doesn't have the open telemetry support that grpcio has. And grpclib doesn't seem to receive as much love as it should. All this makes grpclib a bit risky in a real prod setting.

Also, I really love the simplicity of betterproto and the code it generates. Getting something usable without it is a bit painful and feels brittle, and the output still isn't as nice as betterproto.

That's why I think having the choice for a backend would be a big plus. I'm happy to contribute (code and/or test) if help is required to get this PR merged.

PR #329, opened 2022-01-31
Fix from_dict() in the presence of optional datetime fields.

### LINKED ISSUE #333: from_dict() raises TypeError in the presence of proto3 optional google.protobuf.Timestamp field
```
>                       setattr(self, field_name, cls().from_dict(value[key]))
E                       TypeError: function missing required argument 'year' (pos 1)

src/betterproto/__init__.py:1204: TypeError
```

emosenkis:
Fixes #333

The added test fails without the fix to `__init__.py`:

```
>                       setattr(self, field_name, cls().from_dict(value[key]))
E                       TypeError: function missing required argument 'year' (pos 1)

src/betterproto/__init__.py:1204: TypeError
```


src/betterproto/__init__.py
```diff
@@ -1191,16 +1191,15 @@ def from_dict(self: T, value: Dict[str, Any]) -> T:
                             ]
                         else:
                             v = [cls().from_dict(item) for item in value[key]]
-                    elif isinstance(v, datetime):
+                    elif cls == datetime:
```
Gobot1234, line 1194: Can you change these to if cls is datetime?

emosenkis: For whatever reason, there are a lot of instances of `cls == foo` in this file and none of `cls is foo`. I think it makes sense to maintain consistency but I'm happy to change all of them to use `is` if you prefer.

emosenkis: Done

kalzoo: I agree with both of you, and it was courteous of you to match the existing style, @emosenkis . I'd say given that it's a style point, leave as-is and fix on a refactor.

kalzoo (approved):
Thanks @emosenkis for the test and fix!

Before we merge - please create an issue describing the problem (basically the same as the MR description) and link this MR to it. That will help other people discover both the bug and the fix when they run into it.

emosenkis:
Thanks! Could you push to pypi?

PR #335, opened 2022-02-07
generate into separate files, do not try to understand ast

### LINKED ISSUE #300: overriding files with incorrect content -> package name should not be file name
I feel that I use the plugin in a wrong way or there is something not designed correctly.

example:

I have directory with three files

status.proto 

```
message Success {
}

```
and
package svcs
servicesA.proto

```
package svcs
import "status.proto"
service ServiceA {
... rpcs
}
```
service.B.proto

```
package svcs
import "status.proto"
service ServiceB {
... rpcs
}
```
running protoc 3 times (serviceA.proto, serviceB.proto status.proto) in that order:
1. svcs.py with content of A
2. svcs.py with content of B
3. svcs.py with content of status.proto

I have been using protobuf for more than 12 years, but this is not correct. It is normal to have one package with multiple file being included around and similar things. the generator must either append or somehow deal with it. It is not correct expectation that each file will have a different namespace

morapet:
fixes #300

### DIFF
--- a/betterproto/plugin.py
+++ b/betterproto/plugin.py
@@ -44,6 +44,8 @@ WRAPPER_TYPES: Dict[str, Optional[Type]] = defaultdict(
     },
 )
 
+def _get_proto(request, name: str):
+    return next(f for f in request.proto_file if f.name == name)
 
 def get_ref_type(
     package: str, imports: set, type_name: str, unwrap: bool = True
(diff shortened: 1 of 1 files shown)


cetanu (changes requested):
If you could please remove some of the print statements, or convert them into debug logs, that would be good

This will also re-trigger the build checks

Gobot1234:
This PR isn't targeting the right branch. I don't think this PR is worth pursuing

cetanu:
Alright, will close, if it's still desired please open a new PR that targets master

Issue #323, opened 2022-01-20
Add ParseFromString for compatibility


Laurens-T:
I currently use the standard protobuf method `ParseFromString` quite a bit, but noticed that the generated `betterproto` dataclasses don't have this method, but they do have `SerializeToString` for compatibility.

Is there a reason `ParseFromString` isn't added?

```protobuf
syntax = "proto3";

package hello;

message Hello {
  string message = 1;
}
```
```python
from better_proto.hello_pb2 import Hello

hello = Hello()
hello.ParseFromString(bytes(Hello))
```

Gobot1234:
There isn't any reason I know of

simon-saliba:
This is a problem I've encountered before and I believe that the compatibility with the standard protobuf implementation has merit to being supported.
I opened a PR for this.

PR #336, opened 2022-02-12
feat: added ParseFromString method to message for compatibility 

simon-saliba:
This PR adds the ParseFromString attribute to the message class for compatibility with the official Google protobuf implementation and method exposure.
Related Issue #323


Laurens-T:
Thanks!

cetanu (approved):

src/betterproto/__init__.py
```diff
@@ -1022,6 +1022,29 @@ def FromString(cls: Type[T], data: bytes) -> T:
...
+        -----------
+        data: :class:`bytes`
+            The data to parse the protobuf from.
+
+        Returns
+        --------
+        :class:`Message`
+            The initialized message.
+        """
+        return self.parse(data)
```
Gobot1234, line 1046: Isn't this slightly different from the offical implementation which clears any currently set values? https://github.com/protocolbuffers/protobuf/blob/2a2a9b6e64e0cda49b0e5377cad58c790087e1c9/python/google/protobuf/message.py#L193-L199

Gobot1234 (changes requested):

src/betterproto/__init__.py
Laurens-T: This indeed seems to be the case.

I've added a simple example which shows the difference in result from `ParseFromString` with the one from this PR:

```protobuf
syntax = "proto3";

package some.package.name;

message Cat {
  int64 age = 1;
  int64 lives = 2;
}
```

```python
def ParseFromString(self, data: bytes):
    return self.parse(data)

if __name__ == '__main__':
    # Create bytes
    cat = Cat(
        age=5,
        # lives=9,
    )

    data = cat.SerializeToString()

    # Dynamically add ParseFromString to the Cat class
    Cat.ParseFromString = ParseFromString

    # ParseFromString for the betterproto cat and the pb2 cat
    cat_betterproto = Cat(age=3, lives=8)
    cat_pb2 = Cat_pb2(age=3, lives=8)

    cat_betterproto.ParseFromString(data)
    cat_pb2.ParseFromString(data)

    print(cat_betterproto)  # Cat(age=5, lives=8), lives are not cleared
    print(cat_pb2)  # age: 5, lives are cleared
    assert cat_pb2.lives == 0
```

PR #337, opened 2022-02-15
Stop overwriting `__init__.py`s

### LINKED ISSUE #168: Package __init__ files overwritten at compile
Hi,
Great library!

The generated `__init__.py` files that create packages are written (and empty) even if those were already present.
This creates a conflict with normally-non-empty `__init__` files, e.g. importing from subpackages.
For instance with a layout like this:

```
/
/src                                  # C code also using protobufs
/src/protobuf                  # protobuf definitions, shared between C and python
/src_py                            # python package root
/src_py/driver
/src_py/driver/__init__.py
/src_py/driver/driver.py
/src_py/driver/protobuf/ # compiled protobufs go here
```

and a generation command being:
`python -m grpc_tools.protoc -I ./src --python_betterproto_out=src_py/driver ./src/protobuf/*.proto
`
the file `./src_py/driver/__init__.py` gets overwritten

Perhaps the `__init__.py` generation could check if the file is empty?

### LINKED ISSUE #260: generating python should not override __init__.py files
Better proto creates an empty __init__.py file above the package level, however if an __init__.py file exists, it should not override the existing file.

Gobot1234:
Fixes #260
Fixes #168
@nat-n mentioned in #168 that there could be "a legit use case if content has been removed from the proto since the previous version." but I don't think that matters cause v2 can have breaking changes


cetanu (approved):

src/betterproto/plugin/parser.py
```diff
@@ -124,13 +124,12 @@ def generate_code(request: CodeGeneratorRequest) -> CodeGeneratorResponse:
 
     # Make each output directory a package with __init__ file
     init_files = {
-        directory.joinpath("__init__.py")
-        for path in output_paths
-        for directory in path.parents
+        directory / "__init__.py" for path in output_paths for directory in path.parents
```
abn, line 127:
```suggestion
        directory / "__init__.py"
        for path in output_paths for directory in path.parents
        not directory.joinpath( "__init__.py").exists()
```

This might be better that way we also only add "Writing  " messages for the ones we actually write.

src/betterproto/plugin/parser.py
```diff
@@ -124,9 +124,10 @@ def generate_code(request: CodeGeneratorRequest) -> CodeGeneratorResponse:
 
     # Make each output directory a package with __init__ file
     init_files = {
-        directory.joinpath("__init__.py")
+        directory / "__init__.py"
```
abn, line 127:
```suggestion
        directory.joinpath("__init__.py")
```

Nitpicky, but might be better to leave this as it was to keep the change minimal

Gobot1234: That's probably a good idea

abn (approved):

PR #338, opened 2022-02-15
Remove the poetry.lock

Gobot1234:
Poetry says that commiting this is not neccessary for libraries it seems to be behind a lot of merge conflicts


abn:
While this is not necessary for libraries, without it, the development experience will become rather problematic since every install will require a re-resolution of the entire dependency tree.

Merge conflicts typically happen if every one keeps updating the lock file (or sometimes due to dependabot doing lockfile surgery), which should not happen during normal cases, ie. lock file should only change if there is a change in dependencies.

PR #346, opened 2022-03-01
configure pre-commit for project

abn:
This change moves black and doc8 checking into pre-commit hooks. This also simplifies the ci code quality workflow. In addition, doc8 configuration is now moved to `pyproject.toml`.

This is more a proposal to make development experience a bit easier and standardised. It helps to not have to remember to execute all the required checks before pushing up code. Additionally, having a standard config mechanism helps the CI job to be simpler. This can potentially be outsourced to [pre-commit.ci](https://pre-commit.ci/) as well to keep  the hooks up-to-date.

**peo: fix black task to include gitignore**

When `--exclude` is used, the default exclude list is overwritten. The expected option is `--extend-exclude`. However, since the generated files are already ignored, these flags can be safely skipped,


abn:
Not really sure why the CI is failing on generation as none of that code has been touched here.

abn:
The CI failures were caused as `black` was removed from development dependency and `poetry` does not activate extras by default.

abn:
The new failures should be resolved via #345.

abn:
@Gobot1234 would be great to get this one in too; should speed up CI by ~2.5 minutes based on the current active runs.

Gobot1234 (approved):

Issue #347, opened 2022-03-02
Lazy initialisation of default values can cause unexpected behaviour


abn:
I have the following code genereated from the openconfig/gnmi proto files.

```py
@dataclass(eq=False, repr=False)
class Update(betterproto.Message):
    """
    Update is a re-usable message that is used to store a particular Path,
    Value pair. Reference: gNMI Specification Section 2.1
    """

    path: "Path" = betterproto.message_field(1)
    value: "Value" = betterproto.message_field(2)
    val: "TypedValue" = betterproto.message_field(3)
    duplicates: int = betterproto.uint32_field(4)

    def __post_init__(self) -> None:
        super().__post_init__()
        if self.value:
            warnings.warn("Update.value is deprecated", DeprecationWarning)
```

Field `value` is deprecated in favour of `val`.  See relevant content from proto files.

```proto
message Update {
  Path path = 1;                      // The path (key) for the update.
  Value value = 2 [deprecated=true];  // The value (value) for the update.
  TypedValue val = 3;                 // The explicitly typed update value.
  uint32 duplicates = 4;              // Number of coalesced duplicates.
}

message Value {
  option deprecated = true;
  bytes value = 1;      // Value of the variable being transmitted.
  Encoding type = 2;    // Encoding used for the value field.
}
```

Due to the changes introduced in #130, specifically in `Message.__getattribute__`, when the deprecation logic checks value (`Update.value`), this value gets set via this logic.

https://github.com/danielgtaylor/python-betterproto/blob/74205e3319f1bf1c42f6789bfe44ca07a07343a1/src/betterproto/__init__.py#L684-L695 

The side-effects of this are the following:
1. The deprecation warning for `Update.value` is not raised.
2. A new instance (default value) of `Value()` is initialised. This raises a deprecation warning for the message itself. This happens in cases where this is not set explicitly.
3. In cases where implementations check both `Update.value` and `Update.val`, we could end up with unexpected behaviour as the values contained in these might differ as one is default and the other is not.

**Expected Behaviour**
- Access to deprecated field when not set should be `None`.
- No deprecation warnings are raised when fields are not explicitly set.

abn:
cc: @Gobot1234 @nat-n @chris-chambers as this stems from the code introduced in #130 to handle recursive message types as far as I can tell.

abn:
[...]

PR #348, opened 2022-03-03
fix incorrect deprecation warnings on defaults

abn:
This change ensures that deprecation warnings are only raised when
either a deprecated field is explicitly set or a deprecated message is
initialised.

Resolves: #347


Gobot1234 (approved):
LGTM thanks

abn:
Can we get this merged please? :pray:

PR #349, opened 2022-03-03
tests.generate: remove deprecated get_event_loop

abn:
The use of `asyncio.get_event_loop()` has been deprecated in python
3.10+. We replace this usage with `asyncio.run()` for python 3.7+.

**References:**
- https://bugs.python.org/issue38599
- https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop

### DIFF
--- a/tests/generate.py
+++ b/tests/generate.py
@@ -161,7 +161,11 @@ def main():
     if platform.system() == "Windows":
         asyncio.set_event_loop(asyncio.ProactorEventLoop())
 
-    asyncio.get_event_loop().run_until_complete(generate(whitelist, verbose))
+    try:
+        asyncio.run(generate(whitelist, verbose))
+    except AttributeError:
+        # compatibility code for python < 3.7
+        asyncio.get_event_loop().run_until_complete(generate(whitelist, verbose))
 
 
 if __name__ == "__main__":
--- a/tests/generate.py
+++ b/tests/generate.py
@@ -159,7 +159,13 @@ def main():
         whitelist = set(sys.argv[1:])
 
     if platform.system() == "Windows":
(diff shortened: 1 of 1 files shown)


Gobot1234 (dismissed):

tests/generate.py
```diff
@@ -159,9 +159,13 @@ def main():
         whitelist = set(sys.argv[1:])
 
     if platform.system() == "Windows":
-        asyncio.set_event_loop(asyncio.ProactorEventLoop())
+        asyncio.get_event_loop_policy().set_event_loop(asyncio.ProactorEventLoop())
```
Gobot1234, line 162: Isn't manually instanciating an event loop discouraged? 
```suggestion
        asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
```

abn: Ah that is a better fix.

Gobot1234 (approved):

abn:
Hmm guess this is why the loop was being set explicitly.

```console
AttributeError: module 'asyncio' has no attribute 'WindowsProactorEventLoopPolicy'
```

Gobot1234:
> Hmm guess this is why the loop was being set explicitly. For python 3.6 on windows.
> 
> ```
> AttributeError: module 'asyncio' has no attribute 'WindowsProactorEventLoopPolicy'
> ```

Maybe this is just a sign that we should drop 3.6, that is ok then if that doesn't work

abn:
That seems to have done the trick. Removing 3.6 support might be required if more compatibility code starts showing up in the codebase.

Gobot1234 (approved):

PR #354, opened 2022-03-12
pre-commit: add isort hook and apply

abn:
Split from #352 as requested by @Gobot1234 

~~Need resolution on https://github.com/danielgtaylor/python-betterproto/pull/352#discussion_r825172934~~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,17 +101,14 @@ cmd = "black . --check --diff"
 help = "Check if code style is correct"
 
 [tool.isort]
-py_version = 36
+py_version = 37
 profile = "black"
 force_single_line = false
 combine_as_imports = true
-lines_between_types = 1
 lines_after_imports = 2
+include_trailing_comma = true
+force_grid_wrap = 2
 src_paths = ["src", "tests"]
-extend_skip = [
-    "^tests/output_betterproto",
-    "^tests/output_reference",
-]
 
 [tool.black]
 target-version = ['py36']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,18 @@ help = "Run tests with multiple pythons"
 cmd = "black . --check --diff"
 help = "Check if code style is correct"
 
+[tool.isort]
+py_version = 36
+profile = "black"
+force_single_line = false
+combine_as_imports = true
+lines_between_types = 1
+lines_after_imports = 2
+src_paths = ["src", "tests"]
+extend_skip = [
+    "^tests/output_betterproto",
+    "^tests/output_reference",
+]
 
 [tool.black]
 target-version = ['py36']
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -7,9 +7,16 @@ import sys
 import typing
 import warnings
 from abc import ABC
-from base64 import b64decode, b64encode
+from base64 import (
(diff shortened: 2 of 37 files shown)


pyproject.toml
```diff
@@ -100,6 +100,18 @@ help = "Run tests with multiple pythons"
 cmd = "black . --check --diff"
 help = "Check if code style is correct"
 
+[tool.isort]
+py_version = 36
+profile = "black"
+force_single_line = true
```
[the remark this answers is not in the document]
Gobot1234: I've personally never seen conflicts on imports especially cause black will put them on their own separate "not-conflict-able" line if it's long enough. I think a potential solution could just be to make all imports end in a trailing comma and let black handle it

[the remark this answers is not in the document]
abn: I think that's unnecessary effort. That said, for the purposes of this PR, I'll remove that config and leave the current status quo so that we can get basic sorting in.

src/betterproto/__init__.py
```diff
@@ -25,13 +25,16 @@
     get_type_hints,
 )
 
+import dataclasses
+
```
Gobot1234, line 29: Why was dataclasses moved down to here?

src/betterproto/lib/google/protobuf/__init__.py
```diff
@@ -2,10 +2,13 @@
...
 import warnings
-from dataclasses import dataclass
+
 from typing import Dict, List
 
+from dataclasses import dataclass
+
 import betterproto
+
 from betterproto.grpc.grpclib_server import ServiceBase
```
Gobot1234, line 12: Which setting is putting from and normal imports on separate lines, this looks weird?

src/betterproto/lib/google/protobuf/compiler/__init__.py
```diff
@@ -1,10 +1,12 @@
...
 # sources: google/protobuf/compiler/plugin.proto
 # plugin: python-betterproto
-from dataclasses import dataclass
 from typing import List
 
+from dataclasses import dataclass
+
 import betterproto
+
 from betterproto.grpc.grpclib_server import ServiceBase
```
Gobot1234, line 10: Same comment here

src/betterproto/__init__.py
abn: I believe it's because of py36 compat dataclass is considered as third party.

src/betterproto/lib/google/protobuf/__init__.py
abn: Iirc that's pep8 recommendation. 

From the pep

```py
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```

Gobot1234: Sorry I meant with the new line inbetween

abn: `lines_between_types`. This just looks weird here becomes i think its a single import and then `from` import from the same package.

src/betterproto/__init__.py
abn: switched the python version to 37 for now to resolve that.

Gobot1234:
LGTM

Gobot1234 (approved):

PR #361, opened 2022-03-28
Fix is_set for optional proto3 fields

Gobot1234:
It just occured to me that this was broken, realistically however this is easy enough for the end user to perform.

### DIFF
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -517,3 +517,13 @@ def test_copyability():
     assert spam == deepcopied
     assert spam is not deepcopied
     assert spam.baz is not deepcopied.baz
+
+
+def test_is_set():
+    @dataclass
+    class Spam(betterproto.Message):
+        foo: bool = betterproto.bool_field(1)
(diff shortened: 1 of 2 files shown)


cetanu (dismissed):

abn (commented):
Should this add a test case to prevent regression?

tests/test_features.py
```diff
@@ -517,3 +517,14 @@ class Spam(betterproto.Message):
...
+
+def test_is_set():
+    @dataclass
+    class Spam(betterproto.Message):
+        foo: bool = betterproto.bool_field(1)
+        bar: Optional[int] = betterproto.int32_field(2, optional=True)
+
+    assert not Spam().is_set("foo")
+    assert not Spam().is_set("bar")
+    assert Spam(foo=True).is_set("foo")
```
abn, line 530:
```suggestion
    assert Spam(foo=True).is_set("foo")
    assert Spam(foo=0).is_set("bar")
```

To round the cases off.

abn (approved):

PR #365, opened 2022-04-11
Fix comparison with nan

Gobot1234:
Also fixed some typos in the docstrings

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -467,7 +467,7 @@ def _dump_float(value: float) -> Union[float, str]:
         return INFINITY
     if value == -float("inf"):
         return NEG_INFINITY
-    if math.isnan(value):
+    if isinstance(value, float) and math.isnan(value):
         return NAN
     return value
 
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -433,7 +433,7 @@ def _parse_float(value: Any) -> float:
 
     Parameters
     ----------
-    value : Any
+    value: Any
         Value to parse
 
     Returns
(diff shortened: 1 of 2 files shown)


cetanu (dismissed):

src/betterproto/__init__.py
```diff
@@ -455,20 +455,19 @@ def _dump_float(value: float) -> Union[float, str]:
...
-        Dumped valid, either a float or the strings
-        "Infinity" or "-Infinity"
+        Dumped value, either a float or the strings
     """
     if value == float("inf"):
         return INFINITY
     if value == -float("inf"):
         return NEG_INFINITY
-    if value == float("nan"):
+    if math.isnan(value):
```
abn, line 470:
```suggestion
    if isinstance(value, float) and math.isnan(value):
```

I am not sure why the test case ended up passing a `None` value in the first place when this method does not expect it.

abn (changes requested):

cetanu (approved):

src/betterproto/__init__.py
```diff
@@ -455,20 +455,19 @@ def _dump_float(value: float) -> Union[float, str]:
...
-        Dumped valid, either a float or the strings
-        "Infinity" or "-Infinity"
+        Dumped value, either a float or the strings
     """
     if value == float("inf"):
         return INFINITY
     if value == -float("inf"):
         return NEG_INFINITY
-    if value == float("nan"):
+    if isinstance(value, float) and math.isnan(value):
```
abn, line 470: @Gobot1234 should we skip the other checks as well if the instance is not a float? Seems pointless if it is not a float.

Gobot1234: I think a guard against None in to/from_dict would be the best solution

abn (approved):

src/betterproto/__init__.py
Gobot1234: I'll do this in #367 just cause I want to get this merged

PR #370, opened 2022-04-18
Update README.md

asfaltboy:
The change is to add a linebreak after bullet-point, and remove redundant conjunction at the start of the section

_This is really a minor nitpick, clearly, but I personally found the existing style a bit confusing when reading the README_


cetanu (approved):
Seems innocuous enough

abn (approved):

PR #378, opened 2022-04-23
[fix] to_dict modifies the underlying message (#151)

### LINKED ISSUE #151: to_dict modifies the underlying message
Hi,

It seems that the `to_dict` method modifies the nested message.

In short, here's the code and the output (version from master, pulled 24-09-2020)

```python
from typing import Dict

import betterproto


class Code(betterproto.Enum):
    ZERO = 0
    ONE = 1

class Message(betterproto.Message):
    codes: Dict[str, "Nested"] = betterproto.map_field(2, betterproto.TYPE_STRING, betterproto.TYPE_MESSAGE)

        super().__post_init__()


class Nested(betterproto.Message):
    code: "Code" = betterproto.enum_field(1)

        super().__post_init__()

msg = Message(codes={'test': Nested(code=Code.ONE)})
print(msg.codes['test'])
print(msg.to_dict())
print(msg.codes['test'])
```

```
Nested(code=<Code.ONE: 1>)
{'codes': {'test': {'code': 'ONE'}}}
{'code': 'ONE'}
```

Also, if you'll change `ONE` to `ZERO`, it won't print anything, as `ZERO` is a default value (even though it was set manually)

GrownNed:
Fixes #151

### DIFF
--- a/tests/test_mapmessage.py
+++ b/tests/test_mapmessage.py
@@ -4,7 +4,7 @@ from tests.output_betterproto.mapmessage import (
 )
 
 
-def test_mapmessage():
+def test_mapmessage_to_dict_preserves_message():
     message = Test(
         items={
             "test": Nested(
--- a/tests/test_mapmessage.py
+++ b/tests/test_mapmessage.py
@@ -1,7 +1,8 @@
 from tests.output_betterproto.mapmessage import (
-    Test,
     Nested,
+    Test,
 )
+
 
 def test_mapmessage():
     message = Test(
--- a/tests/test_mapmessage.py
+++ b/tests/test_mapmessage.py
@@ -13,6 +13,6 @@ def test_mapmessage_to_dict_preserves_message():
         }
     )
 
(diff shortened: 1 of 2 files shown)


Gobot1234:
Would you mind adding a test for this?

Gobot1234:
I'm not entirely sure what the issue on the tests is but I'd guess the line before the assignment to output should check if output_map (I haven't tested this)

tests/test_mapmessage.py
```diff
@@ -0,0 +1,18 @@
+from tests.output_betterproto.mapmessage import (
+    Nested,
+    Test,
+)
+
+
+def test_mapmessage():
```
abn, line 7:
```suggestion
def test_mapmessage_to_dict_preserves_message():
```

Nitpick, but something like that might be good.

Gobot1234 (approved):

Gobot1234:
Thanks!

PR #380, opened 2022-04-24
Lazy evaluate Deadline parameter in pytest

efokschaner:
This test was failing with a Deadine error for me locally and is solved by the following.

The pytest parameters are evaluated when the tests are loading.
The Deadline.from_timeout is a fixed point in time.
By deferring the evaluation it helps ensure that the deadline
is not reached before the test is executed.


Gobot1234:
This should probably help but this issue should have been fixed upstream https://github.com/vmagamedov/grpclib/issues/156#issuecomment-1106430144

Gobot1234 (approved):

Gobot1234:
Thanks!

PR #382, opened 2022-05-09
Added support for @generated marker

anriha:
This PR adds support for @generated marker as specified at https://generated.at/ This marks generated code as such and allows it to be skipped in CI/CD. I am just not sure if this is the right format or if adding it after current message would be better.


Gobot1234:
I think sticking to the same format as protoc might be a good idea, the `# @generated` part doesn't have to be on the first line does it? Could it be moved to below the plugin line?

Gobot1234 (dismissed):

anriha:
Yes you are probably right, it seems that it just has to be somewhere at the beginning (not clearly defined). But from examples I found it seems that it can be below.

Gobot1234 (dismissed):

Gobot1234:
One more idea would be to regenerate the pre-compiled files in lib/google/protobuf

anriha:
I regenerated all the lib/google/protobuf files. Looking at the output it seems like the generated files had wrong documentation for attributes (before attribute definition, not after). So this also fixes that.

Gobot1234:
Thanks, this looks great

Gobot1234 (approved):

PR #397, opened 2022-07-01
check if value is iterable that could support on both list and tuple

esunvoteb:
When serializing the message, it would throw the below exception when processing the "Tuple" type value.
```
  File "D:\dev\woven\compusher\venv\lib\site-packages\betterproto\__init__.py", line 840, in SerializeToString  
    return bytes(self)
  File "D:\dev\woven\compusher\venv\lib\site-packages\betterproto\__init__.py", line 815, in __bytes__
    output += _serialize_single(
  File "D:\dev\woven\compusher\venv\lib\site-packages\betterproto\__init__.py", line 396, in _serialize_single  
    value = _preprocess_single(proto_type, wraps, value)
  File "D:\dev\woven\compusher\venv\lib\site-packages\betterproto\__init__.py", line 364, in _preprocess_single 
    return value.encode("utf-8")
AttributeError: 'tuple' object has no attribute 'encode'
```
My PR involves very simple changes to support more generic iterable types which include both "List" and "Tuple".

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -11,6 +11,7 @@ from base64 import (
     b64decode,
     b64encode,
 )
+from collections.abc import Iterable
 from copy import deepcopy
 from datetime import (
     datetime,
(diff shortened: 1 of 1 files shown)


Gobot1234:
I don't think this is a good idea cause it encourages breaking type safety, assigning a field that is supposed to be a list to an Iterable is not a good idea. You should be getting a type checker error here that should be fixable by simply casting the tuple to a list yourself

Issue #401, opened 2022-07-15
betterproto v2.0.0b4


Liu-Weijie:
When I used the command as follow:
protoc -I . --python_betterproto_out=lib e2sm_rc.proto 
I got the error as follow:
--python_betterproto_out: protoc-gen-python_betterproto: Plugin failed with status code 1.
My version of the protobuf compiler is 3.15
I don't know where it goes wrong?

Gobot1234:
I can't tell from your issue, try using the latest version from GitHub.

chitalian:
I am also hitting the same issue! @Liu-Weijie Can you please elaborate what you mean by `I found the version of the jinja2 is out of date, so I update it manually and it can work normally now`?

I ran 
`pip install jinja2 -U` and reran the command
```bash
python3 -m grpc_tools.protoc -I . --python_betterproto_out=. api.proto
--python_betterproto_out: protoc-gen-python_betterproto: Plugin failed with status code 1.
```

With the same error

aydarmynbay:
@chitalian try importing jinja2 in some other script. Don't know how, but it worked for me

Liu-Weijie:
I am sorry that I just see the message.
you should pip uninstall jinja2 and pip install --upgrade jinja2
you can try it.

chitalian:
@Gobot1234 
```protobuf
syntax = "proto3";

package sisu.v1.api;

message AnalysesResultRequest {
  int64 limit = 1;
  int64 starting_after = 2;
  int64 csv = 3;
  int64 top_drivers = 4;
}

```

chitalian:
I just tried in a brand new docker container and it failed the same way

Gobot1234:
What does `pip freeze` look like?

chitalian:
```python
autopep8==1.6.0
betterproto==2.0.0b4
bleach==5.0.1
build==0.8.0
certifi==2022.6.15
charset-normalizer==2.1.0
commonmark==0.9.1
docutils==0.19
googleapis-common-protos==1.56.4
grpcio==1.47.0
grpcio-tools==1.47.0
grpclib==0.4.3rc3
h2==4.1.0
hpack==4.0.0
hyperframe==6.0.1
idna==3.3
importlib-metadata==4.12.0
Jinja2==3.1.2
keyring==23.6.0
MarkupSafe==2.1.1
multidict==6.0.2
packaging==21.3
pep517==0.12.0
pkginfo==1.8.3
[...]

PR #402, opened 2022-07-19
Update Jinja 2 version

chitalian:
Update jinja version.
see: https://github.com/danielgtaylor/python-betterproto/issues/401

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -22,7 +22,7 @@ hg = ["python-hglib (>=1.5)"]
 
 [[package]]
 name = "atomicwrites"
-version = "1.4.0"
+version = "1.4.1"
 description = "Atomic file writes."
 category = "dev"
 optional = false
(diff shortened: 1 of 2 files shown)


Gobot1234:
You need to update the poetry.lock.

chitalian:
@Gobot1234 

> You need to update the poetry.lock.

When I try to update it I get
```
For jinja2, a possible solution would be to set the `python` property to ">=3.7,<4.0"
```

Is it okay to change python version min to be 3.7?

Gobot1234:
> @Gobot1234 
> 
> 
> 
> > You need to update the poetry.lock.
> 
> 
> 
> When I try to update it I get
> 
> ```
> 
> For jinja2, a possible solution would be to set the `python` property to ">=3.7,<4.0"
> 
> ```
> 
> 
> 
> Is it okay to change python version min to be 3.7?
> 
> 
> 
> 

As a last resort maybe but can you see if there's a version that still supports 3.6?

Gobot1234 (approved):

Gobot1234:
Thanks

PR #405, opened 2022-08-03
Pull down the `include_default_values` argument to `to_json`

SamuelYvon:
The `to_json` method does not include the `include_default_values`
option from `to_dict`. While the implementation to `to_json` is
almost litteraly only a call to `to_dict()` (and thus trivial to
replace), it is more convenient for users to not have to reimplement it
when the option is desired.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1260,7 +1260,9 @@ class Message(ABC):
                     setattr(self, field_name, v)
         return self
 
-    def to_json(self, indent: Union[None, int, str] = None) -> str:
+    def to_json(
+        self, indent: Union[None, int, str] = None, include_default_values: bool = False
+    ) -> str:
         """A helper function to parse the message instance into its JSON
         representation.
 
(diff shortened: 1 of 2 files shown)


Gobot1234:
Might also be a good idea to add support for the casing arg, not too fussed about it though. Thanks for the PR.

SamuelYvon:
Added

Gobot1234:
Would you mind also adding a couple of tests for this (1 or 2 should be enough). The newest version on pypi was released yesterday (I hope it's not too buggy) you can install it now but you need to use the `--pre` flag

SamuelYvon:
> (I hope it's not too buggy)

I was using 1.2.5 I believe, which was not working properly w.r.t the `oneof`. I installed the pre-release and it works fine.

SamuelYvon:
I'll add tests when I get a moment today

SamuelYvon:
@Gobot1234 You got the tests :)

Gobot1234:
Thanks!

Gobot1234 (approved):

Issue #427, opened 2022-09-27
Service stub method generates without rpc parameter sometimes.


adriangaro:
As the title says, got the following code generated, `generate_geo_condition_sql_request` missing from the method signature. From a quick look in code seems that sometimes the message does not exist in `ServiceMethodCompiler.request.all_messages`, `py_input_message` returning none and thus the jinja template not writing it.

```python
class SqlStub(betterproto.ServiceStub):
    async def geo_conditions(
        self,
        *,
        timeout: Optional[float] = None,
        deadline: Optional["Deadline"] = None,
        metadata: Optional["MetadataLike"] = None
    ) -> "GenerateSqlResponse":
        return await self._unary_unary(
            "/csm.shaihulud.sql.SQL/geo_conditions",
            generate_geo_condition_sql_request,
            GenerateSqlResponse,
            timeout=timeout,
            deadline=deadline,
            metadata=metadata,
        )
```
  
Proto:
  
```proto
message GenerateGeoConditionSQLRequest {
  csm.common.RequestHeader header = 1;
  GenerateGeoConditionSQL payload = 2;
}
...
  
service SQL {
  rpc geo_conditions(GenerateGeoConditionSQLRequest) returns (GenerateSQLResponse);
  ...
}
```

Other grpc services work fine so am not sure what is causing in this case. Betterproto version 2.0.0beta5

AlexanderZender:
Can subscribe to this, usually the first method in the stub is missing something

paul-HDD:
Same here !

AlexanderZender:
I found the issue or solution(?). @paul-HDD  @adriangaro 
When entering message class names or services and using capital letters 2 or 3 in a row this happens
change 
GenerateGeoConditionSQLRequest -> GenerateGeoConditionSqlRequest
GenerateSQLResponse -> GenerateSqlResponse

this should fix the missing parameters, never more than 1 capital letter in a row is my personal rule now

paul-HDD:
Thanks @AlexanderZender ! Confirmed this work locally for me.
Sounds like an issue to me as I cannot change the proto file 😓

hf-kklein:
> When entering message class names or services and using capital letters 2 or 3 in a row this happens
change

Can confirm. The problem is the casing. For a MWE see #438

wpdonders:
Will be fixed by #436

PR #436, opened 2022-10-24
Pythonize input_type name in py_input_message

### LINKED ISSUE #438: betterproto-2.0.0b5 generated methods do not contain message instance as positional argument if message type name is not in UpperPascal case
### Install betterproto-2.0.0b5 + tooling:

```bash
pip install --pre betterproto[compiler]
pip install grpcio-tools
```

other versions:
> grpcio==1.50.0
grpcio-tools==1.50.0
grpclib==0.4.3

### Create an `example.proto`

```protobuf
syntax = "proto3";

package echo;

message EchoRequest {
  string value = 1;
  // Number of extra times to echo
  uint32 extra_times = 2;
}

message EchoResponse {
  repeated string values = 1;
}

message EchoStreamResponse  {
  string value = 1;
}

service Echo {
  rpc Echo(EchoRequest) returns (EchoResponse);
  rpc EchoStream(EchoRequest) returns (stream EchoStreamResponse);
}
```
### Create the directory
```bash
mkdir lib
```

### Generate Python Code:
```python
python -m grpc_tools.protoc -I . --python_betterproto_out=lib example.proto
```
Check that `lib/echo/__init__.py` contains a class `EchoStub` with a method that has `echo_request` as a positional argument ✔
```python
class EchoStub(betterproto.ServiceStub):
    async def echo(
        self,
        echo_request: "EchoRequest",
        *,
        timeout: Optional[float] = None,
        deadline: Optional["Deadline"] = None,
        metadata: Optional["MetadataLike"] = None
    ) -> "EchoResponse":
    ...
```

Up to here, this is just the MWE from `README.md`.

### Change Casing of Request Message Type Name

Now modify the above `example.proto` and change the casing of the request message from `EchoRequest` to `EChoRequest` with an upper case `ECH` at all 3 occurrences. (Note that the bug is _not_ reproducable if you only change to `EChoRequest` with upper case `EC` but lower case `h`.)

```proto
syntax = "proto3";

package echo;

message ECHoRequest {
  string value = 1;
  // Number of extra times to echo
  uint32 extra_times = 2;
}

message EchoResponse {
  repeated string values = 1;
}

message EchoStreamResponse  {
  string value = 1;
}

service Echo {
  rpc Echo(ECHoRequest) returns (EchoResponse);
  rpc EchoStream(ECHoRequest) returns (stream EchoStreamResponse);
}
```

Repeat the next steps up to the code generation (don't forget to clean up lib before re-generating).

The newly generated `lib/echo/__init__.py` now still contains a class `EchoStub`:
```python
class EchoStub(betterproto.ServiceStub):
    async def echo(
        self,
        *,
        timeout: Optional[float] = None,
        deadline: Optional["Deadline"] = None,
        metadata: Optional["MetadataLike"] = None
    ) -> "EchoResponse":
            "/echo.Echo/Echo",
            ec_ho_request, # <--- Bug: ec_ho_request is not provided as an argument
            EchoResponse,
            timeout=timeout,
            deadline=deadline,
            metadata=metadata,
        )
    ...
```
The `ec_ho_request` is undefined (other than `echo_request` in the MWE above) ❌

This leads to the following error at runtime:
> TypeError: EchoStub.echo() takes 1 positional argument but 2 were given

## Expected Behaviour

The code generation shall work for different casing, too.

wouterh:
Fixes https://github.com/danielgtaylor/python-betterproto/issues/427
Fixes #438

### DIFF
--- a/tests/inputs/service_uppercase/test_service.py
+++ b/tests/inputs/service_uppercase/test_service.py
@@ -1,0 +1,7 @@
+import inspect
+from tests.output_betterproto.service_uppercase import TestStub
+
+
+def test_parameters():
+    sig = inspect.signature(TestStub.do_thing)
+    assert len(sig.parameters) == 5, "Expected 5 parameters"
--- a/tests/inputs/service_uppercase/test_service.py
+++ b/tests/inputs/service_uppercase/test_service.py
@@ -1,4 +1,5 @@
 import inspect
+
 from tests.output_betterproto.service_uppercase import TestStub
 
 
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -750,7 +750,7 @@ class ServiceMethodCompiler(ProtoContentBase):
         # comparable with method.input_type
         for msg in self.request.all_messages:
             if (
-                msg.py_name == name.replace(".", "")
+                msg.py_name == pythonize_class_name(name.replace(".", ""))
                 and msg.output_file.package == package
             ):
                 return msg
--- a/tests/inputs/config.py
+++ b/tests/inputs/config.py
@@ -19,6 +19,7 @@ services = {
     "googletypes_service_returns_googletype",
     "example_service",
     "empty_service",
+    "service_uppercase",
 }
 
 
--- a/tests/inputs/service_uppercase/service.proto
+++ b/tests/inputs/service_uppercase/service.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+
+package service_uppercase;
+
+message DoTHINGRequest {
+  string name = 1;
+  repeated string comments = 2;
+}
+
+message DoTHINGResponse {
+  repeated string names = 1;
+}
+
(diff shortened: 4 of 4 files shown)


hf-kklein:
Seems like this would also fix #438

tests/inputs/service_uppercase/test_service.py
```diff
@@ -0,0 +1,7 @@
+import inspect
+from tests.output_betterproto.service_uppercase import TestStub
```
hf-kklein, line 2:
```suggestion

from tests.output_betterproto.service_uppercase import TestStub
```
this fixes the isort failure in the CI. See also https://github.com/wouterh/python-betterproto/pull/1

hf-krechan:
I can approve that this PR will fix the issue #438 
The missing keyword arguments are now generated as expected :)

Gobot1234 (approved):

Gobot1234:
Thank you

PR #444, opened 2022-12-02
Drop support for python3.6

Gobot1234:


MichaelTiemannOSC:
By merging this, it would unblock #451, which would in turn make that issue labeled "good first issue" an actual good first issue.  It's not a great first issue if it's blocked by paradoxical CI/CD rules.

Gobot1234:
As much as I would love to merge it, I don't have permissions to do so

cetanu (approved):
Seems fine, people stuck on 3.6 can use the previous version and it's advisable that they upgrade anyway.

PR #445, opened 2022-12-05
Add Python 3.11 to CI Runs

hf-kklein:


Gobot1234:
I'm pretty sure this won't pass (I tried yesterday) but it's worth a shot

hf-kklein:
Thanks for the fast feedback! Was trying to setup a local poetry env with Python 3.11 and this PR is to check if the problems are on my side or in the code ;)

hf-kklein:
On my local machine the setup fails when installing greenlet. Their release notes say, greenlet >=v2 is recommended for 3.11:
https://greenlet.readthedocs.io/en/latest/changes.html#id4

> Please upgrade to greenlet 2 if you’re using Python 3.11.

Gobot1234:
What requires greenlet, is there a newer version of that project that will work with 3.11?

Gobot1234 (approved):
Thanks

PR #448, opened 2022-12-09
Enforce decimal notation for timedelta JSON representation

### LINKED ISSUE #447: Very small Durations are formatted using scientific notation when converted to dict
I have a field in a message that is a very small timedelta (initialized with `timedelta(microseconds=11)`). When converting this to JSON using the `to_dict` method, I end-up with the value represented like `1.1e-050s`.

The reason is this line https://github.com/danielgtaylor/python-betterproto/blob/master/src/betterproto/__init__.py#L1535, the float is converted to a string using `str`, which defaults to scientific notation for small values. Then, the padding part is adding 0s until there is 3/6/9 chars after the ., which in this case result in the value being incorrectly shifted 48 positions to the right :joy: .

Switching the line to `"{:f}".format(delta.total_seconds()).split(".")` should solve the issue by preventing scientific notation. This seems to be the original intention, and as a bonus it makes interoperability with other parsers easier. For example, I discovered with because the program consuming said message once converted in JSON is written in Go, and the `time.ParseDuration` doesn't understand scientific notation and is considering the `e` as an unit (which it doesn't know about, thus fail.)

Do you want me to submit a PR? I've had a look around the test suite and couldn't find the correct place to add a testcase for this, but I'll be glad to add it too if pointed in the right direction.

elwinar:
Fixes #447

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1532,8 +1532,9 @@ class _Duration(Duration):
 
     @staticmethod
     def delta_to_json(delta: timedelta) -> str:
-        parts = str(delta.total_seconds()).split(".")
+        parts = f"{delta.total_seconds():f}".split(".")
         if len(parts) > 1:
+            parts[1] = parts[1].rstrip('0')
             while len(parts[1]) not in (3, 6, 9):
                 parts[1] = f"{parts[1]}0"
         return f"{'.'.join(parts)}s"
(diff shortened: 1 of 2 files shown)


src/betterproto/__init__.py
```diff
@@ -1532,8 +1532,9 @@ def to_timedelta(self) -> timedelta:
 
     @staticmethod
     def delta_to_json(delta: timedelta) -> str:
-        parts = str(delta.total_seconds()).split(".")
+        parts = f"{delta.total_seconds():f}".split(".")
         if len(parts) > 1:
+            parts[1] = parts[1].rstrip('0')
```
Gobot1234, line 1537:
```suggestion
            parts[1] = parts[1].rstrip("0")
```
We use black to format code

Issue #459, opened 2023-02-21
Optional fields are added inside a group


SamuelYvon:
If one generates a message with an `optional` field, a group is added around the field.

Input:
```protobuf
syntax = "proto3";

package bug;

message BuggyMessage {
    uint32 field_a = 1;
    optional uint32 field_b = 2;
};
```

Expected:
```python
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: bug.proto
# plugin: python-betterproto
# This file has been @generated

from dataclasses import dataclass

import betterproto


@dataclass(eq=False, repr=False)
class BuggyMessage(betterproto.Message):
    field_a: int = betterproto.uint32_field(1)
    field_b: int = betterproto.uint32_field(2, optional=True)
```

Actual:
```python
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: bug.proto
# plugin: python-betterproto
# This file has been @generated

from dataclasses import dataclass
from typing import Optional

import betterproto


@dataclass(eq=False, repr=False)
class BuggyMessage(betterproto.Message):
    field_a: int = betterproto.uint32_field(1)
    field_b: Optional[int] = betterproto.uint32_field(
        2, optional=True, group="_field_b"
    )
```

I am not sure why it is done that way, but having the field in a group interferes with the custom flag `--custom_opt=pydantic_dataclasses`: every `None` value will fail because no member of the group will be set.

It seems it's not been a problem for the users of this library, so I'm assuming it is the intended behavior. However, it must have changed recently, because between the merge of PR #406 and it's opening, this was not how it was generated.

Gobot1234:
Nothing touching code generation has been merged to the master branch, so there must have been a bug in the original implementation

SamuelYvon:
Are you sure? We tested our usage with my fork yesterday with success, but today our deployment with upstream failed due to this error :(

I can fix the pydantic related issues by tweaking the group check (ie, if the fields are optional, allow them to be left None) but if you think this is a bug in the original code generation, I can take a look and fix that instead

Gobot1234:
The only code that's been merged semi recently was dropping 3.6 and fixing CI due to that merge https://github.com/danielgtaylor/python-betterproto/commits/master

SamuelYvon:
[...]

PR #406, opened 2022-08-04
Add support for pydantic dataclasses

SamuelYvon:
Pydantic dataclasses are a drop-in replacement for builtin dataclasses
to offer a stronger type validation. This PR adds support for
generating pydantic dataclasses instead of the builtin variant.

This enable usage of protobuf classes in API endpoint (like FastAPI)
which supports parsing pydantic objects but not builtin dataclasses.

PR #460, opened 2023-02-21
Fix for #459 (pydantic code gen only)

SamuelYvon:
This pull request addresses my concerns with how the oneofs groups are generated for `optional` fields in proto3 files. This stops pydantic from complaining that the groups have unset values when the field is optional.

I also re-included the old

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
else:
    from pydantic.dataclasses import dataclass
```

In pull request #406 it was decided to remove it, but it causes issues with keywords arguments and `mypy` v >= 1.0.0

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1471,12 +1471,24 @@ class Message(ABC):
 
     @classmethod
     def _validate_field_groups(cls, values):
-        meta = cls._betterproto_meta.oneof_field_by_group  # type: ignore
+        group_to_one_ofs = cls._betterproto_meta.oneof_field_by_group  # type: ignore
+        field_name_to_meta = cls._betterproto_meta.meta_by_field_name  # type: ignore
 
-        for group, field_set in meta.items():
+        for group, field_set in group_to_one_ofs.items():
+
+            if len(field_set) == 1:
+                (field,) = field_set
+                field_name = field.name
+                meta = field_name_to_meta[field_name]
+
+                # This is a synthetic oneof; we should ignore it's presence and not consider it as a oneof.
+                if meta.optional:
+                    continue
+
             set_fields = [
                 field.name for field in field_set if values[field.name] is not None
             ]
+
             if not set_fields:
                 raise ValueError(f"Group {group} has no value; all fields are None")
             elif len(set_fields) > 1:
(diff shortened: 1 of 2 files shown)


Gobot1234 (approved):

SamuelYvon:
This looks like this now:

Message

```proto
syntax = "proto3";

package bug;

message BuggyMessage {
    uint32 field_a = 1;
    optional uint32 field_b = 2;
};

message OneOfMessage {
    uint32 field_a = 1;
    oneof oof {
        uint32 field_b = 2;
        uint32 field_c = 3;
    };
};
```

Compiled
```python

class BuggyMessage(betterproto.Message):
    field_a: int = betterproto.uint32_field(1)
    field_b: Optional[int] = betterproto.uint32_field(
        2, optional=True, group="_field_b"
    )

    @root_validator()
    def check_oneof(cls, values):
        return cls._validate_field_groups(values)


class OneOfMessage(betterproto.Message):
    field_a: int = betterproto.uint32_field(1)
    field_b: Optional[int] = betterproto.uint32_field(2, optional=True, group="oof")
    field_c: Optional[int] = betterproto.uint32_field(3, optional=True, group="oof")

    @root_validator()
    def check_oneof(cls, values):
        return cls._validate_field_groups(values)
```

Behavior
```shell
(ecam-ai-appliance-py3.9)  syvon@pop-os  garda/appliance   misc/hotfixes ±  python
Python 3.9.16 (main, Dec  7 2022, 01:12:08)
[GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from messages.bug import BuggyMessage
>>> from messages.bug import OneOfMessage
>>> b = BuggyMessage(field_a=5)
>>> o = OneOfMessage(field_a=5)
  File "<stdin>", line 1, in <module>
  File "pydantic/dataclasses.py", line 287, in pydantic.dataclasses._add_pydantic_validation_attributes.handle_extra_init

  File "<string>", line 6, in __init__
  File "pydantic/dataclasses.py", line 309, in pydantic.dataclasses._add_pydantic_validation_attributes.new_post_init
    return ('_DataclassParams('
  File "pydantic/dataclasses.py", line 429, in pydantic.dataclasses._dataclass_validate_values
    else:
pydantic.error_wrappers.ValidationError: 1 validation error for OneOfMessage
__root__
  Group oof has no value; all fields are None (type=value_error)
>>>
```

Issue #467, opened 2023-03-15
Inconsistancy between Timestamp field serailized via protobuf and json


semicolonTransistor:
This happens when a timezone-aware `datetime` object is assigned to a Timestamp field. The protobuf serialization correctly serializes the timestamp objects via `datetime.timestamp()` method. However, the JSON serializer assumes the date time is UTC regardless of whether the `datetime` object is timezone-aware. This results in messages deserialized from JSON to have an incorrect `datetime` value while messages deserialized from protobuf have the correct `datetime` value.

Here is a short example of the problem.

```protobuf
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message Timestamp {
  google.protobuf.Timestamp ts = 1;
}
```

```python
from proto.timestamp import Timestamp
from datetime import datetime, timezone, timedelta


def main():

    # create a timezone aware date time object of timezone utc-4
    current_time = datetime.now(tz=timezone(timedelta(hours=-4)))

    print(f"{current_time=}")

    original_message = Timestamp()
    original_message.ts = current_time

    print(f"{original_message.ts=}")

    # serialize and then deserialize the message
    protobuf_serialized = bytes(original_message)
    json_serialized = original_message.to_json()
    
    print()
    print(f"{json_serialized=}")

    protobuf_deserialized_message = Timestamp()
    protobuf_deserialized_message.parse(protobuf_serialized)

    json_deserialized_message = Timestamp()
    json_deserialized_message.from_json(json_serialized)

    print()
    print(f"{protobuf_deserialized_message.ts=}")
    print(f"{json_deserialized_message.ts=}")

    print()
    print(f"{(protobuf_deserialized_message.ts==original_message.ts)=}")
    print(f"{(json_deserialized_message.ts==original_message.ts)=}")


if __name__ == "__main__":
    main()

```

running the above code generates the following output:

```
current_time=datetime.datetime(2023, 3, 14, 21, 52, 16, 513477, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=72000)))
original_message.ts=datetime.datetime(2023, 3, 14, 21, 52, 16, 513477, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=72000)))

json_serialized='{"ts": "2023-03-14T21:52:16.513477Z"}'

protobuf_deserialized_message.ts=datetime.datetime(2023, 3, 15, 1, 52, 16, 513477, tzinfo=datetime.timezone.utc)
[...]

PR #468, opened 2023-03-16
Fix dict encoding for timezone aware datetimes

semicolonTransistor:
See #467

When a timezone-aware datetime object is encoded via `_Timestamp.timestamp_to_json` its timezone information is stripped and it is encoded as if it is UTC. This caused the decoded datetime from the dict to refer to a different point in time from the original datetime. Since the JSON serialization routine calls the dict encode routine this problem also affects JSON.
I added a check in `_Timestamp.timestamp_to_json` that converts timezone-aware datetime object to UTC before stripping the timezone and encoding. I also added some tests to check that timezone-aware datetime objects are encoded and decoded correctly. Naive datetime objects are still treated as UTC which is unchanged.


Gobot1234:
Thanks for all this work

Gobot1234 (approved):

PR #476, opened 2023-04-18
Make Message.from_dict() a class method

MarekPikula:
## Rationale

For Pydantic dataclasses, it's impossible to initialize the model with `Message().from_dict(init_dict)` as the initializer of `Message` (in this case `pydantic.dataclasses.dataclass` initializer) performs the validation on init. `Message(**init_dict)` is also impossible, as the default Pydantic dataclass init doesn't perform the necessary casing and type transformations done by `from_dict()`.

## Solution

To minimise existing code breakage, a custom `hybridmethod` was used to implement the `from_dict()` both in the class and instance context. It can be used to have both classmethod and instance method with the same name and execute the right one depending on the context. Thus both the current `Message().from_dict(init_dict)` and `Message.from_dict(init_dict)` should work as expected. Let me know if you know a better or more standard way of implementing it (maybe something in `typing` I missed?).

## Possible breakage – up to discussion

(I think) that the only possible breakage of prior uses would be if instance-level `from_dict()` was applied to an existing instance if the instance has been modified by hand on runtime. That's because for the `meta.proto_type == TYPE_MESSAGE` the field type is checked for the schema and not for the modified instance. The same applies for `meta.proto_type == TYPE_ENUM` and `meta.map_types and meta.map_types[1] == TYPE_MESSAGE` (in the form of overwriting the entire dictionary instead of applying the modification from init dict).

The question is if these behaviours are even used by anyone (or even if they should be – to me, it looks pretty dangerous). In the case of tests, they are not (all tests pass). It is, of course, possible to revert the instance-level `from_dict()` back to the previous form with the downside of duplication of a rather convoluted piece of code (currently common in the form of `_from_dict_init()`). Please let me know what you think.

## Example usage

I currently use this fork in my project [MarekPikula/python-headscale-api](https://github.com/MarekPikula/python-headscale-api), which uses Pydantic dataclasses generated from the Headscale protobuf schema. The usage of the new `from_dict()` is in [headscale.py](https://github.com/MarekPikula/python-headscale-api/blob/v0.2.0/headscale_api/headscale.py#L214).


MarekPikula:
Hi, is there any chance to review this change? It would be brilliant if it was merged and (pre-)released, as I depend on this change in a project, which I want to release on PyPI (so I cannot use git-originating dependencies).

Darkheir:
I agree that missing a class method make it complicated to use `pydantic` models to parse raw messages.

src/betterproto/__init__.py
```diff
@@ -1178,61 +1217,40 @@ def to_dict(
...
-
-        Returns
-        --------
-        :class:`Message`
-            The initialized message.
-        """
-        self._serialized_on_wire = True
+    @classmethod
+    def _from_dict_init(cls: Type[T], value: Mapping[str, Any]) -> Mapping[str, Any]:
+        metadata = ProtoClassMetadata(cls)
```
Gobot1234, line 1222: This is not the correct way to work around this problem. ProtoClassMetadata should be made a cached class property.

src/betterproto/__init__.py
```diff
@@ -1178,61 +1217,40 @@ def to_dict(
...
-            The dictionary to parse from.
-
-        Returns
-        --------
-        :class:`Message`
-            The initialized message.
-        """
-        self._serialized_on_wire = True
+    @classmethod
+    def _from_dict_init(cls: Type[T], value: Mapping[str, Any]) -> Mapping[str, Any]:
```
Gobot1234, line 1221:
```suggestion
    def _from_dict_init(cls, value: Mapping[str, Any]) -> Mapping[str, Any]:
```

src/betterproto/__init__.py
```diff
@@ -1178,61 +1217,40 @@ def to_dict(
...
                     elif meta.wraps:
-                        setattr(self, field_name, value[key])
-                    elif v is None:
-                        setattr(self, field_name, cls().from_dict(value[key]))
+                        v = value[key]
                     else:
-                        # NOTE: `from_dict` mutates the underlying message, so no
-                        # assignment here is necessary.
-                        v.from_dict(value[key])
+                        v = sub_cls.from_dict(value[key])
```
Gobot1234, line 1250: Please keep the comment

src/betterproto/__init__.py
```diff
@@ -1178,61 +1217,40 @@ def to_dict(
...
-        --------
-        :class:`Message`
-            The initialized message.
-        """
-        self._serialized_on_wire = True
+    @classmethod
+    def _from_dict_init(cls: Type[T], value: Mapping[str, Any]) -> Mapping[str, Any]:
+        metadata = ProtoClassMetadata(cls)
+        init_dict: Dict[str, Any] = {}
         for key in value:
```
Gobot1234, line 1224: Please could you update this to use .items()

src/betterproto/__init__.py
```diff
@@ -601,6 +603,43 @@ def _get_cls_by_field(
...
+    def instancemethod(self, finstance: HybridT):
+        return type(self)(self.fclass, finstance, self.__doc__)
+
+    def __get__(self, instance: Optional[T], cls: Type[T]):
+        if instance is None or self.finstance is None:
+            # either bound to the class, or no instance method available
+            return self.fclass.__get__(cls, None)
+        return self.finstance.__get__(instance, cls)
+
+
```
Gobot1234, line 642: This file is starting to get very bloated, would you be able to pull this into a utils file or similar

src/betterproto/__init__.py
```diff
@@ -601,6 +603,43 @@ def _get_cls_by_field(
...
+    """
+
+    def __init__(
+        self,
+        fclass: Type[HybridT],
+        finstance: Optional[HybridT] = None,
+        doc: Optional[str] = None,
+    ):
+        self.fclass = fclass
+        self.finstance = finstance
```
Gobot1234, line 625: Not sure why these are f prefixed. Please remove the prefix and name them something like owner and instance as is commonly  done with the descriptor protocol

Gobot1234 (changes requested):
Might be nice to add support for parse as a classmethod as well with this

src/betterproto/__init__.py
```diff
@@ -882,18 +889,19 @@ def __copy__(self: T, _: Any = {}) -> T:
                 kwargs[name] = value
         return self.__class__(**kwargs)  # type: ignore
 
-    @property
+    @classproperty
     def _betterproto(self) -> ProtoClassMetadata:
```
Gobot1234, line 893:
```suggestion
    def _betterproto(cls) -> ProtoClassMetadata:
```

Gobot1234 (approved):

cetanu (approved):

PR #481, opened 2023-04-25
Handle empty value objects properly

### LINKED ISSUE #479: Some message structures are not serialised properly
Hi, the project I am working on makes use of "AppEmpty" messages which signify that a certain element should be sent back. The proto looks something like this:

```proto
message Request {
		required uint32 seq = 1;
		required uint64 playerId = 2;
		required int32 playerToken = 3;
		
		optional AppEmpty getInfo = 8;
		optional AppEmpty getTime = 9;
}

message AppEmpty {	
}
```

Where I am then using betterproto to be able to do this:
```py
request = AppRequest()
request.seq = 1
request.player_id = 5
request.player_token = 10
request.get_info = AppEmpty()
print(betterproto.serialized_on_wire(request.get_info)) # False
```
However, the `get_info` request is not then serialised. I may be doing something wrong, however currently I am just setting the internal serialisation state with:
```py
app_request.get_info._serialized_on_wire = True
```

Am I doing something wrong or is there a problem?

olijeffers0n:
Closes #479 by attempting to set the serialization state upon setting the attribute. I am not familiar with this codebase however I believe this should work properly, and does from my testing.

For context, the previous approach only set the serialisation state to be truthy when an attribute of this "child" message was updated. This does not work for empty message structures so it needs to be set when the new object is assigned instead.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -704,6 +704,10 @@ class Message(ABC):
             return value
 
     def __setattr__(self, attr: str, value: Any) -> None:
+    
+        if hasattr(value, "_serialized_on_wire") and hasattr(value, "_betterproto") and len(value._betterproto.meta_by_field_name) == 0:
+            value._serialized_on_wire = True
+    
         if attr != "_serialized_on_wire":
             # Track when a field has been set.
             self.__dict__["_serialized_on_wire"] = True
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -704,7 +704,11 @@ class Message(ABC):
             return value
 
     def __setattr__(self, attr: str, value: Any) -> None:
-        if isinstance(value, Message) and hasattr(value, "_betterproto") and not len(value._betterproto.meta_by_field_name):
+        if (
+            isinstance(value, Message) 
+            and hasattr(value, "_betterproto") 
+            and not len(value._betterproto.meta_by_field_name)
+        ):
             value._serialized_on_wire = True
     
         if attr != "_serialized_on_wire":
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -707,7 +707,7 @@ class Message(ABC):
         if (
             isinstance(value, Message)
             and hasattr(value, "_betterproto")
-            and not len(value._betterproto.meta_by_field_name)
+            and not value._betterproto.meta_by_field_name
         ):
             value._serialized_on_wire = True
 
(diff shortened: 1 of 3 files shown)


Gobot1234:
I think this is better handled in `__post_init__` no?

olijeffers0n:
I thought about that, but i’m not sure how that would work as I do not want the instance that is assigned statically to be serialised. If you know how to fix this better, please supersede this PR

olijeffers0n:
@Gobot1234 How would you like to proceed with this / what can I do better here?

src/betterproto/__init__.py
```diff
@@ -704,6 +704,10 @@ def __getattribute__(self, name: str) -> Any:
             return value
 
     def __setattr__(self, attr: str, value: Any) -> None:
+    
+        if hasattr(value, "_serialized_on_wire") and hasattr(value, "_betterproto") and len(value._betterproto.meta_by_field_name) == 0:
```
Gobot1234, line 708:
```suggestion
        if isinstance(value, Message) and hasattr(value, "_betterproto") and not len(value._betterproto.meta_by_field_name):
```

olijeffers0n:
Looks good to me

src/betterproto/__init__.py
```diff
@@ -704,6 +704,9 @@ def __getattribute__(self, name: str) -> Any:
             return value
 
     def __setattr__(self, attr: str, value: Any) -> None:
+        if isinstance(value, Message) and hasattr(value, "_betterproto") and not len(value._betterproto.meta_by_field_name):
```
Gobot1234, line 707:
```suggestion
        if (
            isinstance(value, Message) 
            and hasattr(value, "_betterproto") 
            and not len(value._betterproto.meta_by_field_name)
        ):
```

Gobot1234: Needs formatting with black

Gobot1234:
Would you mind adding a test case for this regression? (and also fixing the formatting better than I did 😉 `poe fmt`)

olijeffers0n:
Thanks for reviewing - I am in the middle of exams at the moment though so won’t be able to get to this for a while. Plus, I don’t really know how to write tests but i could work that out

Gobot1234:
@olijeffers0n friendly poke, if you don't have time to finish this I can take over if you want

olijeffers0n:
Yeah, I have time I just don’t know how to write unit tests. I’m happy for you to take over

Gobot1234 (approved):

Gobot1234 (approved):

Gobot1234:
Thanks

Issue #490, opened 2023-05-14
`to_dict` or `to_pydict` miss integer and boolean fields


alirezasafi:
when i use `to_dict` or `to_pydict` function to serialize message, `False` and `0` values are not work for boolean and integer fields.
here is the sample code:
```python3
from dataclasses import dataclass
import betterproto


@dataclass
class Greeting(betterproto.Message):
    """Greeting represents a message you can tell a user."""

    int_field: int = betterproto.int32_field(1)
    bool_fielde: bool = betterproto.bool_field(2)


if __name__ == "__main__":
    greeting = Greeting(
        int_field=0,
        bool_fielde=False
    )
    print(greeting.to_dict())
    greeting.int_field = 1
    greeting.bool_fielde = True
    print(greeting.to_dict())
``` 
and outpu is:
```
{}
{'intField': 1, 'boolFielde': True}
```

alirezasafi:
we could change `to_pydict` to this:
```python3
    def to_pydict(
        self, casing: Casing = Casing.CAMEL, include_default_values: bool = False
    ) -> Dict[str, Any]:
        """
        Returns a python dict representation of this object.

        Parameters
        -----------
        casing: :class:`Casing`
            The casing to use for key values. Default is :attr:`Casing.CAMEL` for
            compatibility purposes.
        include_default_values: :class:`bool`
            If ``True`` will include the default values of fields. Default is ``False``.
            E.g. an ``int32`` field will be included with a value of ``0`` if this is
            set to ``True``, otherwise this would be ignored.

        Returns
        --------
        Dict[:class:`str`, Any]
            The python dict representation of this object.
        """
        output: Dict[str, Any] = {}
        defaults = self._betterproto.default_gen
        for field_name, meta in self._betterproto.meta_by_field_name.items():
            field_is_repeated = defaults[field_name] is list
            if self.__raw_get(field_name) is PLACEHOLDER and \
                not include_default_values:
                continue
            value = getattr(self, field_name)
            cased_name = casing(field_name).rstrip("_")  # type: ignore
            if meta.proto_type == TYPE_MESSAGE:
                if isinstance(value, datetime):
                    if (
                        value != DATETIME_ZERO
                        or include_default_values
                        or self._include_default_value_for_oneof(
                            field_name=field_name, meta=meta
                        )
                    ):
                        output[cased_name] = value
                elif isinstance(value, timedelta):
                    if (
                        value != timedelta(0)
                        or include_default_values
[...]

PR #494, opened 2023-05-27
Fix to-pydict function. (Issue #490)

alirezasafi:
[#490](https://github.com/danielgtaylor/python-betterproto/issues/490)

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1344,6 +1344,9 @@ class Message(ABC):
         defaults = self._betterproto.default_gen
         for field_name, meta in self._betterproto.meta_by_field_name.items():
             field_is_repeated = defaults[field_name] is list
+            if self.__raw_get(field_name) is PLACEHOLDER and \
+                not include_default_values:
+                continue
             value = getattr(self, field_name)
             cased_name = casing(field_name).rstrip("_")  # type: ignore
             if meta.proto_type == TYPE_MESSAGE:
@@ -1388,13 +1391,7 @@ class Message(ABC):
 
                 if value or include_default_values:
                     output[cased_name] = value
-            elif (
-                value != self._get_field_default(field_name)
-                or include_default_values
-                or self._include_default_value_for_oneof(
-                    field_name=field_name, meta=meta
-                )
-            ):
+            else:
                 output[cased_name] = value
         return output
 


Gobot1234:
Hi, thank you for the PR, would you be able to run `poe fmt` and add a regression test for this code change?

Gobot1234:
I'm very hesitant to make a change of behaviour like this just for bool fields. Does the proposed change apply to all fields which are set and include them even if bool(value) returns False?

alirezasafi:
> proposed
yes its applied to all of them and include bool field even if the value is `False`.
here is the sample code:
```python3

import betterproto


@dataclass
class Greeting(betterproto.Message):
    """Greeting represents a message you can tell a user."""

    int_field: int = betterproto.int32_field(1)
    bool_field: bool = betterproto.bool_field(2)


if __name__ == "__main__":
    greeting = Greeting(
        bool_field=False,
    )
    print("default value is disabled (bool_field=False): ", greeting.to_pydict(include_default_values=False))
    print("default value is enabled (bool_field=False): ", greeting.to_pydict(include_default_values=True))
    greeting = Greeting()
    print("default value is disabled (unset bool_field): ", greeting.to_pydict(include_default_values=False))
    print("default value is enabled (unset bool_field): ", greeting.to_pydict(include_default_values=True))
```
output: 
```
default value is disabled (bool_field=False):  {'boolField': False}
default value is enabled (bool_field=False):  {'intField': 0, 'boolField': False}
default value is disabled (unset bool_field):  {}
default value is enabled (unset bool_field):  {'intField': 0, 'boolField': False}
```

Gobot1234:
Sorry if I wasn't clear, I meant what's the output for things that implement bool that aren't int subclasses. ie what does 
```py

import betterproto

@dataclass
class Msg(bettterproto.Message):
    bar: bool = betterproto.bool_field(1)

@dataclass
class Greeting(betterproto.Message):
    """Greeting represents a message you can tell a user."""

    msg_field: Msg = betterproto.message_field(1)
    str_field: str = betterproto.string_field(2)


if __name__ == "__main__":
    greeting = Greeting(
        Msg(), ""
    )
    print(greeting.to_pydict())
```
print?

alirezasafi:
> ```python
> from dataclasses import dataclass
> 
> import betterproto
> 
> @dataclass
> class Msg(bettterproto.Message):
>     bar: bool = betterproto.bool_field(1)
> 
> @dataclass
> class Greeting(betterproto.Message):
>     """Greeting represents a message you can tell a user."""
> 
>     msg_field: Msg = betterproto.message_field(1)
>     str_field: str = betterproto.string_field(2)
> 
> 
> if __name__ == "__main__":
>     greeting = Greeting(
>         Msg(), ""
>     )
>     print(greeting.to_pydict())
> ```

only `strField` is in the output:
```
{'strField': ''}
```

Gobot1234:
Interesting, I would expect an empty dict for the msg field cause it's set

alirezasafi:
> ```python
> bool_field
> ```

why?
i think the output is correct. because bool field `bar` didn't set and default value disabled.

Gobot1234:
I'd expect the output to be the output you've shown if the `Greeting(str_field="")` but according to proto spec setting the value of msg_field still means it should show up in the wire format so I think the output should be `{'msgField': Msg(), 'strField': ''}` for the first input I gave in https://github.com/danielgtaylor/python-betterproto/pull/494#issuecomment-1565356331

Gobot1234:
Please could you add some unit tests for this feature please

cetanu:
> I'd expect the output to be the output you've shown if the `Greeting(str_field="")` but according to proto spec setting the value of msg_field still means it should show up in the wire format so I think the output should be `{'msgField': Msg(), 'strField': ''}` for the first input I gave in [#494 (comment)](https://github.com/danielgtaylor/python-betterproto/pull/494#issuecomment-1565356331)

I agree, if a message is placed into a field, I would consider it set and expect there to be an empty dict there, if the message itself has no fields set.

PR #495, opened 2023-05-27
Fix pydict serialization for optional fields

### LINKED ISSUE #475: TypeError in to_pydict with optional fields
See https://github.com/danielgtaylor/python-betterproto/issues/475#issuecomment-1512296041

<details>
<summary>Outdated</summary>

I am using betterproto to load a ton of messages at once from a binary file, however this operation is very slow so I wanted to cache/parse it into json using `json.dump` with `to_dict`, and then load it directly from json. Problem is that `from_json` is also very slow, so my idea was to use `to_dict(casing=Casing.SNAKE, include_default_values=true)` to generate a dictionary that matches the attributes of the message dataclass, and then wrap it in an utility class that redirects dict.attribute calls to dict["attribute"] so that for the consumer (which is using the message dataclasses exclusively for their data) there's no difference.

However the problem is that when serializing to dictionary, enum values are serialized based on their name instead of value. Maybe an optional parameter can be added to change this behaviour? Thank you

<details>
<summary>Utility dictionary class for redirecting attributes to items</summary>

```py
class AttrDict(dict):
    def __getattr__(self, key):
        try:
            return self[key]
        except KeyError:
            raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{key}'")

    def __setattr__(self, key, value):
        self[key] = value

    def __getitem__(self, key):
        value = super().__getitem__(key)
        if isinstance(value, dict) and not isinstance(value, AttrDict):
            value = AttrDict(value)
            self[key] = value
        if isinstance(value, list) and any(isinstance(x, dict) for x in value):
            value = [AttrDict(x) if isinstance(x, dict) else x for x in value]
            self[key] = value
        return value
```

</details>

</details>

nickderobertis:
## What

Adds a block in `Mesaage.to_pydict` that explicitly checks for `None` values and assigns them to the dict if `include_default_values=True`.

## Why

Fixes #475 where `to_pydict` fails on optional fields with the error:

```
  File "...", in <listcomp>
    json.dump([f.to_pydict(casing=Casing.SNAKE, include_default_values=True) for f in frames], file)
  File "...\Python\Python310\lib\site-packages\betterproto\__init__.py", line 1365, in to_pydict
    output[cased_name] = value.to_pydict(casing, include_default_values)
  File "...\Python\Python310\lib\site-packages\betterproto\__init__.py", line 1359, in to_pydict
    value._serialized_on_wire
AttributeError: 'NoneType' object has no attribute '_serialized_on_wire'
```

## Testing

Added a test case that checks `to_pydict` and `to_dict` output when passing the message value and not, and when passing `include_default_values` and not.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1373,6 +1373,9 @@ class Message(ABC):
                     value = [i.to_pydict(casing, include_default_values) for i in value]
                     if value or include_default_values:
                         output[cased_name] = value
+                elif value is None:
+                    if include_default_values:
+                        output[cased_name] = None
                 elif (
                     value._serialized_on_wire
                     or include_default_values
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -265,6 +265,32 @@ def test_optional_flag():
     assert Request().parse(b"\n\x00").flag is False
 
 
+def test_optional_datetime_to_dict():
+    @dataclass
+    class Request(betterproto.Message):
+        date: Optional[datetime] = betterproto.message_field(1, optional=True)
+
+    # Check dict serialization
+    assert Request().to_dict() == {}
+    assert Request().to_dict(include_default_values=True) == {"date": None}
+    assert Request(date=datetime(2020, 1, 1)).to_dict() == {
+        "date": "2020-01-01T00:00:00Z"
+    }
+    assert Request(date=datetime(2020, 1, 1)).to_dict(include_default_values=True) == {
+        "date": "2020-01-01T00:00:00Z"
+    }
+
+    # Check pydict serialization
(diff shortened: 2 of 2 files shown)


Gobot1234 (approved):
Thank you for the PR

nickderobertis:
Thanks for reviewing and merging it quickly! When do you think we'll cut a new release with this change? Trying to see if I need to peg to git version or if it will be available in a v2 beta release soon.

Gobot1234:
I can't make any guarantees as I can't make releases by myself so I'll have to recommend that you pin to a git rev, sorry.

nickderobertis:
Ok that makes sense, thank you for the response.

PR #496, opened 2023-05-29
Swappable service implementations

jcomo:
To support Twirp. Instead of the hack in #1, this PR refactors the library to be modular and allow different service generators.


jcomo:
Apologies - meant to open this PR against my fork. Would be happy to clean this up a bit more and submit upstream if there's interest

PR #508, opened 2023-07-06
Rich support

### LINKED ISSUE #507: Support Rich, pretty printing
By manually modifying generated code from `@dataclass(eq=False, repr=False)` to `@dataclass(eq=False, repr=True)`, I managed to get Rich's `pprint` to work, so being able to tell it to not set it to `False` would already help.

Rich has also this: https://rich.readthedocs.io/en/latest/pretty.html#rich-repr-protocol

By defining a `__rich_repr__` method (no need to have Rich as a dependency at all, the method doesn't depend on it) that yields the attributes, it's also possible to have better support for Rich.

A quick and probably wrong attempt at implementing it:
```py
    def __rich_repr__(self) -> Iterable[Tuple[str, Any, Any]]:
        for field_name in self._betterproto.sorted_field_names:
            yield field_name, self.__raw_get(field_name), PLACEHOLDER
```

So with basically 3 lines of harmless code and no extra dependencies, it's possible to add native support for it.

MicaelJarniac:
Closes #507

https://rich.readthedocs.io/en/latest/pretty.html#rich-repr-protocol

~~I haven't tested it yet, and I've used the GitHub web editor. But I believe it should work. Will test soon I hope.~~

Update: I've been using this for a while and it works!

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -688,6 +688,10 @@ class Message(ABC):
         ]
         return f"{self.__class__.__name__}({', '.join(parts)})"
 
+    def __rich_repr__(self) -> Iterable[Tuple[str, Any, Any]]:
+        for field_name in self._betterproto.sorted_field_names:
+            yield field_name, self.__raw_get(field_name), PLACEHOLDER
+
     if not TYPE_CHECKING:
 
         def __getattribute__(self, name: str) -> Any:


src/betterproto/__init__.py
```diff
@@ -688,6 +688,10 @@ def __repr__(self) -> str:
         ]
         return f"{self.__class__.__name__}({', '.join(parts)})"
 
+    def __rich_repr__(self) -> Iterable[Tuple[str, Any, Any]]:
+        for field_name in self._betterproto.sorted_field_names:
+            yield field_name, self.__raw_get(field_name), PLACEHOLDER
```
MicaelJarniac, line 693: I think this could be a `yield from`, something like this:
```suggestion
        yield (name, self.__raw_get(name), PLACEHOLDER for name in self._betterproto.sorted_field_names)
```

MicaelJarniac: Using `for ... yield` seems clearer, though.

cetanu (approved):

Issue #358, opened 2022-03-23
Proposal to make oneof-deconstruction typecheckable


124C41p:
According to the docs, deconstruction of a oneof-type is meant to be done in the following way:

```proto
syntax = "proto3";

message Test {
  oneof foo {
    bool on = 1;
    int32 count = 2;
    string name = 3;
  }
}
```

```python
foo_type, foo_value = betterproto.which_one_of(test, 'foo')

if foo_type == 'on':
  process_on(foo_value)
elif foo_type == 'count':
  process_count(foo_value)
elif foo_type == 'name':
  process_name(foo_value)
else:
  raise ValueError('No value for foo specified!')
```

Unfortunately, in that way development tooling (pylance, mypy, etc) can neither check for typos nor for completeness. (Imagine to introduce a new field `string address = 4;` or to remove on of the other fields later. A type checker could not see any problem with our python code after executing protoc.)

There are different possible ways to solve that issue.

### Variant 1 - Make all sub-fields of `foo` optional
You could change the implementation of the `Test` dataclass such that all sub fields of `foo` are optional, and the unset fields are set to `None`. Then our deconstruction could look like this:

```python
if test.on is not None:
  process_on(test.on)
elif test.count is not None:
  process_count(test.count)
elif test.name is not None:
  process_name(test.name)
else:
  raise ValueError('No value for foo specified!')
```

### Variant 2 - Introduce an enum specifying the field set
You could add an enum field to the `Test` dataclass which encodes the actual field set. Then our deconstruction could look like this:

```python
if test.foo_type == FooType.ON:
  process_on(test.on)
elif test.foo_type == FooType.COUNT:
  process_count(test.count)
elif test.foo_type == FooType.NAME:
  process_name(test.name)
else:
  raise ValueError('No value for foo specified!')
```

There are certainly many other ways to make typechecking possible. I really hope you agree with me that such a feature would increase development quality (especially since the rest of your auto generated code is beautifully typecheckable - I like that a lot!).

Thank you in advance!

Gobot1234:
I really like the idea and I was actually thinking about this a while back. I'm not a huge fan of making all the fields optional as it could lead to confusion as to why this is different from optional fields and it makes unwrapping fields a pain. Adding a new field and generating members for the enums would be a big compiler change. 

My ideal solution would be to implement oneofs as a context manager, maybe we could try something like.
```py
@dataclass(repr=False)
class Something(betterproto.Message):
[...]

PR #510, opened 2023-07-11
Raise `AttributeError` on attempts to access unset `oneof` fields

a-khabarov:
This commit modifies `Message.__getattribute__` to raise `AttributeError` whenever an attempt is made to access an unset `oneof` field. This provides several benefits over the current approach:

* There is no longer any risk of `betterproto` users accidentally relying on values of unset fields.

* Pattern matching with `match/case` on messages containing `oneof` groups is now supported. The following is now possible: 
  ```python3
  @dataclasses.dataclass(eq=False, repr=False)
    x: int = betterproto.int32_field(1, group="g")
    y: str = betterproto.string_field(2, group="g")

  match Test(y="text"):
    case Test(x=v):
      print("x", v)
    case Test(y=v):
      print("y", v) 
  ```
  Before this commit the code above would output `x 0` instead of `y text`, but now the output is `y text` as expected. The reason this works is because an `AttributeError` in a `case` pattern does not propagate and instead simply skips the `case`.

* We now have a type-checkable way to deconstruct `oneof`. When running `mypy` for the snippet above `v` has type `int` in the first `case` and type `str` in the second `case`. For versions of Python that do not support `match/case` (before 3.10) it is now possbile to use `try/except/else` blocks to achieve the same result:
  ```python3
  t = Test(y="text")
  try:
    v0: int = t.x
  except AttributeError:
    v1: str = t.y  # `oneof` contains `y`
  else:
    pass  # `oneof` contains `x`
  ```

This is a breaking change. The previous behavior is still accessible via `Message.__unsafe_get`.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -693,8 +693,28 @@ class Message(ABC):
         def __getattribute__(self, name: str) -> Any:
             """
             Lazily initialize default values to avoid infinite recursion for recursive
-            message types
+            message types.
+            Raise :class:`AttributeError` on attempts to access unset ``oneof`` fields.
             """
+            try:
+                group_current = super().__getattribute__("_group_current")
+            except AttributeError:
+                pass
+            else:
+                if name not in {"__class__", "_betterproto"}:
+                    group = self._betterproto.oneof_group_by_field.get(name)
+                    if group is not None and group_current[group] != name:
+                        if sys.version_info < (3, 10):
+                            raise AttributeError(
+                                f"{group!r} is set to {group_current[group]}, not {name!r}"
+                            )
+                        else:
+                            raise AttributeError(
+                                f"{group!r} is set to {group_current[group]}, not {name!r}",
+                                name=name,
+                                obj=self,
+                            )
+
             value = super().__getattribute__(name)
             if value is not PLACEHOLDER:
                 return value
@@ -761,7 +781,10 @@ class Message(ABC):
         """
         output = bytearray()
         for field_name, meta in self._betterproto.meta_by_field_name.items():
-            value = getattr(self, field_name)
+            try:
+                value = getattr(self, field_name)
+            except AttributeError:
+                continue
 
             if value is None:
                 # Optional items should be skipped. This is used for the Google
@@ -775,9 +798,7 @@ class Message(ABC):
             # Note that proto3 field presence/optional fields are put in a
             # synthetic single-item oneof by protoc, which helps us ensure we
             # send the value even if the value is the default zero value.
-            selected_in_group = (
-                meta.group and self._group_current[meta.group] == field_name
-            )
+            selected_in_group = bool(meta.group)
 
             # Empty messages can still be sent on the wire if they were
             # set (or received empty).
@@ -1016,7 +1037,12 @@ class Message(ABC):
                     parsed.wire_type, meta, field_name, parsed.value
                 )
 
-            current = getattr(self, field_name)
+            try:
+                current = getattr(self, field_name)
+            except AttributeError:
+                current = self._get_field_default(field_name)
+                setattr(self, field_name, current)
+
             if meta.proto_type == TYPE_MAP:
                 # Value represents a single key/value pair entry in the map.
                 current[value.key] = value.value
@@ -1077,7 +1103,10 @@ class Message(ABC):
         defaults = self._betterproto.default_gen
         for field_name, meta in self._betterproto.meta_by_field_name.items():
             field_is_repeated = defaults[field_name] is list
-            value = getattr(self, field_name)
+            try:
+                value = getattr(self, field_name)
+            except AttributeError:
+                value = self._get_field_default(field_name)
             cased_name = casing(field_name).rstrip("_")  # type: ignore
             if meta.proto_type == TYPE_MESSAGE:
                 if isinstance(value, datetime):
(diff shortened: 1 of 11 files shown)


a-khabarov:
I think this PR solves all of the problems mentioned in #358

src/betterproto/__init__.py
```diff
@@ -691,17 +704,18 @@ def __repr__(self) -> str:
...
+            # Raise `AttributeError` on attempts to access unset `oneof` fields
+            try:
+                group_current = super().__getattribute__("_group_current")
+            except AttributeError:
+                pass
+            else:
+                if name not in ("__class__", "_betterproto"):
+                    group = self._betterproto.oneof_group_by_field.get(name)
+                    if group is not None and group_current[group] != name:
+                        raise AttributeError()
```
mattyclarkson, line 716: It may be useful to add context to the missing attribute to the exception message.

The common attribute error that Python raises is as follows:
```
raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
```
However, you may want to give it even more context:
```
raise AttributeError(f"'{self.__class__.__name__}.{group}' is set to '{group_current[group]}' not '{name}'")
```
You may want to use `str.format` rather than an f-string depending on which Python versions this project supports.

src/betterproto/__init__.py
```diff
@@ -691,17 +704,18 @@ def __repr__(self) -> str:
...
-            value = super().__getattribute__(name)
-            if value is not PLACEHOLDER:
-                return value
+            # Raise `AttributeError` on attempts to access unset `oneof` fields
+            try:
+                group_current = super().__getattribute__("_group_current")
+            except AttributeError:
+                pass
+            else:
+                if name not in ("__class__", "_betterproto"):
```
mattyclarkson, line 713: You may want to change this to a set to get slightly better performance:

```suggestion
                if name not in {"__class__", "_betterproto"}:
```

```
$ python -m timeit '"abc" in ("def", "abc")'
10000000 loops, best of 5: 21.7 nsec per loop
$ python -m timeit '"abc" in {"def", "abc"}'
20000000 loops, best of 5: 13.5 nsec per loop```

Comment (a-khabarov) [reply src/betterproto/__init__.py]: I've amended the change to use `{"__class__", "_betterproto"}`.

**Comment by Gobot1234** [comment]
Thanks for the pr it looks good apart from the changes requiring you to access _Message__raw_get which should not be api that anyone should have to deal with

**Comment by a-khabarov** [comment]
_Message__unsafe_get is just for testing/use within betterproto itself, not for the users of betterproto. I don't see any reason why the users of betterproto would want to call _Message__unsafe_get instead of just using `.`.

**Comment by Gobot1234** [comment]
If it's just for testing like that can you just change it to `object.__getattribute__(message, name)`?

**Comment by a-khabarov** [comment]
Okay, I will remove it and update the tests to use `object.__getattribute__(message, name)`.

**Comment by a-khabarov** [comment]
I've removed `__unsafe_get` and updated the tests to use `object.__getattribute__(message, name)`.

File: src/betterproto/__init__.py [hunk 3]
```diff
@@ -693,8 +693,21 @@ def __repr__(self) -> str:
...
+                group_current = super().__getattribute__("_group_current")
+            except AttributeError:
+                pass
+            else:
+                if name not in {"__class__", "_betterproto"}:
+                    group = self._betterproto.oneof_group_by_field.get(name)
+                    if group is not None and group_current[group] != name:
+                        raise AttributeError(
+                            f"'{self.__class__.__name__}.{group}' is set to '{group_current[group]}', not '{name}'"
+                        )
```
Gobot1234, line 709:
```suggestion
                        if sys.version_info < (3, 10):
                            raise AttributeError(
                                f"{group!r} is set to {group_current[group]}, not {name!r}"
                            )
                        else:
                            raise AttributeError(
                                f"{group!r} is set to {group_current[group]}, not {name!r}",
                                name=name,
                                obj=self,
                            )
```
Should make error messages nicer if python/cpython#103232 is ever merged

a-khabarov: I've updated the change.

Gobot1234:
In future would you mind avoiding force pushing, it makes things harder to review and we squash the commits anyway

a-khabarov:
Sure. For `{group_current[group]}` in attribute errors, do we also need `!r`? The output at the moment seems to be of the form `AttributeError: 'g' is set to a, not 'b'` (`__repr__` not called for `group_current[group]`).

Gobot1234:
Would you mind also adding a small match test case for the feature you have shown in your OP?

tests/test_features.py
```diff
@@ -151,17 +151,18 @@ class Foo(betterproto.Message):
     foo.baz = "test"
 
     # Other oneof fields should now be unset
-    assert foo.bar == 0
+    assert not hasattr(foo, "bar")
+    assert object.__getattribute__(foo, "bar") == betterproto.PLACEHOLDER
     assert betterproto.which_one_of(foo, "group1")[0] == "baz"
 
-    foo.sub.val = 1
```
Gobot1234, line 157: Is this a behaviour change?

a-khabarov: This PR makes it impossible to use `foo.sub.val = 1` when `foo.sub` is unset the group.

a-khabarov: I think this is a bit of a tradeoff. With this change the users of `betterproto` can no longer use the `foo.sub.val = 1` syntax for fields that are unset in groups, but this also means that there is less risk of them accidentally changing which field is set in a group.

a-khabarov:
I've added a test for pattern matching. Sorry for all the commits, I should've installed `pre-commit` locally before pushing.

Gobot1234 (approved):

a-khabarov:
@Gobot1234 Thank you! Would you mind merging this PR? (since I don't have write access)

MicaelJarniac:
I think it'd be nice to have this reflected in the docs.

MicaelJarniac:
I've opened a PR to change the README to reflect this:
- #558

PR #513, opened 2023-07-28
Fix Message equality comparison

### LINKED ISSUE #512: Unconventional `__eq__` method behavior
Possibly due to

https://github.com/danielgtaylor/python-betterproto/blob/6faac1d1ca4b979704c145d968dfed8e38fe159c/src/betterproto/__init__.py#L654

returning `False` instead of `NotImplemented` (see Python docs [here](https://docs.python.org/3/reference/datamodel.html#object.__eq__)), the following snippet doesn't work as expected:

```python
from unittest.mock import ANY
import betterproto

class Foo(betterproto.Message):
    pass

assert Foo() == ANY  # Expected to pass but raises AssertionError
```

This is a minimal reproducible example. The actual use case is related to how [syrupy](https://github.com/tophat/syrupy) and pytest check assertions. The following test crashes with an enigmatic error `KeyError: -1` which is indirectly caused by betterproto's `__eq__` method, which doesn't delegate comparison to the right hand side argument in case of a type mismatch. (Which is what returning `NotImplemented` is for)

```python
# pytest 7.4.0
# syrupy 4.0.8

import betterproto

class Foo(betterproto.Message):
    pass

def test_foo(snapshot):
    foo = Foo()
    assert foo == snapshot
```

coffee-lord:
Fixes #512

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -651,7 +651,7 @@ class Message(ABC):
 
     def __eq__(self, other) -> bool:
         if type(self) is not type(other):
-            return False
+            return NotImplemented
 
         for field_name in self._betterproto.meta_by_field_name:
             self_val = self.__raw_get(field_name)
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -18,6 +18,7 @@ from typing import (
     List,
     Optional,
 )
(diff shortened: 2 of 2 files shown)


Gobot1234 (approved):
Thank you

PR #517, opened 2023-08-06
Improve (de-)serialization performance for scalar arrays

### LINKED ISSUE #515: Improve (de-)serialization performance for scalar arrays
One of my personal use cases for `betterproto` is to call `SciPy` functions from other languages, which do not have such nice math libraries. That is, I have a Python grpc service which essentially is receiving large float arrays, doing math with them, and sending large result arrays in return. Unfortunately, serializing and deserializing (numpy) arrays with `betterproto` does not seem to be as efficient, as it could be.

When serializing a float array to the protobuf format, the serialized protobuf message happens to be exactly in the right byte format to be interpreted as a numpy array, as the following example shows:

```python3
@dataclass
class Array(betterproto.Message):
    values: List[float] = betterproto.double_field(1)

proto_array = Array(values=[1.23, 2.34, 3.45, 4.56])
serialized_array = bytes(proto_array)
np_array = np.frombuffer(serialized_array[2:])

print(np_array)
```

However, when deserializing the protobuf message with `betterproto`, the array is converted into a Python list right away. If I need a numpy array, I have no choice but to convert it back and forth (same for serialization) which is computationally expensive.

I have two ideas for solving that issue:

**Idea 1:** Instead of storing protobuf scalar arrays as Python lists, you could store their byte representation inside a slim wrapper which behaves like a list, but which can also be converted into a numpy array without efforts:

```python3
class Float64Array:
    __data: bytes

    def __len__(self):
        ...
    
    def __getitem__(self, i):
        ...

    def to_numpy_array(self):
        import numpy as np
        return np.frombuffer(self.__data)
```

**Idea 2:** You could introduce an optional protoc compiler flag for letting the caller decide whether scalar arrays should be stored as Python lists or as numpy arrays.

124C41p:
Fixes #515 

Since (de-)serialization is implemented purely in Python, it is quite slow compared to native implementations. I try to circumvent that issue by not deserializing repeated scalar fields immediately, but wrapping their byte representation inside the `ScalarArray[T]` class instead. This class acts like a list. That is, you can call `len(a)`, `a[i]`, and `list(a)` for any ScalarArray `a`, and only at this point we actually deserialize (which is still very slow for big arrays).

On the other hand, when using numpy you can also call `np.asarray(a)` for any ScalarArray `a` to turn it into a numpy array in no time. Conversely, any numpy array `b` can be turned into a `ScalarArray` by calling `ScalarArray.from_numpy(b)` to be passed to a `betterproto` dataclass field (instead of a list) for faster serialization speed.

I tried to be as non-breaking as possible. That is, you can use lists everywhere you used them before. However, it was necessary to generate `Sequence[T]` type hints where `List[T]` hints were generated before. Also note that ScalarArray is an immutable data structure. So you might not be able to use `.append()` or `.insert()` on repeated fields as before (although it should be possible to make `ScalarArray` mutable if really needed).

What do you think about this approach?


Gobot1234:
Superseded by #545

Issue #482, opened 2023-05-02
Support for coded streams


a-khabarov:
C++ and Java implementations of protocol buffers have support for [writing and reading multiple messages](https://protobuf.dev/programming-guides/techniques/#streaming) via `CodedOutputStream` and `CodedInputStream` . This is also space-efficient because the size of each message can be represented as a `varint`. 

I think having something similar to `CodedOutputStream` and `CodedInputStream` in `betterproto` would be useful. One use case would be when a Python tool uses `betterproto` to emit a binary stream of messages using this and then a C++/Java tool can use `CodedInputStream` to read the stream.

Related Java documentation: 
* [CodedInputStream](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/CodedInputStream.html)
* [CodedOutputStream](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/CodedOutputStream)

PR #518, opened 2023-08-16
Add message streaming support

JoshuaLeivers:
Adds support for serializing/deserializing messages and their components to/from streams. Where possible, existing methods now use this functionality internally, minimising code size.

This is useful as a standalone feature, and is also a step towards the functionality requested in #482

As part of this, the following functions have been created/changed:
- `betterproto.dump_varint` - this function encodes a value as a varint and dumps it into the provided stream. It is mostly the same as the existing `betterproto.encode_varint` was, but based around streams and with some additional error checking.

- `betterproto.encode_varint` - this existing function has the same effects, but now uses `betterproto.dump_varint` internally, keeping code size and complexity down.

- `betterproto.size_varint` - this function calculates the size of the varint for a given value without actually serializing it. This may be useful to some, and similar functions exist in the official C++ and other implementations. It is also used internally by other new functionality to reduce memory and time usage compared to simply running `len(...)` after serializing a varint.

- `betterproto._len_preprocessed_single` - calculates the size of the value that would be returned by `_preprocess_single` without fully serializing it. Used internally by new functionality to reduce memory and time usage over simply serializing and then checking the size of it.

- `betterproto._len_single` - similar to above, but for `_serialize_single`.

- `betterproto.load_varint` - loads a varint from a stream and decodes its value. Mostly the same as `decode_varint` already was, but based around streams.

- `betterproto.decode_varint` - existing function, has the same functionality as it did previously, but now uses `load_varint` internally to keep code size and complexity down.

- `betterproto.load_fields` - does the same as `parse_fields`, but loads the fields from the provided stream, rather than a `bytes` object. Used internally by `Message.load`.

- `betterproto.Message.dump` - does the same as `Message.__bytes__` already did, but dumps the results to a stream rather than a `bytes` object.

- `betterproto.Message.__bytes__` - does as it already did, but now uses `Message.dump` internally to reduce code size and complexity.

- `betterproto.Message.__len__` - returns the size of the encoded message - i.e. does the same as `len(bytes(message))` without fully serializing the message, reducing time and memory usage.

- `betterproto.Message.load` - loads and parses a binary encoded message from a stream. Similar to `Message.parse`, but retrieving the data from a stream rather than a `bytes` object.

- `betterproto.Message.parse` - does as it already did, but now uses `Message.load` internally to reduce code size and complexity.


src/betterproto/__init__.py
```diff
@@ -343,20 +343,43 @@ def _pack_fmt(proto_type: str) -> str:
...
+            "Negative value is not representable as a 64-bit integer - unable to encode a varint within 10 bytes."
+        )
+    elif value < 0:
         value += 1 << 64
 
     bits = value & 0x7F
     value >>= 7
     while value:
-        b.append(0x80 | bits)
+        stream.write((0x80 | bits).to_bytes(length=1, byteorder="little"))
```
Gobot1234, line 358:
```suggestion
        stream.write((0x80 | bits).to_bytes(1, "little"))
```
kwargs here are gonna hurt perf and they don't really add much

src/betterproto/__init__.py
```diff
@@ -343,20 +343,43 @@ def _pack_fmt(proto_type: str) -> str:
...
         bits = value & 0x7F
         value >>= 7
-    return bytes(b + [bits])
+    stream.write(bits.to_bytes(length=1, byteorder="little"))
```
Gobot1234, line 361:
```suggestion
    stream.write(bits.to_bytes(1, "little"))
```
Same here

src/betterproto/__init__.py
```diff
@@ -1051,6 +1289,46 @@ def parse(self: T, data: bytes) -> T:
...
+
+        Returns
+        --------
+        :class:`Message`
+            The initialized message.
+        """
+        with BytesIO(data) as stream:
+            self.load(stream)
+
         return self
```
Gobot1234, line 1332:
```suggestion
            return self.load(stream)
```
Might as well

src/betterproto/__init__.py
```diff
@@ -469,22 +552,34 @@ def _dump_float(value: float) -> Union[float, str]:
...
-        pos += 1
-        if not (b & 0x80):
-            return result, pos
-        shift += 7
+    raw = b""
+    for shift in count(0, 7):
         if shift >= 64:
             raise ValueError("Too many bytes when decoding varint.")
+        b = stream.read(1)
+        if b == b"":
```
Gobot1234, line 565:
```suggestion
        if not b:
```

src/betterproto/__init__.py
```diff
@@ -495,7 +590,35 @@ class ParsedField:
     raw: bytes
 
 
-def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
+def load_fields(stream: BinaryIO) -> Iterator[ParsedField]:
```
Gobot1234, line 593:
```suggestion
def load_fields(stream: BinaryIO) -> Generator[ParsedField, None, None]:
```

src/betterproto/__init__.py
```diff
@@ -495,7 +590,35 @@ class ParsedField:
...
+            raw += r
+            raw += decoded
+        elif wire_type == WIRE_FIXED_32:
+            decoded = stream.read(4)
+            raw += decoded
+
+        yield ParsedField(number=number, wire_type=wire_type, value=decoded, raw=raw)
+
+
+def parse_fields(value: bytes) -> Iterator[ParsedField]:
```
Gobot1234, line 621:
```suggestion
def parse_fields(value: bytes) -> Generator[ParsedField, None, None]:
```
Not really sure why these were touched

src/betterproto/__init__.py
JoshuaLeivers: Reverted - this was just due to misreading the docs around this, I believed them to be equivalent here, but they are apparently not

src/betterproto/__init__.py
JoshuaLeivers: Reverted

src/betterproto/__init__.py
```diff
@@ -1051,6 +1289,46 @@ def parse(self: T, data: bytes) -> T:
...
+        if size is not None and read < size:
+            raise ValueError(
+                f"Expected message of size {size}, but was only able to "
+                f"read {read} bytes - the stream may have ended too soon,"
+                " or the expected size may have been incorrect."
+            )
+
+        return self
+
+    def parse(self: T, data: bytes) -> T:
```
Gobot1234, line 1314:
```suggestion
    def parse(self: T, data: ReadableBuffer) -> T:
```
Would it be possible to change this now that it explicitly supports ReadableBuffer (from _typeshed)

JoshuaLeivers: From what I understand, you wouldn't be able to simply change to using this, you would have to conditionally import from `_typeshed` and also conditionally overload/define the function based on `TYPE_CHECKING` - is this what you wanted me to do?

src/betterproto/__init__.py
Gobot1234: They will do basically the same thing to the end user but it's just more annoying if you ever have to go through the blame.

src/betterproto/__init__.py
JoshuaLeivers: I've made the change, though I'm not familiar with parts of this, so let me know if there's anything wrong with it

Gobot1234 (approved):

Gobot1234:
Thanks for all the work on this

JoshuaLeivers:
Hi, was there anything left over on this PR for me to do? Just wanted to check up on how close it is to being merged, or if there's anything preventing it 🙂

PR #525, opened 2023-09-02
server-streaming: remove unreachable code from ServiceBase

omars44:
example output:
```
class UsersBase(ServiceBase):
    async def get_users(
        self, get_users_request: "GetUsersRequest"
    ) -> "GetUsersResponse":
        raise grpclib.GRPCError(grpclib.const.Status.UNIMPLEMENTED)

    async def get_users_stream(
        self, get_users_request: "GetUsersRequest"
    ) -> AsyncIterator["User"]:
        raise grpclib.GRPCError(grpclib.const.Status.UNIMPLEMENTED)
        yield User() # removed this
```


Gobot1234:
This makes the function no longer a generator

PR #529, opened 2023-09-21
Add support for streaming delimited messages

JoshuaLeivers:
This allows developers to easily dump and load multiple messages
from a stream in a way that is compatible with official
protobuf implementations (such as Java's
MessageLite#writeDelimitedTo(...)).

This also adds compatibility tests between the official Java Protobuf
implementation and the new and existing streaming functions in
Betterproto. These tests stream data such as messages to output
files, have a Java binary read them and then write them back using
the protobuf-java functions, and then read them back in on the
Python side to check that the returned data is as expected. This
checks that the official Java implementation (and so any other
matching implementations) can properly parse outputs from
Betterproto, and vice-versa, ensuring compatibility in these
functions between the two.

The Java binary is built on the fly as part of a Pytest fixture, and
these tests are skipped if any necessary tools (i.e. Java or Maven)
are missing from the host machine, or if the build fails (e.g. due
to an incompatible Java version).


JoshuaLeivers:
Sorry about the force-pushes, was just quickly fixing up the CI test fails before it could be reviewed properly, will avoid them from now on 🙂

JoshuaLeivers:
This should finish addressing the message streaming part of #482 , would it be possible to get a review for this?

JoshuaLeivers:
Just rebased onto current `master`

JoshuaLeivers:
Had to resign the commits

cetanu:
Looks reasonable to me, thanks for this work

cetanu (approved):

Issue #526, opened 2023-09-11
Project Tox (poe full-test) setup can break under normal usage


JoshuaLeivers:
Betterproto's Tox config currently calls the system Poetry install for installing dependencies and running Pytest. Under many conditions, this works fine and will correctly use the Tox virtualenvs, but if `poetry env use <environment>` has ever been called for Betterproto, then the system Poetry will enforce the usage of that environment, causing `poe full-test` to run every test using that same virtualenv, defeating the purpose of using it at all.

I don't know if this has affected the CI at all, but this may affect developers locally if they ever call that (which may happen and shouldn't cause this behaviour), and some tools such as JetBrains' PyCharm IDE seem to do this automatically. For that IDE specifically, this means that whenever it is used to open the project, it will break the Tox setup as described above.

As a temporary fix, you can remove the specified environment's section within Poetry's `envs.toml` whenever this occurs and this should go back to the expected behaviour. This isn't ideal, and I can see this causing problems for some developers depending on their setup and what they are doing.

As a more permanent fix, the Tox config would need to be changed to be more hermetic or otherwise avoid this. From what I understand, [the official Poetry documentation says](https://python-poetry.org/docs/faq/#is-tox-supported) to do exactly what the project is already doing, but clearly this doesn't work entirely as expected. A possible fix might be to use [tox-poetry-installer](https://github.com/enpaul/tox-poetry-installer/), though it is currently in beta overall and may still have some bugs. I am looking into whether the tool would work for Betterproto, and will add comments here on what I find.

JoshuaLeivers:
`tox-poetry-installer` combined with Tox config changes seems to prevent this issue, but it requires Python >= 3.8, meaning that adding it would require that _this_ project also does. I don't personally have an issue with this, as Python 3.7 is EoL, but I know others may still be using it, and I wasn't sure whether the project was okay with removing 3.7 compatibility in exchange for this fix.

PR #531, opened 2023-09-22
Fix Tox to use Poetry hermetically (#526)

JoshuaLeivers:
With the existing setup, Tox could sometimes end up using the Poetry virtualenv for all test runs, rather than isolated environments. This meant that only the developer's Python version would be used, defeating the point of the Tox setup. Using `tox-poetry-installer`, the Tox runs can install the Poetry dependencies hermetically.


JoshuaLeivers:
It looks like the Python version doesn't need to change as I mentioned in the issue, as long as `tox-poetry-installer` is listed as a dependency in `tox.ini` specifically, due to how Tox deals with setting up environments. As long as the version you are using to operate Poetry/Tox itself when running it isn't <=3.7, it should work fine.

JoshuaLeivers:
This change fixes #526 . Would it be possible to get this reviewed?

cetanu:
Looking now, thanks btw

cetanu (dismissed):

cetanu:
Looks like there are some conflicts and I'm not able to sort them out

JoshuaLeivers:
Sorry about the force-pushes - I've just rebased this onto `master` and fixed the conflicts.

JoshuaLeivers:
Did you want me to rebase again, or did you want to do it once any PRs have been merged that were planned? I've set it to allow edits by maintainers, but I'm not sure if that allows you to do that or not.

Gobot1234:
Your PR should be the last thing to touch the deps for a while so please could you rebase again.

JoshuaLeivers:
Rebased.

JoshuaLeivers:
Ah, will quickly have to re-sign the commits again

Gobot1234 (dismissed):

JoshuaLeivers:
I believe can do that if you want, though doing it this way means you can use the `pre-commit` hook to format it and it's a bit easier to deal with in text editors/IDEs. I haven't tested if the legacy key works with `tox-poetry-installer` though, so I'm not 100% sure if it does work or not, and couldn't find any mention online of the combination.

JoshuaLeivers:
Did you want me to try and change the config back to being within `legacy_tox_ini`, or am I good to leave it as-is?

Gobot1234:
If you wouldn't mind giving it a go, that'd be nice. I'd like to keep the number of config files down

JoshuaLeivers:
It seems to work fine (though isn't supported by the `tox-ini-fmt` `pre-commit` hook, so I'll have to remove that). Should I keep the default testing environments as they are, or would it be worth possibly replacing Python 3.11 with Python 3.12? For that matter, is 3.12 planned to be enabled in the CI pipeline?

JoshuaLeivers:
Linking an open issue about the `pre-commit` hook's support for this tox-dev/tox-ini-fmt#128

Gobot1234 (approved):

PR #532, opened 2023-09-30
Fix Comment Parsing

cartesian-theatrics:
This change doesn't seem to break anything. I'm not sure why it was doing word wrapping. My guess is it is to make sure to pass linters for existing proto specs with lots of comments. I think better would be to disable word-wrapping by default and if necessary enabled with a flag of some kind.


Gobot1234:
I don't really agree with this change. But I don't have a particularly great way to solve it without a native cli.

isaackogan:
forgive the naivety, but why not use raw-strings?

Gobot1234 (approved):
I've thought about this a bit more, LGTM

PR #484, opened 2023-05-09
Implementing pickle methods

kevinaud:
Fixes #419

PR #535, opened 2023-10-13
Implementing pickle methods

### LINKED ISSUE #419: Pickling a Betterproto message
Hi,

I have found an issue when I wanted to use this library in combination with multiprocessing/pickling.
I am using version 2.0.0b5

When the resulting message is send back by one of my running processes with the `map` function it changes empty values and also when I only run pickle they are changed.

The message:
```
syntax = "proto3";

message TestMessage {
    string value = 1;
}
```

To code:
```
test_message = TestMessage()
print(test_message)
pickled = pickle.dumps(test_message)
unpickled = pickle.loads(pickled)
print(pickle.loads(pickled))
print(unpickled.value)
print(type(unpickled.value))
```

The output:
```
TestMessage()
TestMessage(value=<object object at 0x7febf0791fe0>)
<object object at 0x7fd9ef242ff0>
<class 'object'>
```

So it seems it adds an object instead of an empty value.

cetanu:
Continuing work from #484 by @kevinaud to fix #419


src/betterproto/__init__.py
```diff
@@ -1124,6 +1132,15 @@ def SerializeToString(self: T) -> bytes:
...
         return bytes(self)
 
+    def __getstate__(self) -> bytes:
+        return bytes(self)
+
+    def __setstate__(self: T, pickled_bytes: bytes) -> T:
+        return self.parse(pickled_bytes)
+
+    def __reduce__(self) -> Union[str, Tuple[Any, ...]]:
+        return (self.__class__.FromString, (bytes(self),))
```
Gobot1234, line 1142:
```suggestion
    def __reduce__(self) -> Tuple[Any, ...]:
        return (self.__class__.parse, (bytes(self),))
```

src/betterproto/__init__.py
```diff
@@ -889,6 +889,14 @@ def __deepcopy__(self: T, _: Any = {}) -> T:
                 kwargs[name] = deepcopy(value)
         return self.__class__(**kwargs)  # type: ignore
 
+    def __copy__(self: T, _: Any = {}) -> T:
+        kwargs = {}
+        for name in self._betterproto.sorted_field_names:
+            value = self.__raw_get(name)
+            if value is not PLACEHOLDER:
+                kwargs[name] = value
+        return self.__class__(**kwargs)  # type: ignore
```
Gobot1234, line 898: Can you add some copy tests?

src/betterproto/__init__.py
cetanu: I don't think I can do this, because parse is a bound method
```
Could not bind method "parse" because "Type[Message]" is not assignable to parameter "self"
```
If I use parse, the pickling tests fail with 
```
TypeError: Message.parse() missing 1 required positional argument: 'data'
```
Not really sure what to provide for the self argument

src/betterproto/__init__.py
cetanu: there appears to be one already, I copied it and added pickling as well:

```python
﻿def test_pickle_copyability():
    msg = PickleMessage(bar=12, baz=["hello"])
    deserialized = pickle.loads(pickle.dumps(msg))

    copied = copy(msg)
    assert msg == copied == deserialized
    assert msg is not copied is not deserialized
    assert msg.baz is copied.baz #is deserialized.baz

    deepcopied = deepcopy(msg)
    assert msg == deepcopied == deserialized
    assert msg is not deepcopied is not deserialized
    assert msg.baz is not deepcopied.baz is not deserialized.baz
```
However the commented assert seems to evaluate false... I think that's correct?? Since pickling and unpickling would technically mean the field is now a different object

cetanu:
Working on a more complicated test, which is now raising some DX questions

tests/test_pickling.py
```diff
@@ -0,0 +1,172 @@
...
+        mapping={
+            "message": google.Any(value=bytes(Fi(abc="hi"))),
+            "string": google.Any(value=b"howdy"),
+        },
+    )
+    deser = unpickled(msg)
+    assert msg == deser
+    assert msg.fe.abc == "1"
+    assert msg.is_set("fi") is not True
+    assert msg.mapping["message"] == google.Any(value=bytes(Fi(abc="hi")))
```
cetanu, line 87: See this assertion and the following two lines

Should fields be perfectly preserved in this manner? My gut instinct is yes... but it may be annoying for users to extract and decode these values. It is probably not the job of this library to take care of that.

src/betterproto/__init__.py
Gobot1234: Got it, thanks

tests/test_pickling.py
Gobot1234: I think it's maybe something to worry about after Any support, for now this seems fine

src/betterproto/__init__.py
Gobot1234: Can the lines in https://github.com/danielgtaylor/python-betterproto/blob/6b36b9ba9ffda8a7b3b488b774729c834db8ae0d/tests/test_features.py#L702-L718 be moved with the new tests?

cetanu: Done

I also moved across the recursive message tests as they now are pickled before testing as well

Gobot1234 (approved):
LGTM

PR #540, opened 2023-10-28
Make everything typecheck?

Gobot1234:
## Summary

Use more 3.7+ features in the library and make everything type check

## Checklist

- [ ] This PR fixes an issue.


Gobot1234:
No idea where this pkg_resources failure came from

cetanu (approved):

Gobot1234:
@cetanu do you have any idea why the tests are failing?

cetanu:
Looking now...

cetanu:
It appears to be related to this

https://github.com/grpc/grpc/issues/33570

but it's just a deprecation warning? I think we can catch this

cetanu:
https://github.com/grpc/grpc/blob/master/tools/distrib/python/grpcio_tools/grpc_tools/protoc.py#L21

cetanu:
Since this is only for generating the tests, we could do something very hacky

```python
    import re
    from importlib import resources

    file = resources.files("grpc_tools") / Path("protoc.py")
    content = file.read_text()

    if "import pkg_resources" in content:
        content = re.sub(
            pattern="import pkg_resources",
            repl=r"try:\n    import pkg_resources\nexcept DeprecationWarning:\n    pass",
            string=content,
        )

    file.open().write(content)
```

Gobot1234:
I'll just send a PR to fix it

tzahush:
@Gobot1234 I see the PR that we were waiting for here https://github.com/grpc/grpc/pull/35329 is closed, what should we do as next steps? perhaps this was fixed in different PR's.

Gobot1234:
Note to self should try and get more places to use positional only params

toby-bro:
mypy still does not manage to type check this correctly, it seems related to [this missing feature](https://github.com/python/mypy/issues/12445).
That is supposedly corrected by : https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime 
I might open a PR for this in the future...

PR #545, opened 2023-12-04
Use external package `betterproto-rust-codec` for better (de-)serialization performance

124C41p:
This PR is an alternative to #520. It uses the same Rust based extension module to improve (de-)serialization performance of `betterproto` significantly. However, instead of integrating the extension into the `betterproto` project, I published it to PyPi separately, and only referenced it here as an optional dependency. For this reason, this branch is ready to use as it is.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1014,13 +1014,6 @@ class Message(ABC):
         """
         Get the binary encoded Protobuf representation of this message instance.
         """
-
-        try:
-            import betterproto_rust_codec
-            return betterproto_rust_codec.serialize(self)
-        except ModuleNotFoundError:
-            pass
-
         with BytesIO() as stream:
             self.dump(stream)
             return stream.getvalue()
@@ -1363,14 +1356,6 @@ class Message(ABC):
         :class:`Message`
             The initialized message.
         """
-        
-        try:
-            import betterproto_rust_codec
-            betterproto_rust_codec.deserialize(self, data)
-            return self
-        except ModuleNotFoundError:
-            pass
-    
         with BytesIO(data) as stream:
             return self.load(stream)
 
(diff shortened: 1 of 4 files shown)


Gobot1234 (changes requested):
As discussed on slack, I think I'd prefer this to patch the parse function or conditionally define it whichever works better for you

cetanu (approved):
Looks okay to me, will only be used if the optional package is installed, correct?

cetanu (commented):
Actually, this introduces two new try/except for every call to __bytes__, is that right?

src/betterproto/__init__.py
```diff
@@ -1868,6 +1868,23 @@ def _validate_field_groups(cls, values):
...
+        betterproto_rust_codec.deserialize(self, data)
+        return self
+
+    def __bytes_patch(self) -> bytes:
+        return betterproto_rust_codec.serialize(self)
+
+    Message.parse = __parse_patch
+    Message.__bytes__ = __bytes_patch
+except ModuleNotFoundError:
+    pass
```
Gobot1234, line 1886: I'd personally just do something along the lines of
```py
def parse():...
def __bytes__():...
if importlib.util.find_spec("betterproto_rust_codec"):
    import betterproto_rust_codec
	parse = betterproto_rust_codec.deserialize  # edit this so it returns self
	__bytes__ = betterproto_rust_codec.serialize
```

124C41p: I don't really understand why, but directly assigning `__bytes__ = betterproto_rust_codec.serialize` does not work for some reason. So editing `deserialize` would probably don't help either.

What do you think about this approach?

```python3

class Message:
    if HAS_ACCELERATOR:
        def parse(self, data):
            ...
    else:
        def parse(self, data):
            ...
```

I think this might be even more transparent.

Gobot1234: I find that problematic because it's another level of indentation which is annoying especially with black. If the function isn't binding as a descriptor just leave it as is

Gobot1234 (approved):
Thanks LGTM!

Issue #549, opened 2023-12-20
`Struct().from_dict(...)` has no effect


woodruffw:
### Summary

`protobuf.Struct` appears to not load correctly with `from_dict(...)`, which also affects `from_json(...)`

### Reproduction Steps

I originally saw this with a more complex message, however I've minified it down to a single API call.

Using `betterproto==2.0.0b6`:

```python
>>> import betterproto.lib.google.protobuf as protobuf
>>> protobuf.Struct().from_dict({"a": "b"})
```

### Expected Results

I expect a `Struct` instance populated from the dictionary:

```python
Struct(a="b")
```

### Actual Results

I receive an empty `Struct` instance:

```python
Struct()
```

### System Information

```
libprotoc 25.1
Python 3.12.0
Name: betterproto
Version: 2.0.0b6
Summary: A better Protobuf / gRPC generator & library
Home-page: https://github.com/danielgtaylor/python-betterproto
Author: Daniel G. Taylor
Author-email: danielgtaylor@gmail.com
License: MIT
Location: /Users/william/devel/attestation/python/env/lib/python3.12/site-packages
Requires: grpclib, python-dateutil
Required-by: in-toto-attestation
```

### Checklist

- [X] I have searched the issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have verified this issue occurs on the latest prelease of betterproto which can be installed using `pip install -U --pre betterproto`, if possible.

Gobot1234:
Duplicate of #332

woodruffw:
Correct me if I'm wrong, but I don't believe these are duplicates: the linked issue says that the types aren't supported, while they *do* appear supported but have buggy behavior in the dictionary/JSON conversion cases.

woodruffw:
The follow-up comment on the linked issue is also about a codegen miscompilation, which is not the observed behavior here.

Gobot1234:
We need to have special behaviour for the classes for which there is currently none, this is something else that requires special behaviour, the follow up comment doesn't necessarily mean that it's the same problem as the original issue

woodruffw:
Okay, thanks for clarifying!

PR #551, opened 2023-12-27
betterproto: support `Struct` and `Value`

### LINKED ISSUE #332: Struct, Value well know type not supported
https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/struct

These should map to `dict[str, any]` and `any`

woodruffw:
## Summary

Fixes #332. See also #549.

~~Opening as a draft until I add tests.~~

## Checklist

- [x] This PR fixes an issue.
    - [x] This change has an associated test.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -55,7 +55,6 @@ from .utils import (
     hybridmethod,
 )
 
-
 if TYPE_CHECKING:
     from _typeshed import (
         SupportsRead,
@@ -1522,6 +1521,12 @@ class Message(ABC):
 
     @classmethod
     def _from_dict_init(cls, mapping: Mapping[str, Any]) -> Mapping[str, Any]:
+        # Special case: google.protobuf.Struct has a single fields member but
+        # behaves like a transparent JSON object, so it needs to first be munged
+        # into `{fields: mapping}`.
+        if cls == Struct:
+            mapping = {"fields": mapping}
+
         init_kwargs: Dict[str, Any] = {}
         for key, value in mapping.items():
             field_name = safe_snake_case(key)
@@ -1552,7 +1557,7 @@ class Message(ABC):
                         if isinstance(value, list)
                         else sub_cls.from_dict(value)
                     )
-            elif meta.map_types and meta.map_types[1] == TYPE_MESSAGE:
+            elif meta.map_types and meta.map_types[1] == TYPE_MESSAGE and cls != Struct:
                 sub_cls = cls._betterproto.cls_by_field[f"{field_name}.value"]
                 value = {k: sub_cls.from_dict(v) for k, v in value.items()}
             else:
@@ -1582,6 +1587,7 @@ class Message(ABC):
                     )
 
             init_kwargs[field_name] = value
+
         return init_kwargs
 
     @hybridmethod
@@ -1926,6 +1932,7 @@ from .lib.google.protobuf import (  # noqa
     Int32Value,
     Int64Value,
     StringValue,
+    Struct,
     Timestamp,
     UInt32Value,
     UInt64Value,
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1582,7 +1582,6 @@ class Message(ABC):
                     )
 
             init_kwargs[field_name] = value
-
         return init_kwargs
 
     @hybridmethod
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1403,15 +1403,6 @@ class Message(ABC):
         Dict[:class:`str`, Any]
             The JSON serializable dict representation of this object.
         """
-        # Mirror of from_dict: Struct's `fields` member is transparently
-        # dispatched through instead.
-        if isinstance(self, Struct):
-            output = {**self.fields}
-            for k in self.fields:
-                if hasattr(self.fields[k], "to_dict"):
-                    output[k] = self.fields[k].to_dict(casing, include_default_values)
-            return output
-
         output: Dict[str, Any] = {}
         field_types = self._type_hints()
         defaults = self._betterproto.default_gen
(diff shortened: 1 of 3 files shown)


woodruffw:
NB: I haven't made any documentation changes, since I believe this behavior was intended (in terms of Protobuf's specified behavior) but just buggy before. However, if there's somewhere I should note this (like the CHANGELOG?) let me know and I'm more than happy to add it 🙂 

Similarly, in terms of semver, I believe this is considered a nonbreaking change (since it's a bugfix).

src/betterproto/__init__.py
```diff
@@ -1404,6 +1403,15 @@ def to_dict(
...
             The JSON serializable dict representation of this object.
         """
+        # Mirror of from_dict: Struct's `fields` member is transparently
+        # dispatched through instead.
+        if isinstance(self, Struct):
+            output = {**self.fields}
+            for k in self.fields:
+                if hasattr(self.fields[k], "to_dict"):
+                    output[k] = self.fields[k].to_dict(casing, include_default_values)
+            return output
```
woodruffw, line 1413: This special-casing is arguably pretty ugly, but I couldn't think of a better (clean) way to accomplish this: `Struct` *is* kind of special in protobuf, in that it's a transparent mapping type (when expressed as JSON) where no other types really are. As such, it doesn't really fit into the `to_dict`/`from_dict` behavior below.

Gobot1234:
Honestly I think just putting this behaviour on the type itself would be the best solution. The generated file has been pretty stable for a long time so I don't really have a problem just tacking this on at this point.

woodruffw:
> Thanks for this! Honestly I think just putting this behaviour on the type itself would be the best solution. The generated file has been pretty stable for a long time so I don't really have a problem just tacking this on at this point.

No problem! Making sure I understand: by "on the type itself," you mean specialize `to_dict()` and `from_dict()` on `Struct` itself?

If so, makes sense -- I can do that in a moment.

Gobot1234:
> > Thanks for this! Honestly I think just putting this behaviour on the type itself would be the best solution. The generated file has been pretty stable for a long time so I don't really have a problem just tacking this on at this point.
> 
> No problem! Making sure I understand: by "on the type itself," you mean specialize `to_dict()` and `from_dict()` on `Struct` itself?
> 
> If so, makes sense -- I can do that in a moment.

Yep exactly

woodruffw:
Alright, moved the specialization to `Struct` itself!

woodruffw:
Gentle ping; please let me know if there's anything else I can do here!

src/betterproto/__init__.py
```diff
@@ -1582,6 +1582,7 @@ def _from_dict_init(cls, mapping: Mapping[str, Any]) -> Mapping[str, Any]:
                     )
 
             init_kwargs[field_name] = value
+
```
Gobot1234, line 1585:
```suggestion

```

Gobot1234 (dismissed):

Gobot1234 (approved):

Gobot1234:
Test failures seem unrelated

woodruffw:
> Test failures seem unrelated

Yeah, I think so -- looks like the failure is in `Message._betterproto`, which then fails to generate `cls._betterproto_meta`. But I can't see anything in these changes that would cause that.

woodruffw:
Triaging a bit more: `_get_field_default_gen` has an `issubclass` check that fails because `t` is not a `type` object.

This suggests that something about type hints is hinky, since `t` is resolved from them. Maybe a deferred/non-deferred annotations thing, causing `t` to be a `str` instead?

src/betterproto/lib/google/protobuf/__init__.py
```diff
@@ -2,14 +2,21 @@
 # sources: google/protobuf/any.proto, google/protobuf/api.proto, google/protobuf/descriptor.proto, google/protobuf/duration.proto, google/protobuf/empty.proto, google/protobuf/field_mask.proto, google/protobuf/source_context.proto, google/protobuf/struct.proto, google/protobuf/timestamp.proto, google/protobuf/type.proto, google/protobuf/wrappers.proto
 # plugin: python-betterproto
 # This file has been @generated
+
+from __future__ import annotations
```
woodruffw, line 6: Ah yep, this is probably doing it.

woodruffw:
(Also, if I can make an additional request: a new beta release after this lands would be greatly appreciated!)

Gobot1234:
Can't do a beta release till https://github.com/grpc/grpc/pull/35329 is merged and https://github.com/danielgtaylor/python-betterproto/pull/540

Gobot1234 (approved):
Thanks

atomicmac:
@woodruffw @Gobot1234 Is there anything special that needs to be done for serialization/deserialization of the Structs? From what I'm seeing, instantiating a Struct as exemplified in the tests and then attempting to send it over the wire fails because the Struct's dump method is eventually called, which falls back to the Message.dump method, which treats the struct as a dict, which expects all keys to be of type 'string' and all values to be of type 'message' - if the value is an intrinsic type like str, the catch-all bytes(value) call fails in _preprocess_single because of TypeError('string argument without an encoding')

Gobot1234:
Oh it seems like it might need special casing for dump yeah

woodruffw:
Yep, thanks for catching that @atomicmac, and sorry for the gap there. I can look into a fix tomorrow.

woodruffw:
I've started work on this in #559, but I'm a little stumped on how to actually specialize `dump()` here -- from a quick look `Struct` should be transparently serialized as `Map<string, Value>` in the binary encoding, but I'm not sure what the most idiomatic/straightforward way to do that is.

atomicmac:
Thanks for looking @Gobot1234 & @woodruffw ! Yes @woodruffw I was looking at implementing the specialized dump() and came to the same crossroads. It would be nice to not have to wrap every value in the struct, but then autoconverting to protobuf types becomes involved if you want to keep transmitted sizes as small as possible.

woodruffw:
> Yes @woodruffw I was looking at implementing the specialized dump() and came to the same crossroads. It would be nice to not have to wrap every value in the struct, but then autoconverting to protobuf types becomes involved if you want to keep transmitted sizes as small as possible.

Glad to hear it's not just me 😅 -- I've been thinking about this a bit, and I still don't have a great idea here.

PR #559, opened 2024-02-25
tests: add an encoding roundtrip for Struct

woodruffw:
## Summary

Follows #551. This currently fails, as expected, since `dump()` is not correctly specialized for `Struct`.

This is a work in progress.

## Checklist

- [x] This PR fixes an issue.
    - [x] This change has an associated test.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -55,7 +55,6 @@ from .utils import (
     hybridmethod,
 )
 
-
 if TYPE_CHECKING:
     from _typeshed import (
         SupportsRead,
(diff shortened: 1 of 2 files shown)


woodruffw:
Just to post a brief update: per https://github.com/danielgtaylor/python-betterproto/pull/551#issuecomment-1972061214 and below, I don't yet have a great idea for a patch here. The basic thing that needs to happen is that `dump()` and `load()` need to be specialized for `Struct`, but doing so is weird because `Struct` is not referentially transparent against the underlying `map<string, ...>` type it contains.

PR #560, opened 2024-02-29
Use identity check with PLACEHOLDER instead of equality test

atomicmac:
## Summary
PLACEHOLDER is a specific instance of a sentinel object, the test here should be identity instead of equality, i.e. "is not" instead of "!="
I am experimenting with adding ndarray support, and the equality test here causes problems.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -749,7 +749,7 @@ class Message(ABC):
                 group_current.setdefault(meta.group)
 
             value = self.__raw_get(field_name)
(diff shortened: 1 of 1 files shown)


Gobot1234 (approved):

Gobot1234:
FWIW ndarray support has been rejected recently in favour of a rust based deserialiser. See https://github.com/danielgtaylor/python-betterproto/pull/517

PR #566, opened 2024-03-27
Add betterproto.Enum __copy__ and __deepcopy__ implementations

atomicmac:
## Summary

betterproto.Enum is missing __copy__ and __deepcopy__ implementations, which were recently added to enum.Enum, see https://github.com/python/cpython/issues/106602 This fixes the bug where betterproto messages with Enums nested within cannot be copied via copy.deepcopy.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/src/betterproto/enum.py
+++ b/src/betterproto/enum.py
@@ -159,7 +159,7 @@ class Enum(IntEnum if TYPE_CHECKING else int, metaclass=EnumType):
     def __copy__(self) -> Self:
         return self
 
-    def __deepcopy__(self, memo):
+    def __deepcopy__(self, memo: Any) -> Self:
         return self
 
     @classmethod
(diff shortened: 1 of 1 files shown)


src/betterproto/enum.py
```diff
@@ -156,6 +156,12 @@ def __delattr__(self, item: Any) -> Never:
             f"{self.__class__.__name__} Cannot delete a member's attributes."
         )
 
+    def __copy__(self):
```
Gobot1234, line 159:
```suggestion
    def __copy__(self) -> Self:
```

src/betterproto/enum.py
```diff
@@ -156,6 +156,12 @@ def __delattr__(self, item: Any) -> Never:
...
+        return self
+
+    def __deepcopy__(self, memo):
```
Gobot1234, line 162:
```suggestion
    def __deepcopy__(self, memo: Any) -> Self:
```

Gobot1234 (changes requested):
Thanks for this

PR #568, opened 2024-04-04
fix: support pydantic-version google pb

ii64:
## Summary

When using pydantic_dataclass option, the code-genererated stub still using non-pydantic version of dataclass, since we are not updating the refs, we blow up stack due to recursion.

## Checklist

- [x ] If code changes were made then they have been tested.
    - [x ] I have updated the documentation to reflect the changes.
- [x ] This PR fixes an issue.


Gobot1234:
This is a breaking change for no real reason. Please can you keep the default lib.google.protobuf as the standard dataclasses version

ii64:
That's the reason why I keep `lib.google.protobuf` and `lib.google.protobuf.compiler` - so that it won't break any existing generated code.

- `lib.google.protobuf` will import `lib.std.google.protobuf`
- `lib.google.protobuf.compiler` will import `lib.std.google.protobuf.compiler`

While pydantic one is just an extra add-on for any `pydantic_dataclass` generated stub

gen.sh
Gobot1234, line 1: I don't want to add script files because they're very temperamental please switch it to a poe command

Gobot1234 (changes requested):
Please can you delete the `@generated` from the top of these files and remind people to include changes from the history (e.g. https://github.com/ii64/python-betterproto/commit/5666393f9d10e13609d8eeac8d1ab3815dce5fd6 which I can see you've included)

gen.sh
ii64: Gothca

ii64:
Removed `@generated` and fix codefmt/linter related error. CI all green @Gobot1234

Gobot1234:
Thanks! One last thing, can you add a test for this situation so it doesn't happen again.

ii64:
Could you try to check and run the CI again? CI is green on my end @Gobot1234

Gobot1234 (approved):

Gobot1234:
Thank you!

PR #576, opened 2024-05-29
Output INT_64_TYPES as int instead of string in to_dict

ihoegen:
## Summary

Previous behavior was int64 fields would be represented as strings when `to_dict` was called, when they should really be represented as ints, at least on 64 bit machines. 

Short term fix is to use int32 instead
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1477,12 +1477,12 @@ class Message(ABC):
             ):
                 if meta.proto_type in INT_64_TYPES:
                     if field_is_repeated:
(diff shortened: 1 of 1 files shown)


Gobot1234:
This behaviour is to match the protobuf behaviour. Have they changed it? 

FWIW if you want a more useful type in python you should use to_pydict anyway

ihoegen:
Actually wasn't familiar with the protobuf behavior. `to_pydict` makes more sense though

PR #587, opened 2024-07-16
Drop 3.7 from ci

imcdo:
Currently, CI is broken for 3.7 as github runs its macos images on arm and doesn't provide arm images for 3.7.  As 3.7 is deprecated the best (at least interim) solution would be to drop 3.7 checks entirely.

## Checklist

- [x] This PR fixes an issue.


Gobot1234:
Seems like pydantic has broken on 3.12

imcdo:
https://github.com/pydantic/pydantic/issues/9637#issuecomment-2162873917 seems like it was fixed in 1.10.16

Gobot1234:
I think this needs relocking

cetanu (approved):

PR #588, opened 2024-07-18
Update to Pydantic V2

### LINKED ISSUE #530: pydantic v2 support
Hi,

Thanks for the great job.
I'm just wondering whether pydantic v2 support is on the way, seems using `root_validator` is raising error in v2.

like `langchain` use a [submodule import](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/pydantic_v1/__init__.py) .

gatesn:
## Summary

Updates betterproto to generate data classes for Pydantic V2. 

The issues I ran into were:
* `root_validator` is deprecated
  * Replaced by `model_validate(mode='after')`, required updating the field validator to use getattr on the built model
* `betterproto.Enum` was not recognized by Pydantic as an enum type.
  * I don't have the history for why this project doesn't use IntEnum, but I worked around this by creating a Pydantic serializer that will validate the integer is within the correct range.
  * An alternate solution here would just be to include the `__get_pydantic_core_schema__` function on every generated Enum class. Open to switching to this.
* Pydantic no longer raises Value error when constructing a data class with a missing argument. 
  * I think this is ok since it now better matches the builtin behavior, I changed the test to allow this.
  * I added a test demonstrating that Pydantic will validate the correct type.

Fixes #530 

## Checklist

    - [X] I have updated the documentation to reflect the changes.
- [X] This PR fixes an issue.
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)


gatesn:
I inlined the Pydantic serializer into every generated enum since `typing.Annotated` only exists in Python >=3.9

Gobot1234:
Fwiw if the code is easier to generate/understand use typing_extensions.Annotated. It is already a dep of the project and probably pydantic

gatesn:
I don't mind either way, your call

gatesn:
I just discovered the `poe format` task 🤦

Gobot1234:
> I don't mind either way, your call

code looks fine to me as is :)

cetanu:
relock and the checks should go green

gatesn:
Should be fixed

gatesn:
@Gobot1234 I've resolved conflicts against develop. Is this g2g now?

src/betterproto/templates/header.py.j2
```diff
@@ -12,11 +12,11 @@ from typing import TYPE_CHECKING
...
 
 if TYPE_CHECKING:
-    from dataclasses import dataclass
+    import dataclasses
 else:
-    from pydantic.dataclasses import dataclass
+    from pydantic import dataclasses
 {%- else -%}
-from dataclasses import dataclass
+import dataclasses
```
Gobot1234, line 19: I'd like to keep this the same for now please

Gobot1234:
CI looks good :)

gatesn:
Reverted the import change

src/betterproto/templates/template.py.j2
```diff
@@ -12,11 +12,18 @@ class {{ enum.py_name }}(betterproto.Enum):
...
+        from pydantic_core import core_schema
+
+        return core_schema.int_schema(ge=0)
+    {% endif %}
 
 {% endfor %}
 {% endif %}
 {% for message in output_file.messages %}
-@dataclass(eq=False, repr=False)
+@dataclasses.dataclass(eq=False, repr=False)
```
Gobot1234, line 26: This needs reverting

gatesn: Yup, just caught that, apologies

gatesn:
@Gobot1234 are we all set? Keen to start getting some model validation in my protos 😄

Gobot1234:
Yep, thanks for this!

Gobot1234 (approved):

gatesn:
@Gobot1234 would you be open to publishing a v2.0.0-beta7?

gatesn:
Actually it looks like there’s some weird behaviour with the PLACEHOLDER object. Should that be replaced with default primitive value for the given type? Not sure…

src/betterproto/lib/pydantic/google/protobuf/__init__.py
```diff
@@ -1,9 +1,16 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # sources: google/protobuf/any.proto, google/protobuf/api.proto, google/protobuf/descriptor.proto, google/protobuf/duration.proto, google/protobuf/empty.proto, google/protobuf/field_mask.proto, google/protobuf/source_context.proto, google/protobuf/struct.proto, google/protobuf/timestamp.proto, google/protobuf/type.proto, google/protobuf/wrappers.proto
 # plugin: python-betterproto
-
+# This file has been @generated
```
JitPackJoyride, line 4: @Gobot1234 I thought you specifically said you didn't want this @generated label on these files during this PR: https://github.com/danielgtaylor/python-betterproto/pull/568.

I'm opening a PR for a CHANGELOG and bump in version, and wondering if you think this is blocking.

Gobot1234: Ah yep good catch, would you mind removing it in your PR so it's easier to merge?

PR #589, opened 2024-08-04
Add prefixed enums in addition to pretty ones

snikch:
## Summary

The representation of enums in JSON should not have the prefix removed as implemented in #187.  This PR provides one option for fixing JSON representation of enums.

From the linked [C# docs](https://protobuf.dev/reference/csharp/csharp-generated/#enum) in https://github.com/danielgtaylor/python-betterproto/issues/174 (referenced in the above PR):

> This name transformation does not affect the text used within the JSON representation of messages.

This fix is a somewhat naive approach and I'm open to alternatives. It simply adds in the original enum field in addition to the stripped field.

I looked at an alternative where the JSON would render based on a `full_name` property, however the scope of this started to grow so I figured I'd push up the naive change and start a discussion.

One option would be to replace the value with a tuple of some sort, however it's possible this would create issues with comparisons.

```py
class ArithmeticOperator(betterproto.Enum):
    """
    A "C" like enum with the enum name prefixed onto members, these should be stripped
    """

    NONE = (0, 'ARITHMETIC_OPERATOR_NONE')
    PLUS = (1, 'ARITHMETIC_OPERATOR_PLUS')
    MINUS = (2, 'ARITHMETIC_OPERATOR_MINUS')
    _0_PREFIXED = (3, 'ARITHMETIC_OPERATOR_0_PREFIXED')
```

Another option I thought of was to add an option as to whether to remove the prefix, however I personally would like to have the best of both words - nice python enums along with correct JSON encoding / decoding.

Thoughts?

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1576,10 +1576,13 @@ class Message(ABC):
                     )
                 elif meta.proto_type == TYPE_ENUM:
                     enum_cls = cls._betterproto.cls_by_field[field_name]
+
+                    def obj(enum_class, value):
+                        return enum_class.from_full_name(value) if hasattr(enum_class, 'from_full_name') else enum_class.from_string(value)
                     if isinstance(value, list):
-                        value = [enum_cls.from_string(e) for e in value]
+                        value = [obj(enum_cls, e) for e in value]
                     elif isinstance(value, str):
-                        value = enum_cls.from_string(value)
+                        value = obj(enum_cls, value)
                 elif meta.proto_type in (TYPE_FLOAT, TYPE_DOUBLE):
                     value = (
                         [_parse_float(n) for n in value]
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1493,24 +1493,30 @@ class Message(ABC):
                     else:
                         output[cased_name] = b64encode(value).decode("utf8")
                 elif meta.proto_type == TYPE_ENUM:
+                    def name(enum_class, value):
+                        obj = enum_class(value)
+                        if hasattr(obj.__class__, 'full_name') and isinstance(obj.__class__.full_name, property):
+                            return obj.full_name
+                        return obj.name
                     if field_is_repeated:
                         enum_class = field_types[field_name].__args__[0]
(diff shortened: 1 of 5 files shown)


Gobot1234:
The full name version seems desirable. Could conditionally add the dict based on if it's necessary. The other options in my mind don't play nicely with type checking.

snikch:
@Gobot1234 I went down a rabbit hole with custom enum types but ended up doing as you suggested. Updated!

src/betterproto/__init__.py
```diff
@@ -1493,24 +1493,30 @@ def to_dict(
                     else:
                         output[cased_name] = b64encode(value).decode("utf8")
                 elif meta.proto_type == TYPE_ENUM:
+                    def name(enum_class, value):
+                        obj = enum_class(value)
+                        if hasattr(obj.__class__, 'full_name') and isinstance(obj.__class__.full_name, property):
+                            return obj.full_name
+                        return obj.name
```
Gobot1234, line 1500:
```suggestion
                    def name(enum_class, value):
                        obj = enum_class(value)
                        if hasattr(obj.__class__, 'full_name') and isinstance(obj.__class__.full_name, property):
                            return obj.full_name
                        return obj.name
```
This should probably go in utils or enum.py

src/betterproto/__init__.py
```diff
@@ -1570,10 +1576,13 @@ def _from_dict_init(cls, mapping: Mapping[str, Any]) -> Mapping[str, Any]:
                     )
                 elif meta.proto_type == TYPE_ENUM:
                     enum_cls = cls._betterproto.cls_by_field[field_name]
+
+                    def obj(enum_class, value):
+                        return enum_class.from_full_name(value) if hasattr(enum_class, 'from_full_name') else enum_class.from_string(value)
```
Gobot1234, line 1581:
```suggestion
                    def obj(enum_class, value):
                        return enum_class.from_full_name(value) if hasattr(enum_class, 'from_full_name') else enum_class.from_string(value)
```
Same as previous here

Gobot1234 (approved):
Please can you also run poe fmt

PR #605, opened 2024-09-04
When decoding strings, added possibility to choose how to manage errors

Brosssh:
Context:
I'm using your library since the public API of a game is using protobuf. I'm builing a bot over these API. 
While parsing a Message received from this API, I noticed I was getting an exception `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd3 in position 82: invalid continuation byte`. This beacuse the API returned a somehow broken Message, with not UTF8 characters in it. The game is however built in C++, so the behaviour is different, causing no issue in the game but an exception in my bot.
image

By default, Python [bytes.decode](https://docs.python.org/3/library/stdtypes.html#bytes.decode) method is strict, so it crashes in case it goes in errror. This PR aim to change this behaviour by adding a parameter that an user can modify in order to ignore or replace the non UTF8 characters.

Regarding the test case, I could provide the bytes and the Message definition I've used, however it contains a lot of other non relevant fields, can I somehow reduce it to only have the broken part?

Regarding the documentation, I'm not native english speaker/writer so it's probably better if I don't update it myself.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -147,6 +147,11 @@ WIRE_LEN_DELIM_TYPES = [TYPE_STRING, TYPE_BYTES, TYPE_MESSAGE, TYPE_MAP]
 # Indicator of message delimitation in streams
 SIZE_DELIMITED = -1
 
+# Enum that contains the possible decoding methods when decoding a str (https://docs.python.org/3/library/stdtypes.html#str)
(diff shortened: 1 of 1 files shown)


Brosssh:
I'm also just an amateur coder so please forgive me if my code is not optimal

Issue #606, opened 2024-09-04
PLACEHOLDER test sometimes fails when using pydantic


AdrienVannson:
### Summary

`PLACEHOLDER` test sometimes fails when using pydantic

### Reproduction Steps

Compile the following messages _with pydantic_:

```
message A {
	B v = 1;
}

message B {
	A v = 1;
}
```

In Python, import the messages and do `print(A())`.

### Expected Results

I would have expected this instruction to print only `A()`.

### Actual Results

This prints: `A(v=<object object at 0x7f081c1f1a10>)`

The problems comes from the test `if ... is PLACEHOLDER` in the `__repr__` function: the field is considered different from `PLACEHOLDER`, even if it is an `object()` as well. Similar errors can occur in other contexts where comparing the field to `PLACEHOLDER` is needed.

After some time, I think I finally understood the reason: pydantic dataclasses behave differently from python dataclasses when given a mutable default value: pydantic automatically detects this and performs a deep copy. Thus, the value put in the field is not the `PLACEHOLDER` but a copy of this instance. This makes all the tests `if field is PLACEHOLDER` fail.

In my example, I used nested messages since it was the easiest way that I found to reproduce the bug, but I think I came across it in other circumstances.

This behavior of pydantic has been discussed here: https://github.com/pydantic/pydantic/issues/3728 I think a potential solution could be to set `PLACEHOLDER` to an immutable value, as suggested by the last message of the conversation. It solves the problem in my example, but I need to further test this solution. I can send a pull-request soon if you agree with this solution.

### System Information

libprotoc 3.12.4
Python 3.12.5
Name: betterproto
Version: 2.0.0b7
Summary: A better Protobuf / gRPC generator & library
Home-page: https://github.com/danielgtaylor/python-betterproto
Author: Daniel G. Taylor
Author-email: [danielgtaylor@gmail.com](mailto:danielgtaylor@gmail.com)
License: MIT
Location: XXX
Requires: grpclib, python-dateutil, typing-extensions
Required-by: pygrpc-poc

### Checklist

- [X] I have searched the issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have verified this issue occurs on the latest prelease of betterproto which can be installed using `pip install -U --pre betterproto`, if possible.

Gobot1234:
I would have thought object would have counted though honestly given the constraints of we can only have valid protobuf types as values we could always just use something like EllipsisType or NoneType

PR #611, opened 2024-09-09
Fix placeholder deep copy

AdrienVannson:
## Summary

[EDIT] Fix https://github.com/danielgtaylor/python-betterproto/issues/606

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -170,15 +170,20 @@ class Casing(builtin_enum.Enum):
 
 
 class Placeholder:
-    def __copy__(self):
+    __slots__ = ()
+
+    def __repr__(self) -> str:
+        return "<PLACEHOLDER>"
+
+    def __copy__(self) -> Self:
         return self
 
-    def __deepcopy__(self, _):
+    def __deepcopy__(self, _) -> Self:
         return self
 
 
 # We can't simply use object() here because pydantic automatically performs deep-copy of mutable default values
-# See https://github.com/danielgtaylor/python-betterproto/issues/606
+# See #606
 PLACEHOLDER: Any = Placeholder()
 
 
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -184,7 +184,7 @@ class Placeholder:
 
 # We can't simply use object() here because pydantic automatically performs deep-copy of mutable default values
 # See #606
-PLACEHOLDER: Placeholder = Placeholder()
+PLACEHOLDER: Any = Placeholder()
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


src/betterproto/__init__.py
```diff
@@ -169,7 +169,17 @@ class Casing(builtin_enum.Enum):
...
+    def __copy__(self):
+        return self
+
+    def __deepcopy__(self, _):
+        return self
+
+
+# We can't simply use object() here because pydantic automatically performs deep-copy of mutable default values
+# See https://github.com/danielgtaylor/python-betterproto/issues/606
+PLACEHOLDER: Any = Placeholder()
```
Gobot1234, line 182:
```suggestion
class Placeholder:
    __slots__ = ()

    def __repr__(self) -> str:
        return "<PLACEHOLDER>"

    def __copy__(self) -> Self:
        return self

    def __deepcopy__(self, _) -> Self:
        return self


# We can't simply use object() here because pydantic automatically performs deep-copy of mutable default values
# See #606
PLACEHOLDER: Any = Placeholder()
```

AdrienVannson: The changes look fine, thanks!

src/betterproto/__init__.py
```diff
@@ -169,7 +169,22 @@ class Casing(builtin_enum.Enum):
...
+    def __copy__(self) -> Self:
+        return self
+
+    def __deepcopy__(self, _) -> Self:
+        return self
+
+
+# We can't simply use object() here because pydantic automatically performs deep-copy of mutable default values
+# See #606
+PLACEHOLDER: Placeholder = Placeholder()
```
Gobot1234, line 187:
```suggestion
PLACEHOLDER = cast(Any, Placeholder())
```

AdrienVannson: I'm not sure to understand, is there any reason for this?

Gobot1234: It's so it can be used anywhere and the type checker won't complain

AdrienVannson: I think the type checker complains anyway: with mypy, your changes turns:
`src/betterproto/__init__.py:224: error: Argument "default" to "field" has incompatible type "Placeholder | None"; expected "Field[Any]"  [arg-type]`
into
`src/betterproto/__init__.py:224: error: Argument "default" to "field" has incompatible type "Any | None"; expected "Field[Any]"  [arg-type]`

It is coherent since if x has type `T` and y has type `Any`, then `x = y` should not be a valid operation. Wouldn't it be better to add `# type: ignore` on the lines where it is needed?

Gobot1234: It should still be an Any

AdrienVannson: I updated the PR

Gobot1234 (approved):

AdrienVannson:
Thanks!

Issue #604, opened 2024-09-03
Pydantic validation error when creating instance without parameter of message with oneof


AdrienVannson:
### Summary

Decoding a message with oneof fails.

### Reproduction Steps

To show the problem, here is a simple example. Compile the following proto messages:

```
message X {
	int32 v = 1;
}

message Y {
	int32 v = 1;
}

message Choice {
	oneof choice {
		X x = 1;
		Y y = 2;
	}
}
```

In a Python script, creating an instance without any parameter ( `Choice()` ) fails with a validation error: `Value error, Group choice has no value; all fields are None`.

This error comes from `_validate_field_groups` in `src/betterproto/__init__.py`.

### Expected Results

In this simple case, the error might seem expected. However, betterproto creates instances of messages without any parameter a bit everywhere in the code.  For example, parsing a message is done as follows: `value = cls().parse(value)`. An instance is created without any parameter, which fails if the message contains a oneof field.

### Actual Results

I would have expected receiving messages with oneof fields to work.

I see two potential solutions:
  - rewrite the code to avoid creating instances of messages without parameters. I think `parse`, `from_dict` , etc should probably be class methods instead. It would probably need many changes in the source code, and would also need an update to the rust codec
  - Remove the check that at least one field in the oneof is specified. It would be easier, but it would it would remove a check that is useful. Plus, would say that the methods like `parse` have no reason to be called on an actual instance of the object (they are always called on an instance created just for the call).

What do you think?

### System Information

libprotoc 3.12.4
Python 3.12.5
Name: betterproto
Version: 2.0.0b7
Summary: A better Protobuf / gRPC generator & library
Home-page: https://github.com/danielgtaylor/python-betterproto
Author: Daniel G. Taylor
Author-email: danielgtaylor@gmail.com
License: MIT
Location: /home/avannson_gcp/pygrpc_poc/.venv/lib/python3.12/site-packages
Requires: grpclib, python-dateutil, typing-extensions
Required-by: pygrpc-poc

### Checklist

- [X] I have searched the issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have verified this issue occurs on the latest prelease of betterproto which can be installed using `pip install -U --pre betterproto`, if possible.

Gobot1234:
The parse methods used to be hybridmethods not sure why this code has disappeared

AdrienVannson:
[...]

PR #612, opened 2024-09-09
Remove error when no fields are set in oneof

AdrienVannson:
## Summary

Fix https://github.com/danielgtaylor/python-betterproto/issues/604

Methods such as `parse` are often called on a new object: `value = cls().parse(value)`. However, is the message contains oneof fields, this used to make the validator `_validate_field_groups` fail when using pydantic.

I removed the check that at least one value of the `oneof` is defined. This is coherent with gRPC's documentation: "If you have a message with many fields and where _at most_ one field will be set at the same time, you can enforce this behavior and save memory by using the oneof feature." and "You can check which value in a oneof is set _(if any)_ using a special case() or WhichOneof() method"

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -1864,9 +1864,7 @@ class Message(ABC):
                 if getattr(values, field.name, None) is not None
             ]
 
-            if not set_fields:
-                raise ValueError(f"Group {group} has no value; all fields are None")
-            elif len(set_fields) > 1:
+            if len(set_fields) > 1:
                 set_fields_str = ", ".join(set_fields)
                 raise ValueError(
                     f"Group {group} has more than one value; fields {set_fields_str} are not None"


AdrienVannson:
I forget to mention: this problem makes some of the tests from `test_inputs.py` fail. The problem only appears when `plugin_output_package` is manually set to `"tests.output_betterproto_pydantic"`, which is not the case by default. Once this problem (and at least another one), I will send a PR to also execute these tests with Pydantic.

Gobot1234 (approved):

Gobot1234:
> I forget to mention: this problem makes some of the tests from `test_inputs.py` fail. The problem only appears when `plugin_output_package` is manually set to `"tests.output_betterproto_pydantic"`, which is not the case by default. Once this problem (and at least another one) is fixed, I will send a PR to also execute the input tests with Pydantic.

That would be a huge help actually

haibo-looki:
Need this fix. Hoping it can be released in the near future.

AdrienVannson:
I would like to fix a few small problems before, but it should not take long. Once this is merged, I think we can release a new version. In the meantime, you can use the last version on master, which should work fine.

PR #614, opened 2024-09-11
Simplify code: delete py_input_message

AdrienVannson:
## Summary

This PR removes the `py_input_message` function from `ServiceMethodCompiler`. This function was only used in two places, in the template file. The return value was not actually used, it was only checked whether it was None or not.

However, the function should never actually return None, since an `rpc` always takes an existing message as parameter. If the function returns None, as far as understand, it means that there is a problem somewhere in betterproto.

In any case, the check is not useful: if the condition that I deleted was not satisfied (which should not happen from what I understand), an invalid code would have been generated since the variable `method.py_input_message_param` would have been used later without having being declared.

I sometimes see things like that in the code, I imagine that it is better that I do small pull-requests each time, what do you think? Or is there any reason to keep the code for now, for example for future use?

## Checklist

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -756,30 +756,6 @@ class ServiceMethodCompiler(ProtoContentBase):
         return f"/{package_part}{self.parent.proto_name}/{self.proto_name}"
 
     @property
-    def py_input_message(self) -> Optional[MessageCompiler]:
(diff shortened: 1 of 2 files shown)


Gobot1234 (approved):

Gobot1234:
Should be alright to cleanup the code I suppose

PR #615, opened 2024-09-11
Prevent users from creating messages with wrong parameters when pydantic is used

AdrienVannson:
## Summary

Python's dataclass prevent the user from creating instances of a dataclass with invalid fields (see the test that I added). However, this is not the case by default with Pydantic: the invalid fields are just removed from the message ( https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra ).

This can be the cause of errors if there is a typo in the name of a parameter. For this reason, I think it is better to add `config={"extra": "forbid"}` to pydantic dataclasses so that a `ValidationError` is raised in such a situation.

To avoid adding type checking errors (this parameter does not exist in python's dataclasses), I had to remove the `if TYPE_CHECKING` condition that was used to import Python's dataclass instead. I updated `mypy` to its last version to avoid having the bug mentionned in https://github.com/danielgtaylor/python-betterproto/pull/460 . To do this, it was needed to remove support for python 3.7... but I think it is fine given the fact that the tests are no longer run with this python version, which is very old anyway. This also has the benefit of making the code a bit more clear.

One point I am not completely sure about: I had to remove `optional = true` for `black`, `jinja` and `isort` in `pyproject.toml` in order to be able to execute the tests... I'm not sure why it used to work before.

## Checklist

- [X] This change has an associated test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,13 +12,12 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.8"
 black = { version = ">=23.1.0", optional = true }
 grpclib = "^0.4.1"
-importlib-metadata = { version = ">=1.6.0", python = "<3.8" }
 jinja2 = { version = ">=3.0.3", optional = true }
 python-dateutil = "^2.8"
-isort = {version = "^5.11.5", optional = true}
+isort = {version = "^5.11.5", optional = true }
 typing-extensions = "^4.7.1"
 betterproto-rust-codec = { version = "0.1.1", optional = true }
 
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -12,21 +12,20 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
-black = { version = ">=23.1.0", optional = true }
```
Gobot1234, line 16: Not using optional here is not acceptable for libraries that bundle their protobufs as it will install black, isort etc

Gobot1234 (commented):
Other than the packaging issues seems like a good change to me

AdrienVannson:
Thanks! The problem that I had was that if I set ` optional = True` , the packages were not installed... but it should be possible somehow since it was how it was done before. I will have a look tomorrow

AdrienVannson:
It is fixed now :)

Gobot1234 (approved):

Gobot1234:
Thank you

Issue #598, opened 2024-08-16
Using Pydantic creates circular imports


AdrienVannson:
### Summary

Using Pydantic creates circular imports

### Reproduction Steps

Compile the following `.proto` files:

`A_a.proto`:
```
syntax = "proto3";
package root.apackage;

import "B.proto";

message AaMsg {
    root.bpackage.BMsg x = 1;
}
```

`A_b.proto`:
```
syntax = "proto3";
package root.apackage;

message AbMsg {
    int32 x = 1;
}
```

`B.proto`:
```
syntax = "proto3";
package root.bpackage;

import "A_b.proto";

message BMsg {
    apackage.AbMsg x = 1;
}
```

There is no circular dependency between the `.proto` files. However, after the compilation, there is a circular dependency between the two packages.

### Expected Results

I compiled the `.proto` files with the `pydantic_dataclasses`.
I would expect importing the library with `import root.apackage` to work correctly.

### Actual Results

`import root.apackage` fails with the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "XXX/root/apackage/__init__.py", line 17, in <module>
    from .. import bpackage as _bpackage__
  File "XXX/root/bpackage/__init__.py", line 21, in <module>
    class BMsg(betterproto.Message):
  File "XXX/.venv/lib/python3.10/site-packages/pydantic/dataclasses.py", line 249, in create_dataclass
    pydantic_complete = _pydantic_dataclasses.complete_dataclass(
  File "XXX/.venv/lib/python3.10/site-packages/pydantic/_internal/_dataclasses.py", line 120, in complete_dataclass
    set_dataclass_fields(cls, types_namespace, config_wrapper=config_wrapper)
  File "XXX/.venv/lib/python3.10/site-packages/pydantic/_internal/_dataclasses.py", line 82, in set_dataclass_fields
    fields = collect_dataclass_fields(cls, types_namespace, typevars_map=typevars_map, config_wrapper=config_wrapper)
  File "XXX/.venv/lib/python3.10/site-packages/pydantic/_internal/_fields.py", line 313, in collect_dataclass_fields
    ann_type = _typing_extra.eval_type_lenient(dataclass_field.type, types_namespace, cls_localns)
  File "XXX/.venv/lib/python3.10/site-packages/pydantic/_internal/_typing_extra.py", line 240, in eval_type_lenient
    return eval_type_backport(value, globalns, localns)
[...]

PR #624, opened 2024-10-15
Fix import loop with Pydantic

AdrienVannson:
## Summary

This fixes https://github.com/danielgtaylor/python-betterproto/issues/598

The bug occured when executing the input tests with Pydantic: an import loop prevented one test from being executed correctly. I will add another PR later to allow executing the input tests with Pydantic.

I moved some imports to the end of the file in order to avoid having the import loop error. I added some quotes around types to prevent Python from raising an error since it doesn't know yet the types that are used.

I removed the calls to `rebuild_dataclass`, as Pydantic already tries to rebuild dataclasses when needed:

```python
from pydantic.dataclasses import dataclass

@dataclass
class Root:
    x: "Child | None" = None

@dataclass
class Child:
    y: int

print(Root.__pydantic_complete__)
obj = Root(x={"y": 12})
print(Root.__pydantic_complete__, obj)
```

I also had to rename a field of a test message in order to prevent invalid Python code from being generated: Python does not allow using the name of the field of a class in a type annotation. For example, the following program is invalid:

```python

@dataclass
class A:
    pass

@dataclass
class B:
    A: A | None = None
```

Solving this problem should be done in another PR.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -238,7 +238,7 @@ class OutputTemplate:
     parent_request: PluginRequestCompiler
     package_proto_obj: FileDescriptorProto
     input_files: List[str] = field(default_factory=list)
-    imports: Set[str] = field(default_factory=set)
+    imports_end: Set[str] = field(default_factory=set)
     datetime_imports: Set[str] = field(default_factory=set)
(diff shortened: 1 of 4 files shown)


Gobot1234 (approved):
Thanks

Issue #486, opened 2023-05-09
Create `__all__` to explicitly export only the important parts


MicaelJarniac:
Currently, it generates the following:
```py
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: example.proto
# plugin: python-betterproto
from dataclasses import dataclass

import betterproto


@dataclass
class Greeting(betterproto.Message):
    """Greeting represents a message you can tell a user."""

    message: str = betterproto.string_field(1)
```

If you try to import from it, you'll get suggestions for importing `dataclass`, `betterproto` and `Greeting`.

Those first two, though, aren't meant to be re-exported there, only `Greeting` is supposed to be importable.

By generating with an `__all__ = ["Greeting"]`, Python would know that the only thing it's supposed to import is `Greeting`:

```py
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: example.proto
# plugin: python-betterproto

__all__ = ["Greeting"]

from dataclasses import dataclass

import betterproto


@dataclass
class Greeting(betterproto.Message):
    """Greeting represents a message you can tell a user."""

    message: str = betterproto.string_field(1)
```

https://docs.python.org/3/tutorial/modules.html#importing-from-a-package
https://pep8.org/#module-level-dunder-names

PR #625, opened 2024-10-15
Define __all__

AdrienVannson:
## Summary

Define `__all__` in the generated files, as suggested by https://github.com/danielgtaylor/python-betterproto/issues/486

## Checklist

- [X] This PR fixes an issue.
- [X] This PR adds something new (e.g. new method or parameters).

### DIFF
--- a/src/betterproto/templates/header.py.j2
+++ b/src/betterproto/templates/header.py.j2
@@ -2,6 +2,20 @@
 # sources: {{ ', '.join(output_file.input_filenames) }}
 # plugin: python-betterproto
 # This file has been @generated
+
+__all__ = [
+    {%- for enum in output_file.enums %}
+        "{{ enum.py_name }}",
+    {%- endfor -%}
+    {%- for message in output_file.messages -%}
+        "{{ message.py_name }}",
+    {%- endfor -%}
+    {%- for service in output_file.services -%}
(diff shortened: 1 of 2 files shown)


Gobot1234:
Please move this to the end of the file, because honestly it's not something I want to see

Gobot1234:
Also please can you change this to a tuple not a list

AdrienVannson:
It is defined at the top of the file on purpose, as suggested by PEP-8: https://peps.python.org/pep-0008/#module-level-dunder-names

Gobot1234:
Please can you add a test that this exists and avoids importing dataclasses etc?

Gobot1234:
> It is defined at the top of the file on purpose, as suggested by PEP-8: [peps.python.org/pep-0008#module-level-dunder-names](https://peps.python.org/pep-0008/#module-level-dunder-names)

I have never seen it go before module imports hmm.

cetanu (approved):
If the python docs say it should go before imports, then okay.

AdrienVannson:
> Please can you add a test that this exists and avoids importing dataclasses etc?

I added a test that check that `__all__` is correctly defined. However, I'm not sure what you mean by importing dataclasses

Gobot1234 (approved):

PR #630, opened 2024-10-18
Fix documentation

AdrienVannson:
## Summary

The documentation of services and methods was not properly added to the generated files. The issue was that `ServiceCompiler` and `ServiceMethodCompiler` didn't have access to the value of `source_file`, unlike the other compilers.

As far as I understand, the field was accessed in the `get_comment ` function. Since it was missing, an `AttributeError` was raised, but this error was catched by jinja automatically, as in the following example:

```python

from jinja2 import Template

TEMPLATE: str = """
AA
{{ my_class.field }}
BB
{{ my_class.nothing }}
CC
"""

@dataclass
class MyClass:
    field: int

print(Template(TEMPLATE).render(my_class=MyClass(field=42)))
```

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/src/betterproto/plugin/models.py
+++ b/src/betterproto/plugin/models.py
@@ -661,6 +661,7 @@ class EnumDefinitionCompiler(MessageCompiler):
 
 @dataclass
 class ServiceCompiler(ProtoContentBase):
+    source_file: FileDescriptorProto
     parent: OutputTemplate = PLACEHOLDER
     proto_obj: DescriptorProto = PLACEHOLDER
     path: List[int] = PLACEHOLDER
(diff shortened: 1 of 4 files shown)


AdrienVannson:
To avoid silencing these errors, it is possible to add a parameter `undefined=jinja2.StrictUndefined` to the jinja2 environment. I didn't add it here since other errors are currently raised, but I think it would be a good thing to do it at some point. Maybe some issues are still hidden by this behavior.

Gobot1234 (approved):

Gobot1234:
Thanks!

Issue #562, opened 2024-03-12
Invalid code generated when package name contains uppercase letter


paskozdilar:
### Summary

Title

### Reproduction Steps

This protofile works:

```proto3
// proto/test.proto
syntax = "proto3";

package p;

service Service {
  rpc Method(Empty) returns (Empty) {}
}

message Empty {}
```

This doesn't:

```proto3
// proto/test.proto
syntax = "proto3";

package P;

service Service {
  rpc Method(Empty) returns (Empty) {}
}

message Empty {}
```

With the latter, betterproto compiles to this:

```python3
# Generated by the protocol buffer compiler.  DO NOT EDIT!
# sources: test.proto
# plugin: python-betterproto
# This file has been @generated

from dataclasses import dataclass
from typing import (
    TYPE_CHECKING,
    Dict,
    Optional,
)

import betterproto
import grpclib
from betterproto.grpc.grpclib_server import ServiceBase

from .. import PEmpty as _PEmpty__

# [...]
```

The `from .. import PEmpty` is invalid import as nothing exists in the parent package.

Command used:

```bash
python3 -m grpc_tools.protoc --proto_path=./proto/ --python_betterproto_out=./proto/ proto/*.proto
```

Python environment (output of `pip freeze`):
```
betterproto @ git+https://github.com/danielgtaylor/python-betterproto@5666393f9d10e13609d8eeac8d1ab3815dce5fd6
black==24.2.0
click==8.1.7
grpcio==1.62.1
grpcio-tools==1.62.1
grpclib==0.4.7
h2==4.1.0
hpack==4.0.0
hyperframe==6.0.1
isort==5.13.2
Jinja2==3.1.3
MarkupSafe==2.1.5
multidict==6.0.5
mypy-extensions==1.0.0
nodeenv==1.8.0
packaging==24.0
pathspec==0.12.1
platformdirs==4.2.0
protobuf==4.25.3
pyright==1.1.353
python-dateutil==2.9.0.post0
six==1.16.0
typing_extensions==4.10.0
```

### Expected Results

Correct code generated

### Actual Results

Incorrect code generated

### System Information

libprotoc 3.12.4
Python 3.11.0rc1
Name: betterproto
Version: 2.0.0b6
Summary: A better Protobuf / gRPC generator & library
Home-page: https://github.com/danielgtaylor/python-betterproto
Author: Daniel G. Taylor
Author-email: danielgtaylor@gmail.com
License: MIT
[...]

Issue #437, opened 2022-10-24
betterproto-2.0.0b5 fails to generate message classes when proto package name contains upper case letters


hf-kklein:
### Install `betterproto-2.0.0b5` + tooling:

```bash
pip install --pre betterproto[compiler]
pip install grpcio-tools
```
other versions: 
> grpcio==1.50.0
grpcio-tools==1.50.0
grpclib==0.4.3

### Create an `example.proto`
```proto
syntax = "proto3";

package mypackage; // this is the only change to the MWE from README.md

message EchoRequest {
  string value = 1;
  // Number of extra times to echo
  uint32 extra_times = 2;
}

message EchoResponse {
  repeated string values = 1;
}

message EchoStreamResponse  {
  string value = 1;
}

service Echo {
  rpc Echo(EchoRequest) returns (EchoResponse);
  rpc EchoStream(EchoRequest) returns (stream EchoStreamResponse);
}
```

### Create the directory
```bash
mkdir lib
```

### Generate Python Code:
```bash
python -m grpc_tools.protoc -I . --python_betterproto_out=lib example.proto
```

Check that `lib/mypackage/__init__.py` contains _no_ unresolvable imports. ✔

Up to here, this is just the MWE from `README.md` with a different but all-lower package name.

### Introduce Upper Case Letters in the Package Name

Now modify the above `example.proto` and change the all-lower case `package mypackage` in line 3 to an UpperCamel `package MyPackage`.

Repeat the next steps up to the code generation (don't forget to clean up `lib` before re-generating).

The newly generate `lib/MyPackage/__init__.py` now contains unresolvable imports:
```python
from .. import (
    MyPackageEchoResponse as _MyPackageEchoResponse__,
    MyPackageEchoStreamResponse as _MyPackageEchoStreamResponse__,
)
```

The `lib/__init__.py` does not contain anything to import.
No classes `MyPackageEchoResponse` or `MyPackageEchoStreamResponse` have been generated ❌

## Expected Behaviour
The code generation shall work for mixed casing package names.

hf-kklein:
For a similar casing issue in the **message** name see #438.

2AtM7:
Hi, any plans on fixing this issue? Seems like there has been a fix (#436) for #438, which is a similar issue, but not sure if this one was also taken care of by the same PR.
Also, any plans for a new release or beta including the above fix? The current beta (2.0.0b5) seems to be 9 months old :)

PR #635, opened 2024-10-25
Fix parse_source_type_name

AdrienVannson:
## Summary

Fix https://github.com/danielgtaylor/python-betterproto/issues/562 , https://github.com/danielgtaylor/python-betterproto/issues/437

The function `parse_source_type_name` used to rely on an ad-hoc regex to work. The function failed when the package name had capital letters, since it believed that the package was actually a message.

Now, the function goes through the packages and the messages that are actually defined (no more guess) to find the right one.

I also did some small refactoring of `models.py`: `FieldCompiler` used to inherit from `MessageCompiler`, which made no sense.

The PR still lacks tests and some doc, but since it can take a while to write, I first wanted to have your point of view on this.

NOTE: I completely deleted `test_gen_ref_type.py` for now, but I plan to replace it once you agree with the content of the PR

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/src/betterproto/compile/importing.py
+++ b/src/betterproto/compile/importing.py
@@ -61,6 +61,7 @@ def parse_source_type_name(
         if package := request.output_packages.get(package_name):
             if object_name in package.messages or object_name in package.enums:
                 if answer:
+                    # This should have already been handeled by protoc
                     raise ValueError(f"ambiguous definition: {field_type_name}")
                 answer = package_name, object_name
 


Gobot1234 (approved):

Gobot1234:
Ping me once the tests are done

AdrienVannson:
Hello, I added some documentation, I would say that it is enough. Concerning the tests, I can't just add many simple unit tests as easily as before since the function `get_type_reference` now also takes a `PluginRequestCompiler`. However, everything should already be tester in the `import_*` input tests.

I just need to find out why the tests are failing, this is not normal since I almost changed nothing

AdrienVannson:
It was just the merge that was not performed correctly, now it should be fine

src/betterproto/plugin/models.py
```diff
@@ -587,12 +600,22 @@ def pydantic_imports(self) -> Set[str]:
 
 @dataclass
 class MapEntryCompiler(FieldCompiler):
-    py_k_type: Type = PLACEHOLDER
-    py_v_type: Type = PLACEHOLDER
-    proto_k_type: str = PLACEHOLDER
-    proto_v_type: str = PLACEHOLDER
+    py_k_type: Type = None
+    py_v_type: Type = None
```
Gobot1234, line 604: Please change this to pass type checking

src/betterproto/templates/header.py.j2
```diff
@@ -4,13 +4,13 @@
...
-    {%- for enum in output_file.enums -%}
+    {% for _, enum in output_file.enums|dictsort(by="key") %}
         "{{ enum.py_name }}",
     {%- endfor -%}
-    {%- for message in output_file.messages -%}
+    {% for _, message in output_file.messages|dictsort(by="key") %}
         "{{ message.py_name }}",
     {%- endfor -%}
-    {%- for service in output_file.services -%}
+    {% for _, service in output_file.services|dictsort(by="key") %}
```
Gobot1234, line 13: How come these are being sorted now?

Gobot1234 (changes requested):

AdrienVannson:
Note: it should be good to check if it works with Buf (apparently Buf can recompile partially a project, but this update makes the assumption that all the structure of the project is loaded.) I think it should be fine, but it might be worth checking

cetanu (approved):

Issue #631, opened 2024-10-18
Simplifying the getattribute mechanism


AdrienVannson:
### Summary

_

### What is the feature request for?

The core library

### The Problem

Simplifying the getattribute mechanism

Hello,
I have a suggestion that I can probably implement myself if you agree to do it. However, I would like to have your opinion on it before working on it.

Currently, messages redefine their getattribute function. This is used for two reasons:
-	It allows raising an AttributeError when a field that is part of a oneof but is not the active one is accessed
-	It allows replacing the Placeholder values by the default value when the field is accessed. 

However, this mechanism has several important issues, and some of them cannot be fixed without switching to a different method:
-	Redefining getattribute is very slow. Each time a field of a message is accessed, this function is called. This is of course the case for the proto fields, but getattribute is also called when other attributes such as the ones containing the metadata are accessed. Plus, each call is slow since getattribute performs queries to dictionaries. In a small benchmark that I did, I found out that a very significant part of the time spent when decoding a message was spent in this function. This is also a problem when using the rust codec.
-	The lack of support of unset messages (as they are automatically replaced by empty messages) makes it impossible to distinguish between an empty message and no message at all. This should be possible since the protobuf documentation states that the presence of message fields should always be tracked, even when the message is not declared as optional (see https://protobuf.dev/programming-guides/field_presence/ ).
-	Match are difficult to type-check since the semantic used by betterproto differs from the one used by the type-checkers, who don’t consider the getattribute function
-	More generally, the behaviour of this mechanism is not intuitive at all in my opinion: reading the value of an attribute may change the value of the message (by adding an empty file). I believe that a simple read should not have this behaviour.
These issues are so important that they made someone I know prefer removing all these functionalities and deal with the placeholders by hands… which is far from being ideal.
I would rather do more of less the following:
-	For primitive fields (int, str, float, …), nothing much changes apart from the default value: it becomes 0, “”, 0.0, … instead of PLACEHOLDER
-	Messages fields are always declared as optional (which is coherent with the protobuf documentation). When no value is provided, None is used instead. None and an empty message are therefore different values, which is expected. A consequence is that empty messages are not created on the fly when accessing unset fields, but I think it is better to have this behaviour anyway.
[...]

Issue #610, opened 2024-09-06
Message with oneof called method .to_pydict/.from_pydict fails 


RF-Tar-Railt:
### Summary

Serializing a message with oneof fails by `to_pydict` or `from_pydict`.

### Reproduction Steps

Define a message and compile:

```protobuf
message Foo {
  oneof group_bar {
    XXX baz = 1;
    XXX qux = 2;
  }
}
```

then called `Foo(...).to_pydict` or `Foo().from_pydict(...)`, there will raise AttributeError: `'group_bar' is set to 'xxxx', bot 'baz' ` / `'group_bar' is set to None, bot 'baz'`

### Expected Results

The method called success like `to_dict` and `from_dict`

### Actual Results

Just one line different:

- `to_dict`:
```python
try:
    value = getattr(self, field_name)
except AttributeError:
    value = self._get_field_default(field_name)
cased_name = casing(field_name).rstrip("_")  # type: ignore
```

- `to_pydict`:
```python
value = getattr(self, field_name)
cased_name = casing(field_name).rstrip("_")  # type: ignore
```

`from_pydict` is the same cause.

### System Information

libprotoc 27.2
Python 3.9.13
2.0.0b7

### Checklist

- [X] I have searched the issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have verified this issue occurs on the latest prelease of betterproto which can be installed using `pip install -U --pre betterproto`, if possible.

Issue #601, opened 2024-08-22
One-of pattern matching not supported by static type analysis tools


124C41p:
According to the [Readme](https://github.com/danielgtaylor/python-betterproto/blob/master/README.md#one-of-support) it should be possible to access fields of a oneof-group by pattern matching so that static analysis tools can provide type hints:

```python3
test = Test()
match test:
    case Test(on=value):
        print(value)  # value: bool
    case Test(count=value):
        print(value)  # value: int
    case Test(name=value):
        print(value)  # value: str
    case _:
        print("No value provided")
```

However, the tool `pyright` (and apparently also mypy) does not provide type hints for the second and third case. I first thought this was a bug in pyright, but [according to a pyright maintainer](https://github.com/microsoft/pyright/issues/8789#issuecomment-2299612496) this is actually intentional. Apparently, from the point of view of a type checker, the second and third case blocks are unreachable (this pattern even triggers a warning from pyright when configured accordingly).

## Possible solution

Let's assume we change the compiler to generate the following class:

```python3
@dataclass(eq=False, repr=False)
class Test(betterproto.Message):
    on: Optional[bool] = betterproto.bool_field(1, group="foo")
    count: Optional[int] = betterproto.int32_field(2, group="foo")
    name: Optional[str] = betterproto.string_field(3, group="foo")
```

Then the following match statement is properly supported by pyright:

```python3
test = Test()
match test:
    case Test(on=bool(value)):
        print(value)  # value: bool
    case Test(count=int(value)):
        print(value)  # value: int
    case Test(name=str(value)):
        print(value)  # value: str
    case _:
        print("No value provided")
```

I think one could argue that it is actually correct to mark these fields optional. What do you think?

Gobot1234:
What does the default implementation do when accessing an unset field? I'm slightly more inclined to follow that behaviour but I've been thinking about this. The change would need to be made quickly to minimise disruption in this regard.

124C41p:
As of version 2.0.0b7, accessing an ordinary unset field gives the default value, whereas accessing an unset field which is part of a oneof group raises an AttributeError:

```protobuf
message Empty {}

message Test {
  oneof foo {
    bool on = 1;
  }
  int32 i = 2;
  Empty e = 3;
}
```

```python3
test = Test()
print(test.i)   # 0
print(test.e)   # Empty()
print(test.on)  # AttributeError
```

[...]

Issue #543, opened 2023-11-14
Nested message with repeated field not in to_json when mutated after initialization


dallagi:
### Summary

Possibly related to #199; Given a message with a nested message containing a repeated field, if the field is mutated (eg. via `list.append`) after initialization, `to_dict` will not include it

### Reproduction Steps

Given a proto schema like this

```protobuf
syntax = "proto3";

message RootMessage {
  NestedMessage nested = 1;
}

message NestedMessage {
  repeated string content = 1;
}
```

Running the following script

```python
from betterproto_compiled import (
    NestedMessage,
    RootMessage,
)
from pprint import pprint

message = RootMessage()

message.nested.content.append("test-content")

print("Including default values: True")
pprint(message.to_dict(include_default_values=True))

print()
print("Including default values: False")
pprint(message.to_dict(include_default_values=False))
```

will result in:

```
Including default values: True
{'nested': {'content': ['test-content']}}

Including default values: False
{}
```

### Expected Results

I would expect the `content` field of the `nested` message to appear in the returned `dict` also when `include_default_values` is set to False, as its content was mutated and it's no longer the default value.

### Actual Results

`to_dict` returns an empty dict

### System Information

```
libprotoc 3.21.6
Python 3.10.13
Name: betterproto
Version: 2.0.0b6
Summary: A better Protobuf / gRPC generator & library
Home-page: https://github.com/danielgtaylor/python-betterproto
Author: Daniel G. Taylor
Author-email: danielgtaylor@gmail.com
License: MIT
Location: /home/marco/.local/share/rtx/installs/python/3.10.13/lib/python3.10/site-packages
Requires: grpclib, python-dateutil
Required-by: 
```

### Checklist

- [X] I have searched the issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have verified this issue occurs on the latest prelease of betterproto which can be installed using `pip install -U --pre betterproto`, if possible.

PR #641, opened 2024-11-12
Remove placeholders

AdrienVannson:
## Summary

Hello!

This is a first step in removing the `PLACEHOLDER` mechanism to greatly improve performance, make the whole code more Pythonic and simple and resolve a lot of bug. The motivation is more deeply described in https://github.com/danielgtaylor/python-betterproto/issues/631

This PR completely removes placeholder values from messages. Instead, fields are always set with there default value (empty string, 0, `None` for optional, members of a oneof and messages, ...). Therefore, all the message field are marked as optional and can be set to None, which is coherent with the specification (see the issue).

As a consequence, this fixes the following issues:
 - https://github.com/danielgtaylor/python-betterproto/issues/601 One-of pattern matching not supported by static type analysis tools
 - https://github.com/danielgtaylor/python-betterproto/issues/543 Nested message with repeated field not in to_json when mutated after initialization
 - https://github.com/danielgtaylor/python-betterproto/issues/199 nested empty types not in to_dict
 - https://github.com/danielgtaylor/python-betterproto/issues/610 Message with oneof called method .to_pydict/.from_pydict fails
 - https://github.com/danielgtaylor/python-betterproto/issues/478 Compatibility of generated Pydantic models with FastAPI for proper OpenAPI Swagger serialization (I haven't tested but probably)
 - https://github.com/danielgtaylor/python-betterproto/pull/200 to_dict on empty nested types (no need for this PR anymore)
 - (others, I need to complete the list)

Before merging it, I think it would be good to finish the work in progress (at least my other PR and another fix of the generation of the comments), and release a new version with all the recent bugfixes.

## Performance

These changes also make it possible to remove the `__getattribute__` from messages. As expected, this has a great impact on the performance.

With the following messages:

```
message Test {
    bool x = 1;
    int32 y = 2;
    float z = 3;
}

message List {
    repeated Test msg = 1;
}
```

The following code takes 3.24s on my computer with the original (`master`) implementation **with** the Rust codec, and only 2s with my PR (without the Rust codec). Thus, even if the Rust codec is not supported for now, it is not a problem since the performance are better anyway.

```python
msg = List(msg=[Test(x=bool(i % 2), y = i, z = i / 2) for i in range(100000)])
b = bytes(msg)
msg2 = List().parse(b)
assert msg == msg2
```

## Breaking changes

This PR introduces two small breaking changes:

 - Lazy-initialization of messages is no longer possible, since uninitialized message fields now have `None` as value and `getattribute` is no longer defined.
 - The `serialized_on_wire` function is now removed, since messages that were not serialized on the wire are now represented as `None` values. This function didn't really make sense in the first place imo, since being serialized is a property of a field of a message, not a property of a message itself (google's version has `HasField`, which indeed takes the field as a parameter).

## Checklist

- [X] This PR fixes an issue.
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)


src/betterproto/__init__.py
```diff
@@ -213,15 +195,27 @@ def get(field: dataclasses.Field) -> "FieldMetadata":
...
     group: Optional[str] = None,
     wraps: Optional[str] = None,
     optional: bool = False,
+    repeated: bool = False,
 ) -> dataclasses.Field:
     """Creates a dataclass field with attached protobuf metadata."""
+    if repeated:
+
+        def default_factory():
+            return []
```
Gobot1234, line 210: use the list constructor here for better performance

src/betterproto/__init__.py
```diff
@@ -213,15 +195,27 @@ def get(field: dataclasses.Field) -> "FieldMetadata":
...
+
+    elif optional or group:
+
+        def default_factory():
+            return None
```
Gobot1234, line 215: use NoneType here

src/betterproto/__init__.py
```diff
@@ -235,114 +229,276 @@ def dataclass_field(
...
+def bool_field(
+    number: int,
+    group: Optional[str] = None,
+    optional: bool = False,
+    repeated: bool = False,
+) -> Any:
+    return dataclass_field(
+        number,
+        TYPE_BOOL,
+        lambda: False,
```
Gobot1234, line 258:
```suggestion
        bool,
```

src/betterproto/__init__.py
```diff
@@ -235,114 +229,276 @@ def dataclass_field(
...
 def int32_field(
-    number: int, group: Optional[str] = None, optional: bool = False
+    number: int,
+    group: Optional[str] = None,
+    optional: bool = False,
+    repeated: bool = False,
 ) -> Any:
-    return dataclass_field(number, TYPE_INT32, group=group, optional=optional)
+    return dataclass_field(
+        number, TYPE_INT32, lambda: 0, group=group, optional=optional, repeated=repeated
```
Gobot1234, line 272:
```suggestion
        number, TYPE_INT32, int, group=group, optional=optional, repeated=repeated
```

src/betterproto/__init__.py
```diff
@@ -235,114 +229,276 @@ def dataclass_field(
...
 def int64_field(
-    number: int, group: Optional[str] = None, optional: bool = False
+    number: int,
+    group: Optional[str] = None,
+    optional: bool = False,
+    repeated: bool = False,
 ) -> Any:
-    return dataclass_field(number, TYPE_INT64, group=group, optional=optional)
+    return dataclass_field(
+        number, TYPE_INT64, lambda: 0, group=group, optional=optional, repeated=repeated
```
Gobot1234, line 283:
```suggestion
        number, TYPE_INT64, int, group=group, optional=optional, repeated=repeated
```
I think you get the idea

src/betterproto/__init__.py
```diff
@@ -727,8 +883,20 @@ def _get_cls_by_field(
                 field_cls[field.name] = dataclasses.make_dataclass(
                     "Entry",
                     [
-                        ("key", kt, dataclass_field(1, meta.map_types[0])),
-                        ("value", vt, dataclass_field(2, meta.map_types[1])),
+                        (
+                            "key",
+                            kt,
+                            dataclass_field(
+                                1, meta.map_types[0], default_factory=lambda: kt()
```
Gobot1234, line 890:
```suggestion
                                1, meta.map_types[0], default_factory=kt
```

src/betterproto/__init__.py
```diff
@@ -727,8 +883,20 @@ def _get_cls_by_field(
...
+                            ),
+                        ),
+                        (
+                            "value",
+                            vt,
+                            dataclass_field(
+                                2, meta.map_types[1], default_factory=lambda: vt()
```
Gobot1234, line 897:
```suggestion
                                2, meta.map_types[1], default_factory=vt
```

Gobot1234 (changes requested):
I do like this after more thought. Really do wish that PEP 515 happened

AdrienVannson:
Thanks! I applied your suggestions

cetanu (approved):

src/betterproto/plugin/models.py
```diff
@@ -127,25 +127,26 @@
...
-        ],
-        "group",
-        "oneof_index",
-    )
-    object.__setattr__(
-        Field.__dataclass_fields__["oneof_index"].metadata["betterproto"],
-        "group",
-        "oneof_index",
-    )
+# TODO patch again to make field optional
```
Gobot1234, line 130: Is this no longer required?

AdrienVannson: I let them for now as a reminder to change something else, but I'll remove it soon (in the fork)

AdrienVannson:
Note: I already merged this PR in the forked repo https://github.com/betterproto as discussed on Discord, so it's probably not worth it to solve the merge conflicts here

src/betterproto/plugin/models.py
AdrienVannson: (done)

PR #643, opened 2024-11-13
Don't repeat tests

AdrienVannson:
## Summary

`test_inputs.py` used to include some mechanism to repeat tests several times. I don't see any reasons for this since serializing / deserializing messages is deterministic. This option was not used anyway: `repeat` always had a value of 1.

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,17 +4,6 @@ import sys
 import pytest
 
 
-def pytest_addoption(parser):
-    parser.addoption(
-        "--repeat", type=int, default=1, help="repeat the operation multiple times"
-    )
-
-
-@pytest.fixture(scope="session")
-def repeat(request):
-    return request.config.getoption("repeat")
-
-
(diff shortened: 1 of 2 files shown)


Gobot1234:
I don't see a problem with leaving this in

AdrienVannson:
I don't see any reason to repeat the test, since everything here is fully deterministic. More importantly, this feature is not used at all: the test is always executed only one single time.

Even if, one day, repeating tests is needed, there are much simpler ways to do it (see https://stackoverflow.com/questions/21764473/how-can-i-repeat-each-test-multiple-times-in-a-py-test-run )

It is not a big change, but I am trying as much as I can to simplify the current implementation of betterproto. For now, I have to say that the code is quite messy and many simple changes can become painful because of this :)

cetanu:
my opinion: we can always add it back if we need it

cetanu (approved):

Issue #488, opened 2023-05-10
Use Ruff on generated code (can replace isort)


MicaelJarniac:
https://beta.ruff.rs

Config example:
```toml
# pyproject.toml

[tool.ruff]
# https://beta.ruff.rs/docs/rules
line-length = 88  # Default
select = ["ALL"]
ignore = [
  # "D203",
  # "D213",
  "D",
  "RUF009",  # See `extend-immutable-calls`
  "ERA001",  # Detects comments as code
  # "A003",  # See `builtins-ignorelist`
  "UP006",  # https://github.com/danielgtaylor/python-betterproto/issues/485
]

[tool.ruff.flake8-builtins]
builtins-ignorelist = ["id", "type"]

[tool.ruff.flake8-bugbear]
# https://github.com/charliermarsh/ruff/issues/4355
extend-immutable-calls = ["betterproto.*"]
```

It's possible to have it only emulate isort, but I've tried to enable as much as possible from it, which is still incredibly fast.

Can be run as `ruff check --fix`.

MicaelJarniac:
This is very low priority, and it's easy enough for users to add this functionality manually, via Nox for example, as I have. The example configuration I showed is basically the one I use to run Ruff on generated code.

redbmk:
I just noticed that when installing `betterproto[compiler]` it includes `isort`, which is now running on all files in my editor (I have it configured to run whichever linting tools are installed in the virtualenv so it can easily vary by project) and conflicting with `ruff` (which I'm using instead of `isort` as you are).

Is either `ruff` or `isort` even needed? Feels like formatting and linting tools shouldn't be a dependency of code generators. It would be pretty hard to predict what tools someone is using and what their settings are. IMO for generated code as long as the code works it doesn't need to look super pretty. I typically tell black, ruff, etc to ignore generated code, but of course you can add it as a post-generate step if you like, as you mentioned you're doing.

MicaelJarniac:
I format it myself with a Nox session after the code is generated, so I wouldn't mind if it didn't format.

In your case, though, I'd suggest installing and running `betterproto[compiler]` only in a Nox session, which would keep its dependencies separate from your dev environment, and fix the issue you're having with your editor running isort.

StefanBRas:
Would you accept a PR that made the formatting optional? That way users can still compile without having to install black and isort and then just run their own formatting after

[...]

PR #644, opened 2024-11-13
Use ruff instead of black+isort

AdrienVannson:
## Summary

Use Ruff instead of black+isort. This has the following avantages:
 - Ruff is much faster than black+isort. In my project, it reduced the compilation time of the proto files by half. It will also make the use in the project more convenient.
 - It will allow to greatly simplify how `import` are added to the output file. Instead of having a lot of code to check which imports need to be added to the generated file, it is possible to add all of them and tell Ruff to remove the useless ones. It will make everything easier and less subject to bugs
 - In some cases, the formatting may be better

This was requested in https://github.com/danielgtaylor/python-betterproto/issues/488 , and I agree it is a good choice.

For now, Ruff is only used as a formatter and not as a linter

## Checklist

- [X] This PR fixes an issue.


cetanu (approved):

AdrienVannson:
Actually, I need to check one last thing before merging

AdrienVannson:
I pinned the Ruff version in `pyproject.toml` to avoid having a problem that occurred previously with `black`+`isort`: the version of `black` was updated in the environment, but the old version was still used in the pre-commit... Which used to cause the pre-commit to fail even after running `poe format`

cetanu (approved):

PR #647, opened 2024-11-21
Add a synchronous grpc client

leonardgerardatomicmachinescom:
## Summary

I find that often, we want a non async grpc client. This PR adds a `*SyncStub` generated class using only the standard synchronous grpc library. It is pretty much exactly the same as the `*Stub` one except it is synchronous.

## Checklist

- [ ] This PR fixes an issue.
- [X] This PR adds something new (e.g. new method or parameters).
    - [X] This change has an associated test.

### DIFF
--- a/src/betterproto/templates/template.py.j2
+++ b/src/betterproto/templates/template.py.j2
@@ -63,6 +63,76 @@ class {{ message.py_name }}(betterproto.Message):
     {%  endif %}
 
 {% endfor %}
+
+
+{% for service in output_file.services %}
+class {{ service.py_name }}SyncStub():
+    {% if service.comment %}
+{{ service.comment }}
+
+    {% elif not service.methods %}
+    pass
+    {% endif %}
+
+    import grpc
+
+    def __init__(self, channel: grpc.Channel):
+        self._channel = channel
+
+    {% for method in service.methods %}
+    def {{ method.py_name }}(self
+        {%- if not method.client_streaming -%}
+            , {{ method.py_input_message_param }}: "{{ method.py_input_message_type }}"
+        {%- else -%}
+            {# Client streaming: need a request iterator instead #}
+            , {{ method.py_input_message_param }}: {{ output_file.typing_compiler.iterable(method.py_input_message_type) }}
+        {%- endif -%}
+            ) -> {% if method.server_streaming %}{{ output_file.typing_compiler.sync_iterator(method.py_output_message_type ) }}{% else %}"{{ method.py_output_message_type }}"{% endif %}:
+        {% if method.comment %}
+{{ method.comment }}
+
+        {% endif %}
+        {% if method.proto_obj.options.deprecated %}
+        warnings.warn("{{ service.py_name }}.{{ method.py_name }} is deprecated", DeprecationWarning)
+
+        {% endif %}
+        {% if method.server_streaming %}
+            {% if method.client_streaming %}
+        for response in self._channel.stream_stream(
+            "{{ method.route }}",
+            {{ method.py_input_message_type }}.SerializeToString,
+            {{ method.py_output_message_type }}.FromString,
+        )({{ method.py_input_message_param }}):
+            yield response
+            {% else %}{# i.e. not client streaming #}
+        for response in self._channel.unary_stream(
+            "{{ method.route }}",
+            {{ method.py_input_message_type }}.SerializeToString,
+            {{ method.py_output_message_type }}.FromString,
+        )({{ method.py_input_message_param }}):
+            yield response
+
+            {% endif %}{# if client streaming #}
+        {% else %}{# i.e. not server streaming #}
+            {% if method.client_streaming %}
+        return self._channel.stream_unary(
+            "{{ method.route }}",
+            {{ method.py_input_message_type }}.SerializeToString,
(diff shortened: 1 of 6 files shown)


src/betterproto/templates/template.py.j2
```diff
@@ -160,9 +227,9 @@ class {{ service.py_name }}Base(ServiceBase):
             , {{ method.py_input_message_param }}: "{{ method.py_input_message_type }}"
         {%- else -%}
             {# Client streaming: need a request iterator instead #}
-            , {{ method.py_input_message_param }}_iterator: {{ output_file.typing_compiler.async_iterator(method.py_input_message_type) }}
+            , {{ method.py_input_message_param }}_iterator: "{{ output_file.typing_compiler.async_iterator(method.py_input_message_type) }}"
```
leonardgerardatomicmachinescom, line 230: This and the next diff are in fact fixes for the existing stub generation. If wanted I could split it into a separate PR.

tests/grpc/test_grpc_client.py
```diff
@@ -0,0 +1,84 @@
+""" Testing the sync version of the client stubs.
+
+This is not testing the lower level grpc calls, but rather the generated client stubs.
+So instead of creating a real service and a real grpc channel,
+we are going to mock the channel and simply test the client.
+
+If we wanted to test without mocking we would need to use all the machinery here:
+https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/testing/_client_test.py
```
leonardgerardatomicmachinescom, line 8: Or I could spawn an actual server. I would think that's the best testing, but it is not done so in the rest of the code (which uses the nice grpclib `ChannelFor`)

tests/grpc/test_grpc_client.py
```diff
@@ -0,0 +1,84 @@
...
+"""
+
+import re
+from sys import version
+from tests.output_betterproto.service import (
+    DoThingRequest,
+    DoThingResponse,
+    GetThingRequest,
+    GetThingResponse,
+    TestSyncStub as ThingServiceClient,
```
leonardgerardatomicmachinescom, line 19: This is the new synchronous stub.

AdrienVannson:
Hello, and thank you for the work! I agree that it would be a great idea to support this. However, I wouldn't merge the PR as it is now: indeed, I would say that adding an option to allow the user to choose between sync and async right at the beginning would be better. Plus, it should be possible to avoid duplicating all the template code.

Then, this should be integrated in the github action so that these implementations are tested too... So it is not a small change.

However, before starting working on that, I would like first to simplify the current implementation, since it will have an impact on the template file.

leonardgerardatomicmachinescom:
Sure, it might be useful to provide ways to selectively enable one or the other. In particular I can imagine not wanting to pull in the grpclib dependency if I don't want the async version.
 
Important to note:
In my current workflow, I need both the asynchronous and synchronous generated code.
This means that they can't share the same class name.

leonardgerardatomicmachinescom:
For testing, are you saying that a new test_*.py file is not automatically picked up and ran? Locally the `poe test` command did work as expected. On that note, I've also improved the README since the section on testing appears to be out-dated.

cetanu:
If you can find a way to put this behind a command line option or config, that would be compelling, but I personally have no issues with adding this to the generated code 👍🏼

gurland:
> I would say that adding an option to allow the user to choose between sync and async right at the beginning would be better

@AdrienVannson  If I understood this correctly, you propose to select which version of the client should be generated on the compilation stage? The following comment is based on that interpretation.

In my opinion it is better to generate both sync and async versions. Common "pythonic" way is to just add an `a` prefix before the method name, e.g. Python has dunder methods `__aiter__, __anext__, __aenter__`, LLama-Index has a lot of methods like `stream_chat`/`astream_chat`, DJango has `Model.objects.create`/`Model.objects.acreate`.

it would be more convenient knowing that both synchronous and asynchronous codebases could use the same betterproto generated package, guaranteeing consistency. Otherwise you would have to reinvent the wheel and distribute both versions.

AdrienVannson:
Hello! Due to the difficulty to have PR approved, the impossibility to become a maintainer of this repo as well as to have a better repo name, after discussing with the current maintainers, I did a fork of betterproto here: https://github.com/betterproto/ From now on, it is very likely that most of the development will be done there

I've been improving various aspects of the project, I fixed a lot of bugs and improved the implementation to make it more easy to maintain.

The project is now distributed in the `betterproto2` and `betterproto2_compiler` packages, and I started writing some documentation for it ( https://betterproto.github.io/python-betterproto2/ )

For now, it is still a bit early, I still need time to fix a few remaining problems (already existing in betterproto) before making an actual release, but it should already be usable now.

I started adding your PR to the fork, see https://github.com/betterproto/python-betterproto2-compiler/pull/14/files . I used jinja blocks to avoid code duplication when possible.

Concerning the tests, I think that spawning a real server would indeed be a better way to test everything.

For now, the new synchronous client are just added to the generated files with a different name, but I would like to add an option to choose between no clients, only sync clients, only async clients or both clients.

Also, I think it would be better to have a ` SyncServiceStub` base class and to keep supporting `metadata`, `timeout` etc.

PR #652, opened 2024-12-19
feat: fix dump one of fields

aymen-k:
## Summary

Hey 👋 
I generated code with pydantic validation. 
A nested class failed the validation. For some reason, the object is validated twice once with the initialized object and once with empty fields. 

```
    @model_validator(mode="after")
    def check_oneof(cls, values):
        return cls._validate_field_groups(values)
```
Switching the order of the condition made it possible to avoid the second 'extra' validation.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/src/betterproto/__init__.py
+++ b/src/betterproto/__init__.py
@@ -968,9 +968,9 @@ class Message(ABC):
                 field_name=field_name, meta=meta
             )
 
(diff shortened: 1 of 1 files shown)


Gobot1234:
Hi there, what does this fix? Please can you add a regression test

PR #654, opened 2025-01-13
Update poetry to version 2

jameshilliard:
Fix some issues flagged by poetry check.


pyproject.toml
```diff
@@ -162,5 +165,5 @@ require_poetry = true
 """
 
 [build-system]
-requires = ["poetry-core>=1.0.0,<2"]
+requires = ["poetry-core>=2.0.0"]
```
abn, line 168:
```suggestion
requires = ["poetry-core>=2.0.0,<3"]
```
Safer to have the upper bound to avoid surprises.

abn (changes requested):

pyproject.toml
```diff
@@ -1,18 +1,21 @@
...
+    {name = "Daniel G. Taylor", email = "danielgtaylor@gmail.com"}
+]
 readme = "README.md"
 repository = "https://github.com/danielgtaylor/python-betterproto"
 keywords = ["protobuf", "gRPC"]
 license = "MIT"
 packages = [
     { include = "betterproto", from = "src" }
 ]
+requires-python = ">=3.8"
```
abn, line 15:
```suggestion
requires-python = ">=3.8,<4.0"
```

For equivalency with previous configuration.

jameshilliard: I think this would be unnecessarily strict, there's no reason to assume 4.0 will break compatibility at this time, generally upper limits for python versions should only be used when there's a known incompatibility AFAIU.

abn: Debatable. But for now my comment is that the prior version was `^3.8` which has an upper bound of `<4.0`.

jameshilliard: applied as suggested

abn:
The CI Is still installing Poetry 1.8.5. 

https://github.com/danielgtaylor/python-betterproto/actions/runs/12758366227/job/35583665469#step:5:142

Might need to update that in a separate PR first.

jameshilliard:
I think I got tests all passing now, had to drop python 3.8 support(python 3.8 is deprecated).

abn (approved):

PR #665, opened 2025-04-16
Have an output file for every class with the name of the class instea…

jwasys:
…d of a __init__.py. Now we can parse a tree of .proto files with multiple .proto files in each sub dir without overwriting the __init__.py with every invocation of the compiler

## Checklist

- [ ] This PR fixes an issue.


jwasys:
This should be configurable of coarse but for now it servers my purpose.

